PR #3, opened 2018-07-13
Added Fish installation-procedure

jakibaki:
The old version was a bit misleading by implying that the installation would work that way in all shells.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,12 +17,14 @@ For macOS and Linux:
 - Paste this command to your terminal:
 
 ```bash
+bash
 (
  set -x; cd "$(mktemp -d)" &&
  curl -fsSLO "https://storage.googleapis.com/krew-test/krew.zip" &&
  unzip krew.zip &&
  "./build/krew-$(uname | tr '[:upper:]' '[:lower:]')" install krew
 )
+exit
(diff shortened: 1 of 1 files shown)


ahmetb:
I am reluctant to accept this. It makes the installation instructions longer. Exotic shells will have different syntax.  This coincidentally works on Zsh because the used syntax is bash and zsh compatible. 

Is it not possible to run bash temporarily from fish, then run these cmds and exit?

jakibaki:
I tried to build an universal way by piping echo into bash but ultimately failed because I couldn't find a way to escape the command in a way which works on all shells.

If adding instructions for different shells is too much then adding a simple `bash` in front of the instructions and an `exit` at the end as you suggested would definitely work fine in fish (and all other shells I can think of).

ahmetb:
Unfortunately I don't think we should do `bash; .... ; exit`.

It's a fair assumption that fish and (other exotic shell–) users know how to start bash from their shells and exit it properly. Hopefully, in the future we'll have a one-liner installer `curl | bash` which will prevent this problem altogether.

Thanks for the PR, but I'll close it. Happy to review other contributions from you.

Issue #10, opened 2018-07-23
Remove: Zip-inflation


lbb:
Zip inflation currently tried to get rid of the first folder in a zip archive.

This behavior is inconsistent, it should be removed before we have to move many plugins.
See: https://github.com/GoogleContainerTools/krew/pull/9

PR #9, opened 2018-07-21
Fix to be able to extract zip file which has no directory structure

superbrothers:
This PR fixes to be able to install zip file which has no directory structure of plugin. Currently `extractZIP` function failed to extract zip file which has no directory structure.

The following steps reproduce failure.
```
$ kubectl version --client
Client Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.0", GitCommit:"91e7b4fd31fcd3d5f436da26c980becec37ceefe", GitTreeState:"clean", BuildDate:"2018-06-27T20:17:28Z", GoVersion:"go1.10.2", Compiler:"gc", Platform:"darwin/amd64"}
$ kubectl plugin krew version
OPTION          VALUE
BasePath        /Users/ksuda/.kube/plugins/krew
DownloadPath    /var/folders/_9/7l9x_d6170qcvcffhsx77wrw0000gn/T/krew
ExecutedVersion 437272f4f52b6d114f87b366cc3d069f54ab7451dc530c9b6517139fc150f908
IndexPath       /Users/ksuda/.kube/plugins/krew/index
IndexURI        https://github.com/GoogleContainerTools/krew-index.git
InstallPath     /Users/ksuda/.kube/plugins/krew/store
IsPlugin        true
$ cat <<EOL > service.yaml
apiVersion: krew.googlecontainertools.github.com/v1alpha1
kind: Plugin
metadata:
  name: service
spec:
  platforms:
  - uri: https://github.com/superbrothers/kubectl-service-plugin/releases/download/v1.1.0/service-darwin-amd64.zip
    sha256: 8197bc5506c628381cf1ae73ec03cc44d0513f6ca102d8b7165ba7dff034208e
    files:
    - from: "*"
      to: "."
    selector:
      matchLabels:
        os: darwin
  version: v1.1.0
  shortDescription: Open the Kubernetes URL(s) for the specified service in your browser.
  description: |
    Open the Kubernetes URL(s) for the specified service in your browser
    through a local proxy server using kubectl proxy.
EOL
$ kubectl plugin install --source service.yaml -v=4
I0721 18:46:01.744615   86316 git.go:62] Going to run git pull --ff-only -v
From github.com:GoogleContainerTools/krew-index
 = [up to date]      master     -> origin/master
Already up-to-date.
Updated index
Will install plugin: service
I0721 18:46:04.838917   86316 install.go:105] Installing plugin: service
I0721 18:46:04.839787   86316 install.go:64] Looking for installed versions
[...]

PR #11, opened 2018-07-23
Remove Zip-inflation

lbb:
Based on superbrothers/fix-extract-zip 1ed5e47

This introduces a breaking change. Single
dir zip files won't be inflated anymore.
File globs after this point should honor
the real file paths.
See: #9 
Closes: #10


ahmetb:
I'm guessing this requires some update to the design or developer guide as well (if we mentioned this zip directory structure).

PR #14, opened 2018-07-28
krew info: Improve error message when plugin doesn't exist

### LINKED ISSUE #12: krew info's error message isn't friendly when plugin doesn't exist
When plugin doesn't exist, `krew info` outputs error message as follows:
```
$ kubectl plugin info hoge
F0728 16:15:27.943022    8904 info.go:40] failed to read the plugin file, err: failed to open index file, err: open /Users/ksuda/.kube/plugins/krew/index/plugins/hoge.yaml: no such file or directory
error: exit status 255
```

Personally, I feel the above message is not user friendly.

I expect error message as follows:
```
$ kubectl plugin info hoge
Error: plugin "hoge" not found
```

superbrothers:
This PR improves `krew info`'s error message when plugin does not exist.

```
$ kubectl plugin info hoge
Error: plugin "hoge" not found

Usage:
  krew info [flags]

Flags:
  -h, --help   help for info

Global Flags:
      --alsologtostderr                  log to standard error as well as files
      --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
      --log_dir string                   If non-empty, write log files in this directory
      --logtostderr                      log to standard error instead of files (default true)
      --stderrthreshold severity         logs at or above this threshold go to stderr (default 2)
  -v, --v Level                          log level for V logs
      --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging

F0728 17:28:23.820914    6508 root.go:49] plugin "hoge" not found
```

/fix #12

### DIFF
--- a/cmd/krew/cmd/info.go
+++ b/cmd/krew/cmd/info.go
@@ -33,17 +33,16 @@ var infoCmd = &cobra.Command{
 	Short: "Info shows plugin details",
 	Long: `Info shows plugin details.
 Use this command to find out about plugin requirements and caveats.`,
-	RunE: func(cmd *cobra.Command, args []string) error {
+	Run: func(cmd *cobra.Command, args []string) {
 		for _, arg := range args {
 			plugin, err := indexscanner.LoadPluginFileFromFS(paths.Index, arg)
-			if err != nil {
-				glog.V(4).Info(err.Error())
-				return fmt.Errorf("plugin %q not found\n", arg)
+			if os.IsNotExist(err) {
+				glog.Fatalf("plugin %q not found", arg)
+			} else if err != nil {
+				glog.Fatal(err)
 			}
 			printPluginInfo(os.Stdout, plugin)
 		}
-
-		return nil
 	},
 	PreRunE: checkIndex,
 	Args:    cobra.MinimumNArgs(1),
--- a/cmd/krew/cmd/info.go
+++ b/cmd/krew/cmd/info.go
@@ -33,14 +33,17 @@ var infoCmd = &cobra.Command{
 	Short: "Info shows plugin details",
 	Long: `Info shows plugin details.
 Use this command to find out about plugin requirements and caveats.`,
-	Run: func(cmd *cobra.Command, args []string) {
+	RunE: func(cmd *cobra.Command, args []string) error {
 		for _, arg := range args {
 			plugin, err := indexscanner.LoadPluginFileFromFS(paths.Index, arg)
 			if err != nil {
-				glog.Fatal(err)
+				glog.V(4).Info(err.Error())
+				return fmt.Errorf("plugin %q not found\n", arg)
 			}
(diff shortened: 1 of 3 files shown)


cmd/krew/cmd/info.go
```diff
@@ -33,14 +33,17 @@ var infoCmd = &cobra.Command{
 	Short: "Info shows plugin details",
 	Long: `Info shows plugin details.
 Use this command to find out about plugin requirements and caveats.`,
-	Run: func(cmd *cobra.Command, args []string) {
+	RunE: func(cmd *cobra.Command, args []string) error {
 		for _, arg := range args {
 			plugin, err := indexscanner.LoadPluginFileFromFS(paths.Index, arg)
 			if err != nil {
-				glog.Fatal(err)
+				glog.V(4).Info(err.Error())
```
ahmetb: I think we can keep `Run` as opposed to `RunE` for consistency right now. I recall we had an issue with RunE that was causing a user-facing experience issue, so please revert that part.

I think we can do something like:

```
if os.IsNotExist(err) {
    glog.Fatal("plugin not found")
} else if err != nil {
    glog.Fatal("err")
}
```

superbrothers: `LoadPluginListFromFS` function returns wrapped error, so we cannot use `os.IsNotExist`.

```go
// LoadPluginListFromFS will parse and retrieve all plugin files.
func LoadPluginListFromFS(indexDir string) (index.IndexList, error) {
	var indexList index.IndexList
	indexDir, err := filepath.EvalSymlinks(indexDir)
	if err != nil {
		return indexList, err
	}

	files, err := ioutil.ReadDir(filepath.Join(indexDir, "plugins"))
	if err != nil {
		return indexList, fmt.Errorf("failed to open index dir, err: %v", err)
	}
```

I will change `LoadPluginListFromFS` function to return non wrapped error. WDYT?

ahmetb: Sorry, please ignore my earlier comment. 

I think the safest option here would be to do return an unwrapped error from `ReadPluginFile` only when the error is `os.IsNotExist`.

```
func ReadPluginFile(indexFilePath string) (index.Plugin, error) {
	f, err := os.Open(indexFilePath)
	if err != nil {
		return index.Plugin{}, fmt.Errorf("failed to open index file, err: %v", err)
	}
	return DecodePluginFile(f)
}
```

I think if you do it like:

```
if os.IsNotExist(err) { return err }
else if err != nil { return wrapped err }
```

and if you document that `...when plugin file not found, it returns an error that can be checked with os.IsNotExist.` we can solve this easily.

+ Need to write unit tests for those.

superbrothers: I fixed at 741f17fd915edb1aae49845e84ba04456bf832c6.

ahmetb:
Thanks!

PR #19, opened 2018-07-30
Update README.md

lbb:
This fixes the installation to be compatible with the https://github.com/GoogleContainerTools/krew/releases/tag/v0.1.0-alpha.1 release

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,20 +18,20 @@ For macOS and Linux:
 
 ```bash
 (
- set -x; cd "$(mktemp -d)" &&
- curl -fsSLO "https://storage.googleapis.com/krew-test/krew.zip" &&
- unzip krew.zip &&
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,20 +18,20 @@ For macOS and Linux:
 
 ```bash
 (
- set -x; cd "$(mktemp -d)" &&
- curl -fsSLO "https://storage.googleapis.com/krew-test/krew.zip" &&
- unzip krew.zip &&
- "./build/krew-$(uname | tr '[:upper:]' '[:lower:]')" install krew
+set -x; cd "$(mktemp -d)" &&
```
ahmetb: Let’s preserve 2-4 spaces indent

lbb: Does not work in my shell, I had to remove the spaces.
This makes it compatible. I don't think this will introduce a break in other setups.

PR #21, opened 2018-07-31
pkg/version: add git version stamps to krew binary

### LINKED ISSUE #20: Add version (commit info) to krew version command
It should be easy to tell which tag/commit that it's built from using `kubectl plugin krew version`. Currently only shows the sha256 of the krew plugin itself.

ahmetb:
I'm not entirely sure if we need both GitTag and GitCommit. This is primarily
for tagging release binaries easily (where the git repo is supposed to exist).

Fixes #20.

### DIFF
--- a/cmd/krew/cmd/version.go
+++ b/cmd/krew/cmd/version.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/GoogleContainerTools/krew/pkg/version"
 	"github.com/spf13/cobra"
 )
 
@@ -37,6 +38,8 @@ DownloadPath is the path used to store download binaries.`,
 		conf := map[string]string{
 			"IsPlugin":        fmt.Sprintf("%v", krewExecutedVersion != ""),
 			"ExecutedVersion": krewExecutedVersion,
+			"GitTag":          version.GitTag(),
+			"GitCommit":       version.GitCommit(),
 			"BasePath":        paths.Base,
 			"IndexPath":       paths.Index,
 			"IndexURI":        IndexURI,
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -1,0 +1,28 @@
+// Package version contains the version information of the krew binary.
+package version
+
+var (
+	// gitCommit contains the git commit idenifier.
+	gitCommit string
+
+	// gitTag contains the git tag or describe output.
+	gitTag string
+)
+
+// GitCommit returns the value stamped into the binary at compile-time or a
+// default "unknown" value.
+func GitCommit() string {
+	if gitCommit == "" {
+		return "unknown"
+	}
+	return gitCommit
+}
+
+// GitTag returns the value stamped into the binary at compile-time or a
+// default "unknown" value.
+func GitTag() string {
+	if gitTag == "" {
+		return "unknown"
+	}
+	return gitTag
+}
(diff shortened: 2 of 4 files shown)


cmd/krew/cmd/version.go
```diff
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/GoogleContainerTools/krew/pkg/version"
```
lbb: please split in 3 regions,
```
import(
stdlib...

project...

vendor...
)
```

pkg/version/version.go
```diff
@@ -0,0 +1,26 @@
...
+var (
+	// GitCommit contains the git commit idenifier.
+	GitCommit string
+
+	// GitTag contains the git tag or describe output.
+	GitTag string
+)
+
+// GitCommitString should be used to retrieve GitCommit or the default value.
+func GitCommitString() string {
```
lbb: We should either have the variable publicly accessible or have a getter function.

Does the variable injection work if a variable is unavailable?
If injection does not work sticking to a var with an empty string is more explicit than defaulting it to `"unknown"`, this seems to be the responsibility of the frontend.

pkg/version/version.go
```diff
@@ -0,0 +1,26 @@
+// Package version contains the version information of the krew binary.
+package version
+
+var (
+	// GitCommit contains the git commit idenifier.
+	GitCommit string
+
+	// GitTag contains the git tag or describe output.
+	GitTag string
```
lbb: As you said, I think we don't need the tag. It is not useful to the user.
The only thing needed for devs is the commit.

ahmetb: On the contrary, we actually need only the tag. As a dev I can figure out the commit id from the sha. But as a user I can’t find out vX.Y.Z easily from the sha OR commitID.

pkg/version/version.go
ahmetb: Var has to be publicly accessible otherwise the -X flag doesn’t work properly. 

This is a fairly common pattern and is safe. There’s a lot of undefined behavior around ldflags when the variable already has a value etc.

cmd/krew/cmd/version.go
ahmetb: I easily found more examples of this already in the repo (se we’re already not there) and as far as I know this is not a convention  + no linter check has failed.

pkg/version/version.go
lbb: I agree that the tag is useful for the User.

We cannot figure out the CommitID from the zip sha. CommitID is for developers, it points to a specific version in time. The tag may point to a past point in time.

pkg/version/version.go
lbb: I just tested to set not exported variables and it worked.
What do you mean by properly?

I think we still should return `""` instead `"unknown"`

ahmetb: Seems like I recall incorrectly. Can't find any references to it.

I'll unexport this field and keep the method.

pkg/version/version.go
ahmetb: Let's keep both, it's not an API.

lbb:
LGTM

IMHO we should no return "unknown"

ahmetb:
> IMHO we should no return "unknown"

I think this is better than confusing users with no value at all and also provides a clean API surface that requires _less_ handling. Docker for example does this: https://github.com/moby/moby/blob/9ebed53c5431a7c8296107869f1b3feabee9e8f3/dockerversion/version_lib.go#L4:34 (I guess default values also work when provided to var directly) and Kubernetes as well has placeholder strings for some fields. https://github.com/kubernetes/kubernetes/blob/259e0743f10a484fcbb85045cb59e1776263db1d/pkg/version/base.go

PR #25, opened 2018-08-13
docs: add plugin naming guide

ahmetb:
This is to be used/enforced in krew index and document best practices for the
plugin dev community.

### DIFF
--- a/docs/DEVELOPER_GUIDE.md
+++ b/docs/DEVELOPER_GUIDE.md
@@ -38,6 +38,9 @@ Each plugin needs a plugin descriptor.
 Those files are called plugin descriptors, they describe the entrypoint of the 
 kubectl plugin. 
 
+See [Plugin Naming Style Guide](NAMING_GUIDE.md) for choosing the right name
+for your plugin.
+
 Commit and push the two new files to your public repository. 
 
 ### Making Plugins Publicly Accessible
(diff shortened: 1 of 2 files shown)


bacongobbler:
would this document be better off in the plugin KEP? Seems a lot of this has more to do with plugin naming issues, less so with krew.

bacongobbler:
I was referring to [the kubectl doc on plugins](https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/), sorry, not the KEP for krew.

ahmetb:
I am okay with having it somewhere in the community (i.e. `kubernetes/community` repo has similar stuff).

But I'm not sure if it belongs to the official Kubernetes documentation on plugins.  Simply because kubernetes documentation doesn't offer naming guides to anything (like extensions, CRDs, resource naming for deployment/pod etc) and right now the only entity that accepts plugins is the `krew-index` repo.

So I don't think there's a community position on suggesting or enforcing these names yet, even though these are good practices to follow.

I'm also not sure if everyone agrees with these. We're hoping to enforce it in the `krew-index` repo ASAP, as the naming has already sparkled a lot of debates in almost every PR we got so far.

bacongobbler:
fair enough!

ahmetb:
@bacongobbler Happy to revisit this and move it out of this repo if someone shepherds this in the community/sig-cli.

PR #32, opened 2018-09-05
Make macOS/Linux install command less scary and exotic

tstromberg:
Remove the subshell, omit debug mode (-x), and apply our shell style guide: https://google.github.io/styleguide/shell.xml

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,12 +17,10 @@ For macOS and Linux:
 - Paste this command to your terminal:
 
 ```bash
-(
-  set -x; cd "$(mktemp -d)" &&
-  curl -fsSLO "https://github.com/GoogleContainerTools/krew/releases/download/v0.1.0-alpha.1/krew.zip" &&
-  unzip krew.zip &&
-  "./out/build/krew-$(uname | tr '[:upper:]' '[:lower:]')" install krew
-)
+cd "$(mktemp -d)" \
+  && curl -LO "https://github.com/GoogleContainerTools/krew/releases/download/v0.1.0-alpha.1/krew.zip" \
+  && unzip krew.zip \
+  && "./out/build/krew-$(uname | tr '[:upper:]' '[:lower:]')" install krew
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,12 +17,10 @@ For macOS and Linux:
 - Paste this command to your terminal:
 
 ```bash
-(
-  set -x; cd "$(mktemp -d)" &&
-  curl -fsSLO "https://github.com/GoogleContainerTools/krew/releases/download/v0.1.0-alpha.1/krew.zip" &&
-  unzip krew.zip &&
-  "./out/build/krew-$(uname | tr '[:upper:]' '[:lower:]')" install krew
-)
```
ahmetb: Removing parentheses cause user to end up in a directory (mktemp -d). but having them actually bring the user back to their $PWD. That said I'm ok with removing `set -x`, that was just for visibility.

tstromberg: What are your thoughts on removing the shell magic entirely?

It causes an unnecessary dependency on us guessing what shell the user is using: For instance, the $() breaks for tcsh and fish users. && breaks for fish users. We can just make it:

```shell
curl -LO https://github.com/GoogleContainerTools/krew/releases/download/v0.1.0-alpha.1/krew.zip
unzip krew.zip
./out/build/krew-linux install krew
```

ahmetb: I'm concerned about just bash/zsh users, and I think people using tcsh/fish are probably already familiar with bash syntax to can convert it to their shell preference.

I'm hoping to replace this with a one liner `curl | sh` installer someday so this no longer becomes a problem.

Removing `&&` and `set -x` confuses beginners especially because:

1. it won't make it clear which command failed (disorganized output)
2. it will still execute all the commands and they will think the last error ("command not found") is the error but it can be curl

ahmetb:
I'll close this per the discussion above. It's a bit scary today but the `set -x` provides decent visibility and `&&` provides good error checking at least.

Issue #33, opened 2018-09-05
[🔥WIP🔥] Migrating to kubectl v1.12 new plugin model (BREAKING CHANGES), krew v0.2 release.


ahmetb:
Context:

- Kubernetes 1.12 implements git-style plugins (executables prefixed `kubectl-*` placed in PATH).

- Krew does **not** support this model (predates it), but we saw this coming during the design phase.

- This change will still **break krew** heavily.

- Now that kubernetes 1.12.0-beta.1 release is available, we need to make krew work with the new kubectl plugin model.

## Plan

- [x] **Determine which krew version will fully support the "new plugin model": (v0.2)**
- [x] **Provide breaking change notices**
  - [x] Release a patch version for krew, adding a warning to every major cmd (#40)
  - [x] Update README about krew version being incompatible (#36)
- [x] **Plugin manifest updates**
  - [x] Add a field to `platforms[]` indicating which file(s) should be symlinked (#35)
    - [x] allow naming the symlink from plugin name e.g: mtail&rarr;kubectl-mtail) (#42)
  - ~(Possibly) Disable unknown field checking in manifest YAML (and validate in "audit" cmd) to allow graceful addition of this field (but I think breaking everyone in this case is ok due to low user count)~
  - [x] (Possibly) Error out for manifests in v1alpha1. (#72)
- [x] **Krew mechanics**
  - [x] Determine a binary symlink path (like `~/.krew/bin`) (#49)
  - [x] Modify installation/upgrade to take symlinks into account. (#42)
  - [x] Manually test upgrading krew.
  - ~~Detect kubectl version (refuse to run if <v1.12) (#41)~~
- [x] **Index updates**
  - [x] Backport and test all krew-index with the new bin path. (https://github.com/GoogleContainerTools/krew-index/pull/24)
- [x] **Documentation**
  - [x] Add note: "krew only works with kubectl-1.12 and above" (remove breaking change notice)
  - [x] Have users add krew symlink path to $PATH
    - [x] Update install script and user guide (show how to update PATH)
    - [x] Mention this in krew meta-plugin CAVEATS section (#114)
  - [x] Update developer guide with changes to kubectl-1.12 plugin model (#100)
  - [x] Update developer guide with changes to krew (bin dir, symlinks etc) (#62)
  - [x] Update user guide with nrew instructions (#93)
- Upstream issues
[...]

PR #35, opened 2018-09-09
Add BIN, entrypoint field for kubectl 1.12

lbb:
This CL adds a bin field.
Bin specifies the entry point for a kubectl plugin after the 1.12 release.

### DIFF
--- a/pkg/environment/environment.go
+++ b/pkg/environment/environment.go
@@ -24,7 +24,7 @@ import (
 	"k8s.io/client-go/util/homedir"
 )
 
-// KrewPaths contains all important enviroment paths
+// KrewPaths contains all important environment paths
 type KrewPaths struct {
 	// Base is the path of the krew root.
 	// The default path is ~/.kube/plugins/krew
@@ -41,6 +41,10 @@ type KrewPaths struct {
 	// Download is a directory where plugins will be temporarily downloaded to.
 	// This is currently a path in os.TempDir()
 	Download string
+
+	// Bin is the path krew links all plugin binaries to.
+	// This path has to be added to the $PATH.
+	Bin string
 }
 
 func parseEnvs(environ []string) map[string]string {
@@ -65,26 +69,16 @@ func MustGetKrewPathsFromEnvs(envs []string) KrewPaths {
 		Index:    filepath.Join(base, "index"),
 		Install:  filepath.Join(base, "store"),
 		Download: filepath.Join(os.TempDir(), "krew"),
+		Bin:      filepath.Join(base, "bin"),
 	}
 }
 
-func getKubectlPluginsPath(envs []string) string {
-	envvars := parseEnvs(envs)
-
-	// Look for "${KUBECTL_PLUGINS_PATH}"
-	if path, ok := envvars["KUBECTL_PLUGINS_PATH"]; ok && path != "" {
-		return path
-	}
-
-	// TODO(lbb): make os call testable! (dep injection)
-	// Look for "${XDG_DATA_DIRS}/kubectl/plugins"
-	if path := envvars["XDG_DATA_DIRS"]; path != "" {
-		fullPath := filepath.Join(path, "kubectl", "plugins")
-		stat, err := os.Stat(fullPath)
-		if err == nil && stat.IsDir() {
-			return fullPath
-		}
-	}
+func getKubectlPluginsPath(_ []string) string {
+	// envvars := parseEnvs(envs)
+	// // Look for "${KREW_HOMEDIR}"
+	// if path, ok := envvars["KREW_HOMEDIR"]; ok && path != "" {
+	// 	return path
+	// }
 
 	// Golang does no '~' expansion
 	return filepath.Join(homedir.HomeDir(), ".kube", "plugins")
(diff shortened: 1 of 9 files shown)


pkg/environment/environment.go
```diff
@@ -65,33 +69,24 @@ func MustGetKrewPathsFromEnvs(envs []string) KrewPaths {
...
 		Download: filepath.Join(os.TempDir(), "krew"),
+		Bin:      filepath.Join(base, "bin"),
 	}
 }
 
-func getKubectlPluginsPath(envs []string) string {
-	envvars := parseEnvs(envs)
-
-	// Look for "${KUBECTL_PLUGINS_PATH}"
-	if path, ok := envvars["KUBECTL_PLUGINS_PATH"]; ok && path != "" {
```
ahmetb: Does the implementation completely ignore KUBECTL_PLUGINS_PATH now? (I'm just trying to double-check.) The 1.12 release branch docs don't seem to be updated. https://github.com/kubernetes/website/blob/release-1.12/content/en/docs/tasks/extend-kubectl/kubectl-plugins.md

pkg/environment/environment.go
```diff
@@ -65,33 +69,24 @@ func MustGetKrewPathsFromEnvs(envs []string) KrewPaths {
...
-	// Look for "${XDG_DATA_DIRS}/kubectl/plugins"
-	if path := envvars["XDG_DATA_DIRS"]; path != "" {
-		fullPath := filepath.Join(path, "kubectl", "plugins")
-		stat, err := os.Stat(fullPath)
-		if err == nil && stat.IsDir() {
-			return fullPath
-		}
-	}
+func getKubectlPluginsPath(_ []string) string {
+	// envvars := parseEnvs(envs)
```
ahmetb: leftover comment

pkg/environment/environment.go
```diff
@@ -65,33 +69,24 @@ func MustGetKrewPathsFromEnvs(envs []string) KrewPaths {
...
+	// // Look for "${KREW_HOMEDIR}"
+	// if path, ok := envvars["KREW_HOMEDIR"]; ok && path != "" {
+	// 	return path
+	// }
 
 	// Golang does no '~' expansion
 	return filepath.Join(homedir.HomeDir(), ".kube", "plugins")
```
ahmetb: I think now that plugins don't have to be in ~/.kube/plugins, do you think we should start using ~/.krew ?

People `rm -rf ~/.kube` to reset their kubectl settings all the time. (I did that myself accidentally the other day).

pkg/index/validate.go
```diff
@@ -69,6 +69,9 @@ func (p Platform) Validate() error {
 	if p.Head == "" && p.URI == "" {
 		return fmt.Errorf("head or URI have to be set")
 	}
+	if p.Bin == "" {
+		return fmt.Errorf("bin has to be set")
+	}
```
ahmetb: hmm what if we also added `if !starswith(p.Bin, "kubectl-")` check?

pkg/index/types.go
```diff
@@ -45,6 +45,11 @@ type Platform struct {
 
 	Selector *metav1.LabelSelector `json:"selector,omitempty"`
 	Files    []FileOperation       `json:"files"`
+
+	// Bin specifies the path to an executable (kubernetes 1.12).
```
ahmetb: > to an executable (kubernetes 1.12).

to

> to the plugin executable.

no need to note 1.12 IMO unless you're also noting krew version.

pkg/environment/environment.go
lbb: I'd like to keep it for now (dev-phase).
We should decide where to install krew, as it is required to install plugin binaries into the PATH. We should maybe introduce **KREW_PATH** and **KREW_BIN** variable.

ahmetb:
Do you mind rolling up the `docs/*` changes to the same PR? We probably shouldn't forget those.

pkg/environment/environment.go
```diff
@@ -65,33 +69,18 @@ func MustGetKrewPathsFromEnvs(envs []string) KrewPaths {
...
-		fullPath := filepath.Join(path, "kubectl", "plugins")
-		stat, err := os.Stat(fullPath)
-		if err == nil && stat.IsDir() {
-			return fullPath
-		}
-	}
-
+func getKubectlPluginsPath(_ []string) string {
 	// Golang does no '~' expansion
 	return filepath.Join(homedir.HomeDir(), ".kube", "plugins")
```
ahmetb: (copying from the disappeared thread)
I think now that plugins don't have to be in ~/.kube/plugins, do you think we should start using ~/.krew ?

People `rm -rf ~/.kube` to reset their kubectl settings all the time. (I did that myself accidentally the other day).

lbb: I'd like to keep it for now (dev-phase).
We should decide where to install krew, as it is also required to install plugin binaries into the PATH. We should maybe introduce KREW_PATH and KREW_BIN variable.

ahmetb: maybe just convert this to a method that doesn't get any args for now.

ahmetb: also maybe if this method is untested, we can just delete it.

ahmetb:
please rebase so the "go test" works now.

ahmetb:
Thanks! What should we check off in #33?

lbb:
The tracking file currently looks good to me 👍

PR #39, opened 2018-09-12
ci: add codecov for code coverage stats

ahmetb:


ahmetb:
looks like github bot for codecov is not working yet but some data at  https://codecov.io/gh/GoogleContainerTools/krew/commits

PR #41, opened 2018-09-17
pkg/kubectl: check version of kubectl

ahmetb:
If someone tries to execute krew binary standalone with kubectl <1.12, adding
a failure.

I'm not entirely sure if it's a valid use case. The only encounter of this
could be when someone installs krew, and directly tries to invoke the binary
(as krew 0.2 won't work on kubectl 1.11 or older).

Probably harmless for now (also added a KREW_DISABLE_KUBECTL_VERSION_CHECK
escape hatch for dev/build purposes). We can remove this pkg and check soon.
Currently adds ~100 ms to each krew cmd.

### DIFF
--- a/cmd/krew/cmd/root.go
+++ b/cmd/krew/cmd/root.go
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/GoogleContainerTools/krew/pkg/kubectl"
+
 	"github.com/GoogleContainerTools/krew/pkg/environment"
 	"github.com/GoogleContainerTools/krew/pkg/gitutil"
 
@@ -39,6 +41,20 @@ var rootCmd = &cobra.Command{
 You can invoke krew through kubectl with: "kubectl plugin [krew] option..."`,
 	PersistentPreRun: func(cmd *cobra.Command, _ []string) {
 		bindEnvironmentVariables(viper.GetViper(), cmd)
+
+		// detect kubectl version to prevent running krew standalone with older
+		// versions of kubectl.
+		// TODO(ahmetb): this may not be necessary after most users are on 1.12
+		glog.V(4).Infof("checking kubectl version")
+		ok, err := kubectl.IsSupportedVersion()
+		if err != nil {
+			glog.Fatalf("error checking kubectl version: %+v", err)
+		}
+		if !ok && os.Getenv("KREW_DISABLE_KUBECTL_VERSION_CHECK") == "" {
+			glog.Fatal("You are running on an unsupported version of \"kubectl\"." +
+				" This version of krew requires minimum kubectl 1.12.0.")
+		}
+		glog.V(4).Infof("checked kubectl version")
 	},
 }
 
--- a/pkg/kubectl/kubectl_version.go
+++ b/pkg/kubectl/kubectl_version.go
@@ -1,0 +1,66 @@
+package kubectl
+
+import (
+	"encoding/json"
+	"fmt"
+	"os/exec"
+	"strings"
+
+	"github.com/coreos/go-semver/semver"
+)
+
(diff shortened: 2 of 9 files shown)


cmd/krew/cmd/root.go
```diff
@@ -39,6 +41,20 @@ var rootCmd = &cobra.Command{
 You can invoke krew through kubectl with: "kubectl plugin [krew] option..."`,
 	PersistentPreRun: func(cmd *cobra.Command, _ []string) {
 		bindEnvironmentVariables(viper.GetViper(), cmd)
+
+		// detect kubectl version to prevent running krew standalone with older
+		// versions of kubectl.
+		// TODO(ahmetb): this may not be necessary after most users are on 1.12
+		glog.V(4).Infof("checking kubectl version")
```
lbb: Capitalize

cmd/krew/cmd/root.go
```diff
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/GoogleContainerTools/krew/pkg/kubectl"
+
```
lbb: remove space

cmd/krew/cmd/root.go
```diff
@@ -39,6 +41,20 @@ var rootCmd = &cobra.Command{
...
 	PersistentPreRun: func(cmd *cobra.Command, _ []string) {
 		bindEnvironmentVariables(viper.GetViper(), cmd)
+
+		// detect kubectl version to prevent running krew standalone with older
+		// versions of kubectl.
+		// TODO(ahmetb): this may not be necessary after most users are on 1.12
+		glog.V(4).Infof("checking kubectl version")
+		ok, err := kubectl.IsSupportedVersion()
+		if err != nil {
+			glog.Fatalf("error checking kubectl version: %+v", err)
```
lbb: Capitalize

cmd/krew/cmd/root.go
```diff
@@ -39,6 +41,20 @@ var rootCmd = &cobra.Command{
...
+		}
+		if !ok && os.Getenv("KREW_DISABLE_KUBECTL_VERSION_CHECK") == "" {
+			glog.Fatal("You are running on an unsupported version of \"kubectl\"." +
+				" This version of krew requires minimum kubectl 1.12.0.")
+		}
+		glog.V(4).Infof("checked kubectl version")
```
lbb: Capitalize

pkg/kubectl/kubectl_version.go
```diff
@@ -0,0 +1,66 @@
...
+	"fmt"
+	"os/exec"
+	"strings"
+
+	"github.com/coreos/go-semver/semver"
+)
+
+var (
+	minKubectl      = semver.New("1.12.0-alpha.0")
+	versionProvider = execKubectlVersion
```
lbb: defaultVersionProvider

lbb: or remove, why it is needed?

pkg/kubectl/kubectl_version.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+func execKubectlVersion() (string, error) {
+	cmd := exec.Command("kubectl", "version", "--client", "--output=json")
+	b, err := cmd.CombinedOutput()
+	if err != nil {
+		return "", fmt.Errorf("failed to execute kubectl version. error=%v, out=%v", err, string(b))
```
lbb: put the cascading error at the end of the message

pkg/kubectl/kubectl_version.go
ahmetb: this is for mocking easily.

PR #42, opened 2018-09-18
Use plugin linking, instead of descriptors

lbb:
This PR removes all functionality that does not exist in kubectl v1.12:
* Plugin descriptors
* KUBECTL_* env variables
* Change resolve plugin function
* Add linking functions to install,update,remove,list


pkg/environment/environment.go
```diff
@@ -96,8 +89,17 @@ func GetExecutedVersion(paths KrewPaths, cmdArgs []string) (string, bool, error)
...
-	_, ok := parseEnvs(environ)["KUBECTL_PLUGINS_DESCRIPTOR_NAME"]
-	return ok
+// DefaultSymlinkResolver resolves symlinks paths
+func DefaultSymlinkResolver(path string) (string, error) {
+	s, err := os.Lstat(path)
+	if err != nil {
+		return "", fmt.Errorf("failed to stat the currently executed path")
+	}
+
+	if s.Mode()&os.ModeSymlink != 0 {
```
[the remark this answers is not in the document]
lbb: The only thing that would also work is `s.Mode()&os.ModeSymlink != os.ModeSymlink`. But I've never seen that :)

pkg/installation/install.go
```diff
@@ -94,5 +100,46 @@ func Remove(p environment.KrewPaths, name string) error {
...
+	var err error
+	if err = os.Symlink(binary, dst); err != nil && !os.IsExist(err) {
+		return fmt.Errorf("failed to create a symlink form %q to %q, err: %v", bindir, dst, err)
+	}
+	glog.V(2).Infof("Created symlink in %q\n", dst)
+
+	return nil
+}
+
+func removeLink(bindir string, plugin string) error {
```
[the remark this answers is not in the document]
lbb: IMHO, both functions: `delete`, `createOrUpdate` are specialized to only execute on the assumptions and file structure we have in krew.
It is really difficult to misuse the function.

docs/DEVELOPER_GUIDE.md
```diff
@@ -210,6 +206,17 @@ The resulting installation directory will look like:
 
 ---
 
+Krew needs to link a plugin executeable to the `PATH`. The `bin` field has to
+point to the executable in the installation folder.
+
+```yaml
+...
+    bin: "./kubectl-foo"
```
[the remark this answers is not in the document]
lbb: We don't need quotes, but I think using quotes is a good practice (Also mentioned in previous reviews. It is consistent through our codebase.)

cmd/krew/cmd/update.go
```diff
@@ -24,7 +24,7 @@ import (
 )
 
 // IndexURI points to the upstream index.
-const IndexURI = "https://github.com/GoogleContainerTools/krew-index.git"
+const IndexURI = "https://github.com/lbb/krew-test-index.git"
```
ahmetb: should we revert this or keep until we update the index
OR maybe we should add a `IndexBranch` feature that checks out to a particular branch after cloning WDYT?

pkg/installation/install.go
```diff
@@ -72,11 +76,19 @@ func Install(p environment.KrewPaths, plugin index.Plugin, forceHEAD bool) error
 	}
 
 	glog.V(1).Infof("Finding download target for plugin %s", plugin.Name)
-	version, uri, fos, err := getDownloadTarget(plugin, forceHEAD)
+	version, uri, fos, bin, err := getDownloadTarget(plugin, forceHEAD)
 	if err != nil {
 		return err
 	}
-	return downloadAndMove(version, uri, fos, filepath.Join(p.Download, plugin.Name), filepath.Join(p.Install, plugin.Name))
+	dst, err := downloadAndMove(version, uri, fos, filepath.Join(p.Download, plugin.Name), filepath.Join(p.Install, plugin.Name))
```
ahmetb: this next 3 calls (downloadAndMove, checkSubPath,createOrUpdateLink) are duplicated between here and upgrade.go with this PR.

can we move them to a common place (like an unexported method) and call them?

when it's just downloadAndMove it's cool, but when it's a 3-method sequence, avoiding repetition would be better.

pkg/installation/install.go
```diff
@@ -94,5 +106,62 @@ func Remove(p environment.KrewPaths, name string) error {
...
+	}
+	if _, ok := pathutil.IsSubPath(subPathAbs, pathAbs); !ok {
+		return fmt.Errorf("the path %q does not extend the sub-path %q", path, subPath)
+	}
+	return nil
+}
+
+func removeLink(binDir string, plugin string) error {
+	dst := filepath.Join(binDir, pluginNameToBin(plugin))
+	if err := checkSubPath(binDir, dst); err != nil {
```
ahmetb: FYI it seems like there's a bit duplication when this method is called from createOrUpdate link. it goes like this:

```
downloadAndMove
checkSubPath <--- first
createOrUpdateLink
  removeLink
    checkSubPath <--- second
    os.Remove
  os.Symlink
```

it would be good to decide whether we'll pass paths that are sanitized _once_ or we'll just check it everywhere it's used. I'm a fan of checking once.

pkg/installation/install.go
```diff
@@ -94,5 +106,62 @@ func Remove(p environment.KrewPaths, name string) error {
...
+	glog.V(2).Infof("Creating symlink from %q to %q\n", binary, dst)
+	if err := os.Symlink(binary, dst); err != nil {
+		return fmt.Errorf("failed to create a symlink form %q to %q, err: %v", binDir, dst, err)
+	}
+	glog.V(2).Infof("Created symlink in %q\n", dst)
+
+	return nil
+}
+
+func checkSubPath(subPath, path string) error {
```
ahmetb: FYI it seems like everywhere we call `pathutil.IsSubPath` we're passing filepath.Abs'd strings.

Do you think if we add filepath.Abs _inside_ `IsSubPath` we can avoid (1) creating little methods like this (currently untested) (2) remove code duplication when calling IsSubPath?

pkg/installation/util.go
```diff
@@ -16,6 +16,7 @@ package installation
 
 import (
 	"fmt"
+	"github.com/GoogleContainerTools/krew/pkg/pathutil"
```
ahmetb: nit: this import should be in a separate block

pkg/installation/install.go
```diff
@@ -94,5 +106,62 @@ func Remove(p environment.KrewPaths, name string) error {
...
+
+func removeLink(binDir string, plugin string) error {
+	dst := filepath.Join(binDir, pluginNameToBin(plugin))
+	if err := checkSubPath(binDir, dst); err != nil {
+		return fmt.Errorf("failed to remove link, err: %v", err)
+	}
+	if err := os.Remove(dst); err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("failed to remove the symlink in %q, err: %v", dst, err)
+	} else if err == nil {
+		glog.V(3).Infof("Removed symlink from %q\n", dst)
```
ahmetb: nit:  trailing `\n` is  redundant. same comment for line 124 and 128:

	124: glog.V(2).Infof("Creating symlink from %q to %q\n", binary, dst)
	128: glog.V(2).Infof("Created symlink in %q\n", dst)

cmd/krew/cmd/update.go
lbb: Oh, that slipped in from local dev. I’ll totally revert it :)
After using pure go-git /& multi-repos we should revisit that idea 👍

ahmetb:
Thanks a lot @lbb.
I think the comments are addressed, although I might be missing a few. If we catch them again we can send smaller patches.
Did I check off the right items in #33?

PR #49, opened 2018-09-24
move krew install path to $HOME/.krew

ahmetb:
cc: @lbb

### DIFF
--- a/pkg/environment/environment.go
+++ b/pkg/environment/environment.go
@@ -28,7 +28,7 @@ import (
 // KrewPaths contains all important environment paths
 type KrewPaths struct {
 	// Base is the path of the krew root.
-	// The default path is ~/.kube/plugins/krew
+	// The default path is ~/.krew
 	Base string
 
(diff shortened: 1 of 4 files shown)


pkg/environment/environment.go
```diff
@@ -59,7 +59,7 @@ func parseEnvs(environ []string) map[string]string {
 
 // MustGetKrewPaths returns ensured index paths for krew.
 func MustGetKrewPaths() KrewPaths {
-	base := filepath.Join(homedir.HomeDir(), ".kube", "plugins", "krew")
+	base := filepath.Join(homedir.HomeDir(), ".krew")
```
lbb: I don't like krew polluting the homedir. We should respect XDG directories if possible!

ahmetb: There's no XDG on macOS though. Not to mention, .kube isn't probably complying with this either?

I think we should take this PR as is, then let people specify a Krew home path through a KREW_ROOT environment variable.

PR #51, opened 2018-09-25
Add .tar.gz support for archive downloads

### LINKED ISSUE #34: Support more download formats than zip
For a lot of Go projects, binary releases are either available directly or in a `.tar.gz` format. With how many k8s related projects are written in Go, it seems like it would make sense to add support for those formats. As an example, I'm using goreleaser to release [rbac-lookup](https://github.com/reactiveops/rbac-lookup). That defaults to `.tar.gz` formats for Mac and Linux releases. Even just supporting downloading binaries directly without trying to unzip them would likely be helpful (and very straightforward).  I'm happy to do some work in this direction if this is a feature if there's interest.

ahmetb:
Added an extractTARGZ method + tests. Inferring archive types from url suffix.

Fixes #34.


ahmetb:
cc: @lbb

pkg/download/downloader.go
```diff
@@ -85,20 +88,89 @@ func extractZIP(targetDir string, read io.ReaderAt, size int64) error {
...
+		if hdr.Name == "pax_global_header" {
+			glog.V(4).Infof("tar: skipping pax_global_header file")
+			continue
+		}
+
+		path := filepath.Join(targetDir, filepath.FromSlash(hdr.Name))
+		switch hdr.Typeflag {
+		case tar.TypeDir:
+			if err := os.MkdirAll(path, os.FileMode(hdr.Mode)); err != nil {
+				return fmt.Errorf("failed to create directory from tar: %+v", err)
```
lbb: , err: %+v .. also for the rest of the fmt calls

ahmetb: I don't think the `err:` in  `failed to do x, err: %+v` is necessary.

PR #52, opened 2018-09-25
Do not require kubectl- prefix for bin

ahmetb:
I think I mistakenly suggested forcing a `kubectl-` prefix for the
"platform.bin" field. At the time I did not realize this is the source file
for symlink operation.

Since we automatically construct the finale executable name of the plugin
from `func pluginNameToBin`, we don't need to enforce the `kubectl-` prefix
for the original executables.

This also makes it possible to use kubectl 1.11 plugins in 1.12 without having
the authors change the file names in their repositories.


lbb:
This PR will break krew. We only need this function to generate the link destination name. We don’t enforce link target naming verification AFAIK.

PR #60, opened 2018-09-26
Convert dashes in plugin names to underscores

### LINKED ISSUE #59: [kubectl-1.12] Allow dash in plugin names
Any krew plugin named like `a-b-c` today is symlinked as `kubectl-a-b-c`.

This causes plugin to be invokable **only** through `kubectl a b c` and loses the dashes in the original name.

Per [kubectl/#540](https://github.com/kubernetes/kubectl/issues/540), the new design allows underscores (`_`) to be treated as `-` at invocation-time. This means, if a symlink is linked as `kubectl-a_b_c`, it can be invoked as `kubectl a-b-c` as desired (and equivalent of pre-1.12 behavior).

## Proposed solution 

Just replace dashes (`-`) with underscores `_`. This way we don't have to replace plugin identifiers in manifest filenames or `name:` fields.

This way krew can hide this implementation detail behind the covers and not expose it to both plugin devs and plugin users.

Will submit a PR shortly.

ahmetb:
For dashes to be preserved in a plugin's name, its executable name must convert
dashes to underscores.

For example a plugin named "foo-bar" would today be linked as:

    kubectl-foo-bar

therefore, it can only be invoked as:

    kubectl foo bar

However, linking the same plugin as:

    kubectl-foo_bar

allows plugin to be called as:

    kubectl foo-bar
    kubectl foo_bar

- convert OS detection pluginNameToBin() to explicit input for testability
- test pluginNameToBin().

Fixes #59.

### DIFF
--- a/pkg/installation/install.go
+++ b/pkg/installation/install.go
@@ -124,7 +124,7 @@ func Remove(p environment.KrewPaths, name string) error {
 }
 
 func createOrUpdateLink(binDir string, binary string, plugin string) error {
-	dst := filepath.Join(binDir, pluginNameToBin(plugin))
+	dst := filepath.Join(binDir, pluginNameToBin(plugin, isWindows()))
 
 	if err := removeLink(binDir, plugin); err != nil {
 		return fmt.Errorf("failed to remove old symlink, err: %v", err)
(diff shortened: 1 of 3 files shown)


ahmetb:
cc: @lbb

PR #61, opened 2018-09-26
removeLink: refactor and add tests

ahmetb:
- removeLink is now separated from plugin logic. moved plugin related argument
  prep to call-site.
- removeLink now fails explicitly for non-symlinks. since bin/ is supposed to
  be only symlinks, this sounds appropriate
- added test cases
- handled an ignored removeLink error in `installation.Remove()` method (this is supposed to be ok since we use the same symlink to resolve the plugin installation, so plugin is probably there...)

### DIFF
--- a/pkg/installation/install.go
+++ b/pkg/installation/install.go
@@ -119,14 +119,18 @@ func Remove(p environment.KrewPaths, name string) error {
 	}
 	glog.V(1).Infof("Deleting plugin version %s", version)
 	glog.V(3).Infof("Deleting path %q", filepath.Join(p.Install, name))
-	removeLink(p.Bin, name)
+
+	symlinkPath := filepath.Join(p.Bin, pluginNameToBin(name, isWindows()))
+	if err := removeLink(symlinkPath); err != nil {
+		return fmt.Errorf("could not uninstall symlink of plugin: %+v", err)
+	}
 	return os.RemoveAll(filepath.Join(p.Install, name))
 }
 
 func createOrUpdateLink(binDir string, binary string, plugin string) error {
 	dst := filepath.Join(binDir, pluginNameToBin(plugin, isWindows()))
 
-	if err := removeLink(binDir, plugin); err != nil {
+	if err := removeLink(dst); err != nil {
 		return fmt.Errorf("failed to remove old symlink, err: %v", err)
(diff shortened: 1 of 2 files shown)


pkg/installation/install.go
```diff
@@ -143,12 +147,20 @@ func createOrUpdateLink(binDir string, binary string, plugin string) error {
...
+func removeLink(path string) error {
+	if fi, err := os.Lstat(path); err == nil {
+		if fi.Mode()&os.ModeSymlink == 0 {
+			return fmt.Errorf("file %q is not a symlink (mode=%s)", path, fi.Mode())
+		}
+		if err := os.Remove(path); err != nil {
+			return fmt.Errorf("failed to remove the symlink in %q, err: %v", path, err)
+		}
+		glog.V(3).Infof("Removed symlink from %q", path)
+	} else if !os.IsNotExist(err) {
```
lbb: Can we please invert the case. !...Not..() is quite difficult to understand

pkg/installation/install.go
```diff
@@ -143,12 +147,20 @@ func createOrUpdateLink(binDir string, binary string, plugin string) error {
...
 }
 
-func removeLink(binDir string, plugin string) error {
-	dst := filepath.Join(binDir, pluginNameToBin(plugin))
-	if err := os.Remove(dst); err != nil && !os.IsNotExist(err) {
-		return fmt.Errorf("failed to remove the symlink in %q, err: %v", dst, err)
-	} else if err == nil {
-		glog.V(3).Infof("Removed symlink from %q", dst)
+// removeLink removes a symlink reference if exists.
+func removeLink(path string) error {
```
lbb: I don't think we need a generic remove link function. Can we keep the old function interface? `removeLink(binDir string, plugin string)`, that makes it easier to use.

ahmetb: @lbb That is really not clean code and vastly reduces testability.

We call `filepath.Join(bin, pluginNameToBin(name))` in so many unrelated places  it's at it is pointing to code smell.

pkg/installation/install.go
ahmetb: Agreed we can unfold and revert this if a bit.

PR #63, opened 2018-09-26
pkg/env: refactor KrewPaths to compute paths

ahmetb:
- remove access to internals of KrewPaths and provide methods
- rename KrewPaths to Paths (tentative)
- make Paths testable + add tests
- MustGetKrewPaths: add test for home dir detection

This PR paves the way to;
- add environment-variable overrides for krew base path and have it testable
- refactor common filepath.Join()s that happen around returned paths so they
  are not repeated in the plugin lifecycle logic.


ahmetb:
cc: @lbb

ahmetb:
I think the tests are fine. One of the builds (/push) was failing reproducibly by failing to clone the branch (it was trying to clone the wrong repo I think).

pkg/environment/environment_test.go
```diff
@@ -15,13 +15,57 @@
...
+	if got, expected := p.InstallPath(), filepath.FromSlash("/foo/store"); got != expected {
+		t.Fatalf("InstallPath()=%s; expected=%s", got, expected)
+	}
+	if got, expected := p.PluginInstallPath("my-plugin"), filepath.FromSlash("/foo/store/my-plugin"); got != expected {
+		t.Fatalf("PluginInstallPath()=%s; expected=%s", got, expected)
+	}
+	if got, expected := p.PluginVersionInstallPath("my-plugin", "v1"), filepath.FromSlash("/foo/store/my-plugin/v1"); got != expected {
+		t.Fatalf("PluginVersionInstallPath()=%s; expected=%s", got, expected)
+	}
+	if got := p.DownloadPath(); !strings.HasSuffix(got, "krew-downloads") {
```
lbb: What do you test for?

ahmetb: DownloadPath() doesn't return some empty string and we can tell the suffix?

pkg/environment/environment_test.go
```diff
@@ -15,13 +15,57 @@
...
 
 import (
+	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 )
 
+func TestMustGetKrewPaths_resolvesToHomeDir(t *testing.T) {
+	home := os.Getenv("USERPROFILE")
```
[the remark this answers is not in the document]
ahmetb: I partially agree but I'll be adding an env-var override soon for this. That's when this will actually be useful.

Also it's an independent way of finding homedir, so nothing wrong with that :) It's a small test.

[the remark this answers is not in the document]
lbb: Even though this is a small test, please don't test/recreate the homedir.Homedir() logic. We won't use this code even with an env-var override.

pkg/environment/environment_test.go
```diff
@@ -15,13 +15,57 @@
...
+	p := MustGetKrewPaths()
+	if got := p.BasePath(); got != expectedBase {
+		t.Fatalf("MustGetKrewPaths()=%s; expected=%s", got, expectedBase)
+	}
+}
+
+func TestPaths(t *testing.T) {
+	base := filepath.FromSlash("/foo")
+	p := newPaths(base)
+	if got := p.BasePath(); got != base {
```
lbb: When comparign paths we should compare the clean paths: `filepath.Clean(got) != filepath.Clean(want)`

ahmetb: Hmm why? If a=b, then f(a)=f(b), too. And right now tests are passing for a=b.

lbb: We can have the case that `"foo/bar" != "foo/bar/"` but it should be, but
`filepath.Clean("foo/bar") == filepath.Clean("foo/bar/")`.

ahmetb: I prefer this method not return a trailing slash. Adding this makes the test more forgiving and there’s no reason to be more forgiving now.

PR #64, opened 2018-09-26
pkg/env: add tests for ResolveLink and rename

ahmetb:
- Changing ResolveLink to Realpath as it sounds more like a readlink/realpath
  operation (e.g. it works on directories and regular files).
- Add unit tests.
- Refusing to run on _relative_ symlinks (add error).
- Use filepath.Clean.

### DIFF
--- a/cmd/krew/cmd/root.go
+++ b/cmd/krew/cmd/root.go
@@ -65,7 +65,7 @@ func init() {
 	if err != nil {
 		glog.Errorf("failed to get the own executable path")
 	}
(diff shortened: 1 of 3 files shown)


ahmetb:
cc: @lbb

ahmetb:
I think the tests are fine. One of the builds (`/push`) was failing reproducibly by failing to clone the branch (it was trying to clone the wrong repo I think).

ahmetb:
thanks for review, merging. It seems like we should probably delete or unexport this method at some point. It's only used to determine the version of krew itself.

PR #65, opened 2018-10-05
pkg/env: allow base path override with KREW_ROOT

### LINKED ISSUE #58: Allow configuration of krew base path ($HOME/.krew)
#49 adds a hardcoded krew base path to `$HOME/.krew`.

It should be possible to override this path for reasons like:
1. user doesn't want to install binaries in $HOME (maybe they can do /usr/local/krew)
2. user/maintainer wants to check out a command on a fresh krew base path like `mktemp -d` or `/tmp/some-test` without causing any disturbance to their main krew installation.

A very simple way to do this appears to be environment variable, such as:

```sh
KREW_BASE="$(mktemp -d)" kubectl krew install foo
KREW_BASE="$HOME/work-profile/krew" krew install foo
```

ahmetb:
Introduces a KREW_ROOT environment variable to allow overriding
the krew base installation directory (defaults to $HOME/.krew).

Updated installation instructions to add default path to bashrc/zshrc.

Fixes #58.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ krew is the missing kubectl plugin manager.
 ## What is krew?
 
 krew is a tool that makes it easy to install
-[kubectl plugins](https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/). 
+[kubectl plugins](https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/).
 krew helps you discover plugins, install and manage them on your machine. It is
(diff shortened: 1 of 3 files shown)


ahmetb:
cc: @lbb 
I think we should either go with KREW_ROOT or KREW_BASE. 
KREW_PATH kind of means PATH (a directory that has executables in it). In our case, the executables are at $VAR/.bin.

lbb:
LGTM, I lie `KREW_ROOT`

PR #67, opened 2018-10-07
Allow mocking of url downloads with local file (--archive)

ahmetb:
- rename --source to --manifest in install command.
- introduce --archive flag to install command to override the installation from
    the uri specified in the manifest file.
- remove getFileFromArg() method (not needed)

This patch allows specifying both plugin yamls and archives to install cmd:

    krew install --manifest foo.yaml --archive foo.zip

It does not download a file, but it still checks integrity (so it can be used to
check the manifest without uploading the archive anywhere).

The method signatures got very long but I think that's partially because
we probably didn't design the methods in install.go well (the code smell here
is the repetitive args passed from one method to another all the way down).

### DIFF
--- a/cmd/krew/cmd/install.go
+++ b/cmd/krew/cmd/install.go
@@ -18,12 +18,11 @@ import (
 	"bufio"
 	"fmt"
 	"os"
-	"path/filepath"
 
+	"github.com/GoogleContainerTools/krew/pkg/index"
 	"github.com/GoogleContainerTools/krew/pkg/index/indexscanner"
 	"github.com/GoogleContainerTools/krew/pkg/installation"
 
-	"github.com/GoogleContainerTools/krew/pkg/index"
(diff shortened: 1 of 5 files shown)


ahmetb:
@lbb  wdyt of this

ahmetb:
Ready for review.

cmd/krew/cmd/install.go
```diff
@@ -32,7 +31,7 @@ import (
 
 func init() {
 	var forceHEAD *bool
-	var manifest *string
+	var manifest, forceDownloadFile *string
```
lbb: IMHO `forceXXXXX` does sound like a boolean value. Can we maybe name it `downloadOverwrite`.

ahmetb: I'll keep it as is. Feel free to send PR.

PR #72, opened 2018-10-08
pkg/index: add manifest apiVersion check

ahmetb:
We currently do not check apiVersion in plugin manifests and fail other
validation steps like "bin field is missing". This is confusing to users.

Adding an apiVersion check. Right now it's an exact match, but it should be ok.
When we can figure out how to parse k8s apiVersion format and compare them,
we can refactor this into isMinAPIversion().

This will require updates to many testdata files next time apiVersion is bumped.

### DIFF
--- a/pkg/index/validate.go
+++ b/pkg/index/validate.go
@@ -21,10 +21,17 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	currentAPIVersion = "krew.googlecontainertools.github.com/v1alpha2"
+)
+
 var (
 	safePluginRegexp = regexp.MustCompile(`^[\w-]+$`)
+
 	// windowsForbidden is taken from  https://docs.microsoft.com/en-us/windows/desktop/FileIO/naming-a-file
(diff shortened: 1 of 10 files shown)


pkg/index/validate.go
```diff
@@ -20,10 +20,17 @@ import (
 	"strings"
 )
 
+const (
+	currentAPIVersion = "krew.googlecontainertools.github.com/v1alpha2"
```
lbb: split in `groupName` and `groupVersion`: https://kubernetes.io/docs/concepts/overview/kubernetes-api/

pkg/index/validate_test.go
```diff
@@ -67,24 +67,45 @@ func Test_IsSafePluginName(t *testing.T) {
 	}
 }
 
+func Test_isSupportedAPIVersion(t *testing.T) {
+	tests := []struct {
+		name string
+		in   string
+		want bool
+	}{
+		{"wrong prefix", "core/v1", false},
```
lbb: wrong groupName

pkg/index/validate.go
ahmetb: not sure how it will help since we're just doing a `==` check.

this can help do the split but IMO it's not necessary? https://github.com/kubernetes/apimachinery/blob/ed135c5b96450fd24e5e981c708114fbbd950697/pkg/runtime/schema/group_version.go#L216

lbb: I was looking for that :). We can rely on a simple string check for now. We  
 have to change this later (at least after the next index file update).

PR #78, opened 2018-10-09
Convert to go module

ahmetb:
- migrate to go modules from dep
- update vendored dependencies
- deletes packages dep was not able to prune
- force using go modules within GOPATH (b/c travis sets GOPATH) through `GO111MODULE`
- force using vendor/ directory through `GOFLAGS=-mod=vendor`


ahmetb:
Nope this isn't working well.

I'm forcing it to vendor using by prefixing go commands with `env GO111MODULE=on GOFLAGS=-mod=vendor go ...`.

But then `gometalinter.v2 --install` is failing to go-get (because `-mod=vendor` forced). Similarly, if I don't add the envs to gometalinter, then it ends up _ignoring vendor/_ and fetching what's listed in go.mod again.

PR #82, opened 2018-10-10
cmd: refactor table printing code

ahmetb:
- convert 2-column printing code to general-purpose table printer (printTable)
- deduplicate tabwriter calls from list/version/search into printTable
- do not discard errors from printTable in cmds (return from RunE)
- cmd/version: do not sort fields (looks more organized)

### DIFF
--- a/cmd/krew/cmd/list.go
+++ b/cmd/krew/cmd/list.go
@@ -40,12 +40,25 @@ Plugins will be shown as "PLUGIN,VERSION"`,
 			if err != nil {
 				return errors.Wrap(err, "failed to find all installed versions")
 			}
+
+			// return sorted list of plugin names when piped to other commands or file
 			if !(isatty.IsTerminal(os.Stdout.Fd()) || isatty.IsCygwinTerminal(os.Stdout.Fd())) {
-				fmt.Fprintf(os.Stdout, "%s\n", strings.Join(sortedKeys(plugins), "\n"))
+				var names []string
+				for name := range plugins {
+					names = append(names, name)
+				}
+				sort.Strings(names)
+				fmt.Fprintln(os.Stdout, strings.Join(names, "\n"))
 				return nil
 			}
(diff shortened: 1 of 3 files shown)


cmd/krew/cmd/list.go
```diff
@@ -41,41 +40,30 @@ Plugins will be shown as "PLUGIN,VERSION"`,
...
-				fmt.Fprintf(os.Stdout, "%s\n", strings.Join(sortedKeys(plugins), "\n"))
+				for plugin := range plugins {
+					fmt.Fprintln(os.Stdout, plugin)
+				}
 				return nil
 			}
-			printAlignedColumns(os.Stdout, "PLUGIN", "VERSION", plugins)
-			return nil
+			var table [][]string
+			for p, version := range plugins {
```
lbb: The plugins should be sorted here (got removed from the tableWriter).

cmd/krew/cmd/search.go
```diff
@@ -67,9 +65,8 @@ Search accepts a list of words as options.`,
 			return nil
 		}
 
-		rowPattern := "%s\t%s\t%s\n"
-		w := tabwriter.NewWriter(os.Stdout, 0, 0, 1, ' ', 0)
-		fmt.Fprintf(w, rowPattern, "NAME", "DESCRIPTION", "STATUS")
+		var rows [][]string
+		cols := []string{"NAME", "DESCRIPTION", "STATUS"}
 		for _, name := range matchNames {
```
lbb: sort plugins

cmd/krew/cmd/list.go
ahmetb: > The plugins should be sorted here (got removed from the tableWriter).

I removed on purpose, because the method actually returns a sorted list anyway?

ahmetb:
I'll add sorting back. `installation.ListInstalledPlugins` was already returning sorted but it's not a guarantee of the function.

PR #95, opened 2018-10-12
Refactor release scripts

### LINKED ISSUE #89: Uninstall travis releases in favor of gcb
After #88 is merged, delete `deploy` script in travis CI as GCB will upload the release binaries.

That said, we need to set  correctly with the pre-calculated GCS URLs so (1) the Release message on GitHub (2) README links to the correct binary URLs.

ahmetb:
- disable travis github deploys on tags (fixes #89)
- split binary building (can also be used during development) and release
  artifact building (krew.{yaml,zip{,sha256}}). we should probably convert
  this to a makefile sometime.

### DIFF
--- a/hack/make-release-artifacts.sh
+++ b/hack/make-release-artifacts.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+
+# Copyright 2018 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e -o pipefail
+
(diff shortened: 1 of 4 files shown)


hack/make-release-artifacts.sh
```diff
@@ -0,0 +1,49 @@
...
+
+KREW_ARCHIVE="krew.zip"
+
+echo >&2 "Creating ${KREW_ARCHIVE} archive."
+zip -X -q -r --verbose out/krew.zip "${bin_dir}"
+
+if hash sha256sum 2>/dev/null; then
+  checksum_cmd=sha256sum
+else
+  checksum_cmd="shasum -a 256"
```
lbb: can we move this before the if, to have `checksum_cmd` in the scope of the document?

hack/make-release-artifacts.sh
```diff
@@ -0,0 +1,49 @@
...
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd "${SCRIPTDIR}/.."
+
+bin_dir="out/bin"
+if [[ ! -d "${bin_dir}" ]]; then
+    echo >&2 "Binaries are not built (${bin_dir}), run build-binaries.sh"
+    exit 1
+fi
+
+KREW_ARCHIVE="krew.zip"
```
lbb: Why do we write this variable capitalized?

ahmetb: you're right

PR #105, opened 2018-10-14
hack: offer .tar.gz releases, use for posix

### LINKED ISSUE #101: Provide .tar.gz release artifact
In addition to .zip (which is good for windows) we should also have .tar.gz for krew releases.

This would unburden the mac and linux users from installing `unzip` tool, which is not really standard, whereas `tar` is.

ahmetb:
- create krew.tar.gz{,.sha256sum}
- use it in krew.yaml for linux/darwin
- use it in README.md for linux/darwin instalation (fixes #101)

### DIFF
--- a/hack/make-release-artifacts.sh
+++ b/hack/make-release-artifacts.sh
@@ -25,23 +25,36 @@ if [[ ! -d "${bin_dir}" ]]; then
     exit 1
 fi
 
-krew_archive="krew.zip"
-echo >&2 "Creating ${krew_archive} archive."
-zip -X -q -r --verbose out/krew.zip "${bin_dir}"
+krew_tar_archive="krew.tar.gz"
+krew_zip_archive="krew.zip"
+
+echo >&2 "Creating ${krew_tar_archive} archive."
+tar czvf "out/${krew_tar_archive}" "${bin_dir}"
+echo >&2 "Creating ${krew_zip_archive} archive."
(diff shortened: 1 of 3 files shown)


hack/make-release-artifacts.sh
```diff
@@ -25,23 +25,36 @@ if [[ ! -d "${bin_dir}" ]]; then
...
+zip_checksum="$(eval "${checksum_cmd[@]}" "out/${krew_zip_archive}" | awk '{print $1;}')"
+echo >&2 "${krew_zip_archive} checksum: ${zip_checksum}"
+echo "${zip_checksum}" > "${zip_sumfile}"
+echo >&2 "Written ${zip_sumfile}."
+
 
 # Copy and process krew manifest
 cp ./hack/krew.yaml ./out/krew.yaml
 tag="$(git describe --tags --always HEAD)"
 sed -i "s/KREW_ZIP_CHECKSUM/${zip_checksum}/g" ./out/krew.yaml
```
lbb: Do we want to keep the zip functionality?

lbb: It's currently not used

ahmetb: > Do we want to keep the zip functionality?

Yes, for windows release. If you're asking about krew.zip.sha256 file:

- It's just good symmetry to keep the file in releases
- It's used KREW_ZIP_CHECKSUM is used in krew.yaml.

PR #108, opened 2018-10-14
Remove archive indirection

### LINKED ISSUE #90: Remove directory indirection from krew.zip
Right now krew.zip has the following structure

```
unzip -l out/krew.zip
drwxr-xr-x  3.0 unx        0 b- stor 18-Oct-10 23:17 out/bin/
-rwxr-xr-x  3.0 unx 11682304 b- defN 18-Oct-10 23:17 out/bin/krew-windows_amd64.exe
-rwxr-xr-x  3.0 unx 11716104 b- defN 18-Oct-10 23:17 out/bin/krew-darwin_amd64
-rwxr-xr-x  3.0 unx 10214825 b- defN 18-Oct-10 23:17 out/bin/krew-linux_arm
-rwxr-xr-x  3.0 unx 11828176 b- defN 18-Oct-10 23:17 out/bin/krew-linux_amd64
```

The `out/` prefix in the .zip yaml is unnecessary. I'm even thinking `bin/` is redundant, too. 

We can keep `bin/` in case we add other stuff to the release someday, but I think it doesn't matter that much.

Fixing this would
- make copy operations in hack/krew.yaml shorter
- make installation instructions look shorter/cleaner

/cc @lbb; wdyt?

lbb:
/fixes #90

### DIFF
--- a/hack/make-release-artifacts.sh
+++ b/hack/make-release-artifacts.sh
@@ -29,9 +29,9 @@ krew_tar_archive="krew.tar.gz"
 krew_zip_archive="krew.zip"
 
 echo >&2 "Creating ${krew_tar_archive} archive."
-pushd "${bin_dir}"; tar czvf "${SCRIPTDIR}/../out/${krew_tar_archive}" ./*; popd
+( cd "${bin_dir}"; tar czvf "${SCRIPTDIR}/../out/${krew_tar_archive}" ./*; )
 echo >&2 "Creating ${krew_zip_archive} archive."
-zip -X -q -j --verbose "out/${krew_zip_archive}" ${bin_dir}/*
+( cd "${bin_dir}"; zip -X -q --verbose "${SCRIPTDIR}/../out/${krew_zip_archive}" ./*; )
 
 checksum_cmd="shasum -a 256"
 if hash sha256sum 2>/dev/null; then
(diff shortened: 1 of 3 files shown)


lbb:
Addressed

hack/make-release-artifacts.sh
```diff
@@ -29,9 +29,9 @@ krew_tar_archive="krew.tar.gz"
 krew_zip_archive="krew.zip"
 
 echo >&2 "Creating ${krew_tar_archive} archive."
-tar czvf "out/${krew_tar_archive}" "${bin_dir}"
+pushd "${bin_dir}"; tar czvf "${SCRIPTDIR}/../out/${krew_tar_archive}" ./*; popd
```
ahmetb: prefer syntax:

```sh
(
   cd $dir
   command
)
```

hack/make-release-artifacts.sh
```diff
@@ -29,9 +29,9 @@ krew_tar_archive="krew.tar.gz"
...
 echo >&2 "Creating ${krew_zip_archive} archive."
-zip -X -q -r --verbose "out/${krew_zip_archive}" "${bin_dir}"
+zip -X -q -j --verbose "out/${krew_zip_archive}" ${bin_dir}/*
```
ahmetb: I think it would be better if we do the above syntax (tar) for consistency instead of adding `-j`

PR #127, opened 2018-10-19
pkg/download: support nested files without dir entries

### LINKED ISSUE #126: Can't figure out failure
I am not sure what failure I am experiencing with this plugin.  Errors seem to be concatenated against each other.
https://github.com/jasonrichardsmith/krew-index/blob/initial-submission/plugins/rbac-view.yaml
```
$ kubectl krew install --manifest=rbac-view.yaml
Installing plugin: rbac-view
W1019 10:03:04.251654   14926 install.go:132] failed to install plugin "rbac-view": failed to dowload and move during installation: failed to download and verify file: failed to create file "/tmp/krew-downloads/rbac-view/bin/linux/rbac-view": open /tmp/krew-downloads/rbac-view/bin/linux/rbac-view: no such file or directory
F1019 10:03:04.251778   14926 root.go:52] failed to install some plugins: [rbac-view]
```

ahmetb:
Regular tar entries without corresponding directory entires are valid, but
could not be extracted because we currently don't create the directory the file
will reside in.

Also adding a new test file to reflect the case of nested files without dir
entries.

Fixes #126.


ahmetb:
Skipping code review since trivial bug fix.

PR #132, opened 2018-12-07
Fix missing krew subcommand in README.md

superbrothers:

PR #133, opened 2018-12-12
Remove the binary

superbrothers:
I made a mistake that commits the binary... I'm really sorry.

/ref #132


ahmetb:
I'll amend the commit and force-push to the master. Otherwise that binary will always be pulled down on a git-clone, making the clone bigger forever.

PR #137, opened 2018-12-24
Use vgo instead of dep

lbb:
This PR moves from the vendor system dep to go modules (vgo) https://github.com/golang/go/wiki/Modules#quick-start


ahmetb:
We probably want to stick to what kubectl does (not vgo).

Issue #45, opened 2018-09-19
Support direct binary (non-archive) downloads


ahmetb:
I've been thinking about this for a while.

Especially with kubectl 1.12 plugin model, it's inevitable people will have Go repositories on Github with static binary releases (without ZIPs).

I've so far heard from at least one plugin dev that `goreleaser` doesn't support creating .zip, so they need to change their release model to accommodate krew distribution.

#### Strawman design:

We can have 3 modes, like:

<table>
<tbody>
<tr>
<td>
archive downloads:
<pre>
download:
  archive: https://.../foo.zip
  sha256: [...]
</pre>
</td>
<td>
file downloads:
<pre>
download:
  file: https://.../kubectl-foo_linux_amd64
  sha256: [...]
</pre>
</td>
<td>
head:
<pre>
download:
  head: [...]
</pre>
</td>
</tr>
</tbody>
</table>

Thoughts? There's time to do this during for krew 0.2.

cc: @lbb

lbb:
We choose ZIP, as it is the default format GitHub offers to download (although tar.gz is also available).

If we choose to offer different download formats we should not use different classifiers to specify an URI. We should focus on MIME types as UNIX does it. We also should fall back to files post-fixes.

ahmetb:
> We choose ZIP, as it is the default format GitHub offers to download

This is not a good argument and only works for python/bash scripts etc. Not to mention, github supports .tar.gz, too?



> we should not use different classifiers
> We also should fall back to files post-fixes.
> We should focus on MIME types as UNIX does it.

All these 3 are basically just "do not be verbose and employ magic", which I'm against.

Good software rarely tries to guess things (esp. from stuff like mime types or extensions).

lbb:
1. In case we want to be explicit by writing the filetype of the resource at the URI:
   If we do so, we should not make it a field in the plugin spec like `file: 
 https://..`, `archive: https://...`. Adding fields is not flexible:
  * We have to add a new field for every compression type. -> It's a spec version bump on every new field.
  * Developers can declare multiple URI's at once.

  I propose to add a new text field `downloadType` which specifies the download format.


2. We only have a few types we want to support (.tar, tar.gz, zip and plain). All of 
  those types are identifiable by a magic number that is part of it's standard.
  MIME detection is not magic; it's just detecting the file header bytes.

I'm not for employing magic. We **know** what type it is after we have downloaded it. We are not guessing.

ahmetb:
[...]

PR #139, opened 2018-12-25
Use variable extractor methods for mime types

### LINKED ISSUE #128: Use file signatures to detect archive types
We currently rely on .zip and .tar.gz extension of downloaded files to choose extraction algorithm. File magic numbers can help here better and make it possible to use plugin archives without explicit file extension.

lbb:
/fixes #128


ahmetb:
-1 for checking in 1 MB blob to test that it returns octet/stream. I'm not even sure why we're trying to check octet/stream.

pkg/download/downloader.go
```diff
@@ -162,20 +163,44 @@ func GetInsecure(uri, dir string, fetcher Fetcher) error {
...
+	}
+	return http.DetectContentType(buf[:n]), nil
+}
+
+type extractor func(targetDir string, read io.ReaderAt, size int64) error
+
+var defaultExtractors = map[string]extractor{
+	"application/zip":      extractZIP,
+	"application/tar+gzip": extractTARGZ,
+	"application/x-gzip":   extractTARGZ,
```
ahmetb: gzip != tar.

Not sure if this will work. I recommend removing and not supporting unless there's a good reason.

pkg/download/downloader_test.go
```diff
@@ -140,3 +147,132 @@ func collectFiles(t *testing.T, scanPath string) []string {
...
+		{
+			name: "type tar.gz",
+			args: args{
+				file: filepath.Join(testdataPath(), "test-with-nesting-with-directory-entries.tar.gz"),
+			},
+			want:    "application/x-gzip",
+			wantErr: false,
+		},
+		{
+			name: "type elf-bin",
```
ahmetb: this test seems unnecessary to me.

pkg/download/downloader_test.go
```diff
@@ -140,3 +147,132 @@ func collectFiles(t *testing.T, scanPath string) []string {
...
+			args: args{
+				file: filepath.Join(testdataPath(), "elf-file"),
+			},
+			want:    "application/octet-stream",
+			wantErr: false,
+		},
+		{
+			name: "type null",
```
ahmetb: ditto. not sure why we're testing text/plain.

pkg/download/downloader.go
```diff
@@ -89,8 +89,9 @@ func extractZIP(targetDir string, read io.ReaderAt, size int64) error {
 }
 
 // extractTARGZ extracts a gzipped tar file into the target directory.
-func extractTARGZ(targetDir string, in io.Reader) error {
+func extractTARGZ(targetDir string, at io.ReaderAt, size int64) error {
 	glog.V(4).Infof("tar: extracting to %q", targetDir)
+	in := io.NewSectionReader(at, 0, size)
```
ahmetb: IMO this change is unnecessary, you should just ignore `size` by naming it as `_`. We're never not read a tar file from a reader's only first N bytes for any reason.

pkg/download/downloader.go
```diff
@@ -162,20 +163,44 @@ func GetInsecure(uri, dir string, fetcher Fetcher) error {
 	return extractArchive(name, dir, body, size)
 }
 
-func extractArchive(filename, dst string, r io.ReaderAt, size int64) error {
+func extractContentType(at io.ReaderAt) (string, error) {
```
ahmetb: call this `detectContentType` or `detectMIMEType`. it's not extracting anything.

also rename `at` to `r`. (ditto in the other method)

pkg/download/downloader.go
```diff
@@ -162,20 +163,44 @@ func GetInsecure(uri, dir string, fetcher Fetcher) error {
...
+}
+
+func extractArchive(filename, dst string, at io.ReaderAt, size int64) error {
+	// TODO: Keep the filename for later direct download
+
 	// TODO(ahmetb) This package is not architected well, this method should not
 	// be receiving this many args. Primary problem is at GetInsecure and
 	// GetWithSha256 methods that embed extraction in them, which is orthogonal.
 
 	// TODO(ahmetb) write tests with this by mocking extractZIP function into a
```
ahmetb: I think we can remove this particular todo. There's tests for tar, zip extract methods now.

pkg/download/downloader.go
ahmetb: Not all gzip uncompress into tarballs though.

GitHub returns `application/x-gzip` for tarballs, but https://superuser.com/questions/901962/what-is-the-correct-mime-type-for-a-tar-gz-file says `application/gzip` would be recommended out in the wild. 🤔 

I guess we can take a risk and assume all GZIPs given to krew will be tarballs and not .zip.gz or something. :)

pkg/download/downloader.go
```diff
@@ -162,20 +163,44 @@ func GetInsecure(uri, dir string, fetcher Fetcher) error {
...
+type extractor func(targetDir string, read io.ReaderAt, size int64) error
+
+var defaultExtractors = map[string]extractor{
+	"application/zip":      extractZIP,
+	"application/tar+gzip": extractTARGZ,
+	"application/x-gzip":   extractTARGZ,
+}
+
+func extractArchive(filename, dst string, at io.ReaderAt, size int64) error {
+	// TODO: Keep the filename for later direct download
```
ahmetb: not sure what this todo says, consider adding some context/username.

pkg/download/downloader.go
```diff
@@ -162,20 +163,44 @@ func GetInsecure(uri, dir string, fetcher Fetcher) error {
...
-		glog.V(4).Infof("detected .tar.gz file")
-		return extractTARGZ(dst, io.NewSectionReader(r, 0, size))
+	t, err := extractContentType(at)
+	if err != nil {
+		return errors.Wrap(err, "failed to determine content type")
+	}
+	glog.V(4).Infof("detected %q file type", t)
+	exf, ok := defaultExtractors[t]
+	if !ok {
+		return errors.Errorf("cannot infer a supported archive type mime: (%s)", t)
```
ahmetb: () syntax unnecessary. also message isn't very meaninful just say.

`mime type %q for downloaded file is not a supported archive format`

pkg/download/downloader.go
```diff
@@ -162,20 +163,44 @@ func GetInsecure(uri, dir string, fetcher Fetcher) error {
...
 	}
-	return errors.Errorf("cannot infer a supported archive type from filename in the url (%q)", filename)
+	return errors.Wrap(exf(dst, at, size), "failed to extract file")
```
ahmetb: I'm a bit confused about this `io.ReaderAt` here.

- This method gets `at io.ReaderAt` which this method doesn't know where it starts.
- `at` is passed to `detectContentType`
- `detectContentType` does `at.ReadAt(buf, 0)` which I'm not sure is a good idea. We could just pass this method an `io.Reader` or `[]byte` and have a cleaner design.
- `detectContentType` caused the current `Read()` cursor to advance (?) as a side-effect.
- we pass the same `at` to `exf(dst,at,size)`.
- now the extractZIP or extractTARGZ function has received an already-advanced reader. so they **have to** use a section reader (they previously weren't) to set the cursor back to zero.

Instead, I wish we could just 
- read first 512 bytes into []byte
- pass []byte to detectContentType()
- reset the `at` to cursor position = 0.
- pass `at` down so the downstream extractors (extractZIP etc) don't need to explicitly say "start reading from 0" and they can just assume it's already at zero (=unconsumed reader).

pkg/download/downloader_test.go
lbb: We need this later in #45 . This verifies the content type of direct downloads. I'm not sure if this testing should be done by us.
We assume that a binary file gets the type `application/octet-stream`. This test makes sure that our assumptions are correct.

pkg/download/downloader.go
lbb: It seems like Go only detects `application/x-gzip` and `application/zip` https://golang.org/src/net/http/sniff.go#L147

pkg/download/downloader_test.go
lbb: This is for checking the behavior with 0-byte files, which could be an edge case.

ahmetb:
I think it's fine I'll probably just merge this as is.

pkg/download/downloader.go
```diff
@@ -142,22 +144,45 @@ func extractTARGZ(targetDir string, in io.Reader) error {
 	return nil
 }
 
-func extractArchive(filename, dst string, r io.ReaderAt, size int64) error {
+func detectMIMEType(at io.ReaderAt) (string, error) {
```
ahmetb: uh oh I just realized there's no tests for this one :) 

can you add a test with simple strings.NewReader like:

- 0 bytes
- <512 bytes
- >512 bytes
- html content(?)

I have a suspicion that buf you're allocating below has 512 bytes length, so if you're reading only 12 bytes, you have 500 bytes of \0 –which may result in wrong results?

lbb: mime type matching has some edge cases:
* 0 0-bytes  ->  `text/plain`
* 1+ 0-bytes -> `application/octet-stream`
This should not be a problem for us, as there are no 0 byte zip/targz files.

PR #141, opened 2018-12-31
hack: consistent checksums for archive files

### LINKED ISSUE #112: archive checksums aren't consistent
If I run `hack/make-all.sh` twice, I get different checksums. e.g.

    krew.tar.gz checksum: 86fff449a95637432c28592b530077a449fe578acc296aacbcc9a7e71ef83126
    krew.zip checksum: 08aa780dc8cc702c553fb622f334b8439b537c8679309680feb5b2d462b3b2ed

and

	krew.tar.gz checksum: 0233e2af055875784db833083d56fc235e09a085f2bcc1624598ab03a8bf43cd
	krew.zip checksum: 8bb9df9aef534442c9d7f5b05c3112ae7cf20f4b5ec54b28c1e5c60b11f47482

ahmetb:
Use "gzip -n" to prevent gzip from changing timestamps and touch all timestamps
for files in out/bin with a static value to ensure .zip checksum consistency.

Fixes #112.

### DIFF
--- a/hack/make-release-artifacts.sh
+++ b/hack/make-release-artifacts.sh
@@ -28,8 +28,16 @@ fi
 krew_tar_archive="krew.tar.gz"
 krew_zip_archive="krew.zip"
 
+# consistent timestamps for files in bindir to ensure consistent checksums
+while IFS= read -r -d $'\0' f
+do
+  echo "modifying atime/mtime for $f"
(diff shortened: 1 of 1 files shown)


ahmetb:
Rebuilding a travis build produced consistent checksums, but ideally we'd need to pin to a go compiler/runtime version, too. I don't see a need for that yet.

PR #158, opened 2019-03-07
Add `spec.homepage` field to the plugin manifest

### LINKED ISSUE #157: Proposal: Add Homepage, License and Author fields to Plugin
I propose new three fields for Plugin.

- `spec.homepage`:  *Optional* project homepage, only for documentation purposes
    - Motivation: When I find bugs in a plugin, I’d like to know the repository of the plugin easily in order to send an issue.
- `spec.license`: *Optional* license such as Apache-2.0 or MIT, only for documentation purposes
    - Motivation: When I find a plugin as it seems convenient, I’d like to know the license of the plugin easily.
- `spec.author`: *Optional* project author, only for documentation purposes
    - Motivation: When I find an awesome plugin, I’d like to know the author of the plugin easily.

Along with adding the new fields, `krew info` command will be changed to show homepage, license, and author as follows:

```
$ kubectl krew info krew
NAME: krew
URI: https://storage.googleapis.com/krew/v0.2.1/krew.tar.gz
SHA256: dc2f2e1ec8a0acb6f3e23580d4a8b38c44823e948c40342e13ff6e8e12edb15a
VERSION: v0.2.1
HOMEPAGE: https://github.com/GoogleContainerTools/krew
LICENSE: Apache-2.0
AUTHOR: https://github.com/ahmetb
CAVEATS:
\
 |  krew is now installed! To start using kubectl plugins, you need to add
 |  krew's installation directory to your PATH:
 |    * macOS/Linux:
 |      - Add the following to your ~/.bashrc or ~/.zshrc:
 |          export PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH"
 |      - Restart your shell.
 |    * Windows: Add %USERPROFILE%\.krew\bin to your PATH environment variable
 |  Run "kubectl krew" to list krew commands and get help.
 |  You can find documentation at https://github.com/GoogleContainerTools/krew.
/
```

The below is `krew.yaml` which is added the new three fields:

```diff
diff --git a/plugins/krew.yaml b/plugins/krew.yaml
index dca2408..e811ac0 100644
--- a/plugins/krew.yaml
+++ b/plugins/krew.yaml
@@ -19,6 +19,9 @@ metadata:
 spec:
   version: "v0.2.1"
   shortDescription: Package manager for kubectl plugins.
+  homepage: https://github.com/GoogleContainerTools/krew
+  license: Apache-2.0
+  author: https://github.com/ahmetb
   caveats: |
     krew is now installed! To start using kubectl plugins, you need to add
     krew's installation directory to your PATH:
```

https://github.com/superbrothers/krew/tree/homepage-license-author is PoC of this proposal.

superbrothers:
This PR adds `spec.homepage` field to the plugin manifest.

/close #157 

---

The `spec.homepage` field of the plugin manifest defines the url of the
plugin project homepage. The field value should be a valid url.

In addition, if the `spec.homepage` field is defined in plugin manifest,
`kubectl krew info` command displays that field value as HOMEPAGE as follows:

    $ kubectl krew info krew
    NAME: krew
    URI: https://storage.googleapis.com/krew/v0.2.1/krew.tar.gz
    SHA256: dc2f2e1ec8a0acb6f3e23580d4a8b38c44823e948c40342e13ff6e8e12edb15a
    VERSION: v0.2.1
    HOMEPAGE: https://github.com/GoogleContainerTools/krew
    ...

### DIFF
--- a/cmd/krew/cmd/info.go
+++ b/cmd/krew/cmd/info.go
@@ -68,14 +68,14 @@ func printPluginInfo(out io.Writer, plugin index.Plugin) {
 			fmt.Fprintf(out, "SHA256: %s\n", platform.Sha256)
 		}
 	}
-	if plugin.Spec.Description != "" {
-		fmt.Fprintf(out, "DESCRIPTION: \n%s\n", plugin.Spec.Description)
-	}
 	if plugin.Spec.Version != "" {
 		fmt.Fprintf(out, "VERSION: %s\n", plugin.Spec.Version)
 	}
 	if plugin.Spec.Homepage != "" {
 		fmt.Fprintf(out, "HOMEPAGE: %s\n", plugin.Spec.Homepage)
+	}
+	if plugin.Spec.Description != "" {
+		fmt.Fprintf(out, "DESCRIPTION: \n%s\n", plugin.Spec.Description)
 	}
 	if plugin.Spec.Caveats != "" {
 		fmt.Fprintln(out, prepCaveats(plugin.Spec.Caveats))
--- a/cmd/krew/cmd/info.go
+++ b/cmd/krew/cmd/info.go
@@ -74,6 +74,9 @@ func printPluginInfo(out io.Writer, plugin index.Plugin) {
 	if plugin.Spec.Version != "" {
 		fmt.Fprintf(out, "VERSION: %s\n", plugin.Spec.Version)
 	}
+	if plugin.Spec.Homepage != "" {
+		fmt.Fprintf(out, "HOMEPAGE: %s\n", plugin.Spec.Homepage)
+	}
 	if plugin.Spec.Caveats != "" {
 		fmt.Fprintln(out, prepCaveats(plugin.Spec.Caveats))
 	}
--- a/pkg/index/validate.go
+++ b/pkg/index/validate.go
@@ -15,6 +15,7 @@
 package index
 
 import (
+	"net/url"
 	"regexp"
 	"strings"
 
@@ -72,6 +73,11 @@ func (p Plugin) Validate(name string) error {
 	for _, pl := range p.Spec.Platforms {
 		if err := pl.Validate(); err != nil {
 			return errors.Wrapf(err, "platform (%+v) is badly constructed", pl)
+		}
+	}
+	if p.Spec.Homepage != "" {
+		if _, err := url.ParseRequestURI(p.Spec.Homepage); err != nil {
+			return errors.Wrapf(err, "the homepage url %q is not a valid url", p.Spec.Homepage)
 		}
 	}
 	return nil
--- a/pkg/index/validate.go
+++ b/pkg/index/validate.go
@@ -15,7 +15,6 @@
 package index
 
 import (
-	"net/url"
 	"regexp"
 	"strings"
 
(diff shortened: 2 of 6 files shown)


pkg/index/validate.go
```diff
@@ -74,6 +75,11 @@ func (p Plugin) Validate(name string) error {
 			return errors.Wrapf(err, "platform (%+v) is badly constructed", pl)
 		}
 	}
+	if p.Spec.Homepage != "" {
```
ahmetb: don't validate this for now.

1. we don't get anything out of it
2. ParseRequestURI is **only** meant to be used for parsing URLs that came in a request (not user input).

cmd/krew/cmd/info.go
```diff
@@ -74,6 +74,9 @@ func printPluginInfo(out io.Writer, plugin index.Plugin) {
 	if plugin.Spec.Version != "" {
 		fmt.Fprintf(out, "VERSION: %s\n", plugin.Spec.Version)
 	}
+	if plugin.Spec.Homepage != "" {
```
ahmetb: let's reorder this as

- name
- version
- homepage
- description
- caveats

what do you think?

superbrothers: OK. I'll change the order as you said.

pkg/index/validate.go
superbrothers: If we want the URL to be used in the homepage field, I think we should better validate the value. But I don't disagree strongly, so I'll remove validating code.

ahmetb: Does url.Parse give enough to you?

superbrothers: Yes, it does.

superbrothers: Not related to this PR, but the `shortDescription` field seems that does not be contained in the output fields in spite of required field.

cmd/krew/cmd/info.go
superbrothers: Not related to this PR, but the shortDescription field seems that does not be contained in the output fields in spite of required field.

pkg/index/validate_test.go
```diff
@@ -240,6 +241,54 @@ func TestPlugin_Validate(t *testing.T) {
 			pluginName: "../foo",
 			wantErr:    true,
 		},
+		{
+			name: "valid homepage url",
+			fields: fields{
+				TypeMeta:   metav1.TypeMeta{APIVersion: currentAPIVersion},
+				ObjectMeta: metav1.ObjectMeta{Name: "foo"},
+				Spec: PluginSpec{
+					Version:          "",
```
ahmetb: can you omit the fields irrelevant to this test, such as this one.
we should strive to keep the test cases as minimal as possible.

pkg/index/validate.go
superbrothers: > Does url.Parse give enough to you?

I'm sorry, actualy not enough :bow:. Because `url.Parse` does not return error with value `this-is-not-a-valid-url`, so I think this validating code has not meaning.

 I removed the code by 32c6d76.

superbrothers:
Should I squash this PR?

ahmetb:
Also home page is  part of documentation technically. It's not good to fail parsing the index during an upgrade because some homepage URL was invalid. I think this is better.

PR #161, opened 2019-03-14
Add OWNERS, kubernetes project template files

ahmetb:
https://github.com/kubernetes/kubernetes-template-project

/cc @juanvallejo @nikhita

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,28 +1,29 @@
-# How to Contribute
+# Contributing Guidelines
 
-We'd love to accept your patches and contributions to this project. There are
-just a few small guidelines you need to follow.
+Welcome to Kubernetes. We are excited about the prospect of you joining our [community](https://git.k8s.io/community)! The Kubernetes community abides by the CNCF [code of conduct](code-of-conduct.md). Here is an excerpt:
 
-## Contributor License Agreement
+_As contributors and maintainers of this project, and in the interest of fostering an open and welcoming community, we pledge to respect all people who contribute through reporting issues, posting feature requests, updating documentation, submitting pull requests or patches, and other activities._
 
-Contributions to this project must be accompanied by a Contributor License
-Agreement. You (or your employer) retain the copyright to your contribution;
-this simply gives us permission to use and redistribute your contributions as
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -1,28 +1,31 @@
...
-Guidelines](https://opensource.google.com/conduct/).
\ No newline at end of file
+- [Mentoring Initiatives](https://git.k8s.io/community/mentoring) - We have a diverse set of mentorship programs available that are always looking for volunteers!
+
+<!---
+Custom Information - if you're copying this template for the first time you can add custom content here, for example:
+
+## Contact Information
+
+- [Slack channel](https://kubernetes.slack.com/messages/kubernetes-users) - Replace `kubernetes-users` with your slack channel string, this will send users directly to your channel.
```
nikhita: Maybe it would be good to add the sig-cli channel and mailing list here? :)

ahmetb: Done. I reckon we'll add way more stuff about SIG CLI post-migration.

OWNERS_ALIASES
```diff
@@ -0,0 +1,4 @@
+aliases:
+  krew-index-maintainers:
```
rlenferink: I think the `krew-maintainers` and `krew-index-maintainers` were swapped by accident. 

Probably this one should be `krew-maintainers` and this one [8888595ca25a7a48c303fbc07f369b2db9b95268/OWNERS_ALIASES#L2](https://github.com/GoogleContainerTools/krew-index/blob/8888595ca25a7a48c303fbc07f369b2db9b95268/OWNERS_ALIASES#L2) should be `krew-index-maintainers`

ahmetb: Argh my bad. I think I’ll do both krew-maintainers if that makes it easier. I suspect we’ll revise them in the future again.

rlenferink: At the moment, it doesn't really matter. The teams in the OWNERS files need to be defined in the OWNERS_ALIASES file _per repository_. Till the moment we can use GitHub teams for this, it is necessary to have the data available in the OWNERS* files for both repositories.

So it is completely up to you if you want to 'combine' or keep the repositories separate (different contributors/maintainers).

PR #171, opened 2019-04-18
Update README.md

axklim:
I have some trouble with OSX 10.14.4 because there is bash 3.x by default and has not built in command mapfile. https://github.com/bminor/bash/blob/bash-4.0/CHANGES#L573
I repaired it with `brew install bash` :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,9 @@ Read the [User Guide](./docs/USER_GUIDE.md) for detailed documentation.
 
 ### Installation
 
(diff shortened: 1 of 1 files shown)


ahmetb:
/hold

Do you mind instead opening an issue with the error you're seeing on bash 3.x?
I suspect we can easily have a backwards-compatible version of the installation script, and I'd prefer to do that instead of adding yet another warning.

axklim:
@ahmetb No problem, https://github.com/kubernetes-sigs/krew/issues/172
:)

PR #176, opened 2019-05-08
Switch linter from gometalinter to golangci-lint

### LINKED ISSUE #174: Switch linter to golangci-lint
Krew currently uses gometalinter. Gometalinter has been deprecated by his author in favor of `golangci-lint`, therefore krew should switch to the better linter.

corneliusweig:
Close #174
Based on #178


corneliusweig:
@ahmetb done

ahmetb:
Thanks for proposing this. I feel a bit uneasy about `hack/install_golangci-lint.sh` which is auto-generated, long and doesn't seem maintainable. 

I think we shouldn't use megalinters like this one either. I'm not sure if the value add is enough.

I'm trying to understand what other kubernetes projects use. Here's a search query: https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/kubernetes+%5Cblint+-file:vendor

I suspect falling back to `x/lint` and `go vet` for now would suffice. What do you think of this?

ahmetb:
x/lint seems fairly popular https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/kubernetes+%22golint+%22+-file:vendor+count:1000

corneliusweig:
I'm not sure I understand your concerns about a metalinter, after all `gometalinter` was one too. I understand your worries about the generated script, though. I had another look at the `golangci-lint` docs and they offer an install script for CI environments. For example, kustomize also uses that https://github.com/kubernetes-sigs/kustomize/blob/540e4023dae93dc9d8fcab37b18734ef9df08339/.travis.yml#L31

Would that address your worries?

corneliusweig:
However, this would mean that users can't run the linter locally without installing golangci-lint manually.

ahmetb:
The part that annoyed me slightly is why there's a need for a very long general-purpose script (`install_golangci-lint.sh`) just to install a go binary?

Can't we just do wget+untar? I know their readme says "use this script" but I don't see the point just yet.

corneliusweig:
Well I think the major argument for this script is convenience for users with different platforms and architectures. The script conveniently downloads the correct build for the current platform. As a new contributor, I would find it very nice if the linter would just work without having to manually download that binary from wherever that is.
Another issue is consistency of linter warnings. The reported issues _do_ change slightly with each version. Pinning the linter version avoids confusion here.

Admittedly, the script is large. But it does have a purpose. (Plus it also checks the hashsum)

ahmetb:
Yeah but we could write our own install script and it would download if the binary is not in $PATH from the pinned version. I suspect the script would be very short and maintainable if we wrote it. 

Having wget/tar installed is a sensible expectation from any contributor so I’m not too worried about it. We already use sha256sum tooling in other scripts.

corneliusweig:
Honestly, if we want to support contributors on different platforms, this script is the best way to go. After all, it's been tried and tested.

I feel you major concern is having the script in the repo. Would it be an option to run it as `curl http:... | sh -`?

pkg/version/version.go
```diff
@@ -15,8 +15,10 @@
 // Package version contains the version information of the krew binary.
 package version
 
+const unknown = "unknown"
```
ahmetb: I don't know why this is a linter error. It’s pretty common coding practice to have package level-`const`s that are used throughout the code rather than duplicating it.

Another example is below. We should probably turn off this linter, it’s too aggressive.

corneliusweig: I don't quite understand what you mean. Travis and my machine think that the linter passes.

Do you mean that the constant should _not_ be there and be inlined instead? This is what the linter complained about so I extracted the constants. If you think this is bad, I can see what linter needs to be disabled.

ahmetb: ah, wait, I now see that the linter is actually trying to help us.

that said, it's erroneous. the "unknown" string on both gitCommit() and gitTag() methods mean different things. so they should not be shared by the same variable. we can do:


    const unknownGitCommit = "unknown"
    const unknownGitTag = "unknown"

similarly, the tests below actually test against `v != unknown` which isn't good.

When we update the value of the const, we might be breaking the API (in this case we aren't), but the if test continues to test `v != "unknown"` literal string comparison, it could catch that.

So I think we should still disable this particular linter.

corneliusweig: good catch. I disabled the `goconst` linter. I think it would have allowed two occurrences of "unknown" in tests, but `goconst` is not critical at all.

corneliusweig: @ahmetb can you take another look or close this PR?

pkg/version/version_test.go
```diff
@@ -16,6 +16,8 @@ package version
 
 import "testing"
 
+const commitSha = "abcdef"
```
ahmetb: Can you revert changes to  this file? We since removed this linter.

corneliusweig: Sure. PTAL

PR #181, opened 2019-05-09
Switch imports to sigs.k8s.io/krew

soltysh:
/assign @ahmetb


corneliusweig:
@soltysh There are quite a few references to the `github.com/kubernetes-sigs/krew` repo left.
I think this is also the reason why travis fails, because it still says
```yaml
go_import_path: github.com/kubernetes-sigs/krew
```

PR #186, opened 2019-05-21
Add CLI for setting Windows path in readme

rbirkby:
Use ```setx``` DOS commandline to set per-user %PATH% environment variable  - [Windows docs](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/setx)

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,12 @@ Read the [User Guide](./docs/USER_GUIDE.md) for detailed documentation.
        .\krew-windows_amd64.exe install --manifest=krew.yaml --archive=krew.zip
 
 1. Add `%USERPROFILE%\.krew\bin` to your `PATH` environment variable
-   ([how?](https://java.com/en/download/help/path.xml))
+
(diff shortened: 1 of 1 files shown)


ahmetb:
/hold
/assign

Problem is, that only takes effect in that command-line window, right?

We want users to launch system settings control panel and update the environment variable there.

rbirkby:
No - using ```set``` only takes effect in the current DOS session (ie current shell) whereas ```setx``` affects the whole environment for new shells only - similar to adding to .bashrc.

I added a hold to this PR because setx for user-local PATH env can duplicate all entries because it merges both user-local and system PATH. That is unacceptable. Using /M switch can help but presents its own problems.

ahmetb:
I think users should use the relevant Control Panel to do this, right?

Maybe we can link to a page showing how to do it in the UI?

ahmetb:
I'll close this for now. Please let us know if you find a better solution. :)
/close

PR #187, opened 2019-05-24
add: description how to install krew on fish

### LINKED ISSUE #179: Install example for fish shell
Currently there is only an example for bash and zsh. There should be an example for fish shell as well.

nammn:
Adds a description section to install krew with fish.
This should close #179

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,9 @@ Read the [User Guide](./docs/USER_GUIDE.md) for detailed documentation.
 > :warning: **Warning:** krew is only compatible with kubectl v1.12 or higher.
 
 **macOS and Linux:**
+
 #### Bash and ZSH
+
 1. Make sure that `git` is installed.
 2. Run this command in your terminal to download and install `krew`:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,6 +59,31 @@ Read the [User Guide](./docs/USER_GUIDE.md) for detailed documentation.
 
    and restart your shell.
 
+#### Fish
+1. Make sure that `git` is installed.
```
ahmetb: Please add an empty line before/after lists.

README.md
```diff
@@ -59,6 +59,31 @@ Read the [User Guide](./docs/USER_GUIDE.md) for detailed documentation.
...
+2. Run this command in your terminal to download and install `krew`:
+
+    ```fish
+    begin
+      set -x; cd (mktemp -d) &&
```
ahmetb: What if mktemp returns a path with space in it?
Shouldnt you quote this to prevent splitting?

nammn: Hmm, good question. Actually, I am not that deep into fish, besides using it.
Taking a look.

README.md
nammn: 👍

README.md
nammn: Just saw this while reading through fish.
https://github.com/fish-shell/fish-shell/issues/159#issuecomment-9533091.
It seems that subcommands in fish prevent space splitting.

Issue #196, opened 2019-06-03
Krew info for multiple plugins is hard to read


corneliusweig:
When calling `krew info` for multiple plugins, the info for all plugins is concatenated without any separators. This makes the output hard to read, for example
```
krew info konfig konfig
```

I see two options:
- forbid passing multiple plugin names and expect exactly one plugin name
- separate the info yamls with "\n---\n" which is more or less k8s standard.

Besides, repeating a plugin prints the info twice. Should that be allowed? This is an absolute edge-case and maybe not worth any extra logic.

ahmetb:
I wasn’t aware it accepted multiple arguments. Feel free to remove the support for multiple args.

PR #197, opened 2019-06-03
cmd: info: Support only one argument

nirmoy:
Fixes: #196


ahmetb:
If you can `git commit --amend --edit` and remove the `Fixes: #...` statement and do a `git push -f`, we should be good here.

nirmoy:
@ahmetb @corneliusweig Thanks for reviewing. Updated

Issue #198, opened 2019-06-03
Integration test for cli behavior


ahmetb:
We've been successfully steering clear from testing the behavior of the CLI, but I think the time is approaching.

Ahead of the major refactoring for allowing multiple indexes (#23) I would love to get some integration test suite going that tests the end-to-end experience of the CLI from a user perspective.

Historically I've used https://github.com/sstephenson/bats for this purpose of running test cases that are mainly command executions in an ordered way, and processing their output/exit code with functionality like teardown/cleanup. 

`bats` is no longer maintained and I think the industry has moved on to doing this with directly writing programs/test suites in Go that exec a bunch of commands with neat helpers.

Having something like this that tests the main user workflow can ensure that we can further safely go into large-scale refactoring, so it would be a great long-term investment.

/kind proposal

corneliusweig:
I also think the go test suite is the best way to go forward. We can write a fluent wrapper, so that this will also read very nicely.

One concern I have is that the index is an external dependency which is not well-controlled. That's always bad for tests. What do you think about making the (single) index configurable? Maybe behind a hidden CLI flag. That way, we can also use file-based git repos such as `file:///tmp/krew/index`. WDYT?

corneliusweig:
On second thoughts.. I think this will work without a configurable repo URL, as long as the test setup initializes the index in `KREW_ROOT` appropriately.

ahmetb:
While I agree that we should allow customization of index, and we will in (#23), testing with a remote index and a real plugin would actually help test the real code path executed by users.

corneliusweig:
Fixed by #203 #208 #242 
/close

PR #202, opened 2019-06-04
Ensure index updates do not fail if the local and remote history diverge

### LINKED ISSUE #199: Determine if index clone can be fast-forwarded, if not re-clone
/kind proposal
/priority important-soon

Once we start offering multiple indexes (#23), inevitably the plugin authors will do weird stuff like rewriting the git history (probably accidentally) and we don't want this to break `krew update`.

**Decision 1:** Should we try to handle this case?

So I think, ahead of time, we need to implement something that does:
1. git fetch-ish something to retrieve the commits (requires connection)
2. determine if the HEAD can be fast-forwarded (offline operation)
3. if can be fast-forwarded, go ahead (offline)
4. if cannot be fast-forwarded (at this point it's safe(?) to assume it's because someone rewrote the history), nuke the local clone and re-clone the repo.

**Decision 2 (if decision 1 is accepted):** Are the steps above a sound approach to this problem?

corneliusweig:
Unfortunately, this is not testable at the moment, but this should be covered in #198.

Close #199

### DIFF
--- a/pkg/gitutil/git.go
+++ b/pkg/gitutil/git.go
@@ -47,7 +47,12 @@ func IsGitCloned(gitPath string) (bool, error) {
 
 // update will fetch origin and set HEAD to origin/HEAD.
 func update(destinationPath string) error {
-	return exec(destinationPath, "pull", "--ff-only", "-v")
+	if err := exec(destinationPath, "fetch", "-v"); err != nil {
+		return errors.Wrapf(err, "fetch index at %q failed", destinationPath)
+	}
+
+	err := exec(destinationPath, "reset", "--hard", "@{upstream}")
+	return errors.Wrapf(err, "reset index at %q failed", destinationPath)
 }
 
 // EnsureUpdated will ensure the destination path exists and is up to date.


pkg/gitutil/git.go
```diff
@@ -47,7 +47,15 @@ func IsGitCloned(gitPath string) (bool, error) {
 
 // update will fetch origin and set HEAD to origin/HEAD.
 func update(destinationPath string) error {
-	return exec(destinationPath, "pull", "--ff-only", "-v")
+	if err := exec(destinationPath, "fetch", "-v"); err != nil {
+		return errors.Wrapf(err, "fetch index at %q failed", destinationPath)
+	}
+
+	if err := exec(destinationPath, "reset", "--hard", "@{upstream}"); err != nil {
+		return errors.Wrapf(err, "reset index at %q failed", destinationPath)
```
ahmetb: you actually don't need to do err!=nil check for the last err, you can just return it (as errors.Wrapf will return nil when err==nil.

corneliusweig: cool, I didn't know that :sunglasses:

PR #206, opened 2019-06-05
hack/run-in-docker: refactoring

ahmetb:
- build krew binary outside container (do not duplicate build process)
  - allows binary to be refreshed without rebuilding container
    this way, we preserve container session/state
  - also good for not leaking unused images to docker engine
    that needs to be cleaned up regularly
- mount KUBECONFIG inside the container (not sure if it's actually useful)
  I probably won't use it (since GKE clusters require gcloud to work, too)
- made the script a bit friendlier to follow as a user.

### DIFF
--- a/hack/run-in-docker.sh
+++ b/hack/run-in-docker.sh
@@ -14,10 +14,42 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# This script starts a development container by mounting the krew binary from
+# the local filesystem.
+
 set -euo pipefail
-
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+log() { echo >&2 "$*"; }
+log_ok() { log "$(tput setaf 2)$*$(tput sgr0)"; }
+log_fail() { log "$(tput setaf 1)$*$(tput sgr0)"; }
+image="krew:sandbox"
 
-docker build -f "${SCRIPTDIR}/sandboxed.Dockerfile" -t krew:sandbox "${SCRIPTDIR}/.."
+krew_bin="${SCRIPTDIR}/../out/bin/krew-linux_amd64"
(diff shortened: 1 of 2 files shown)


corneliusweig:
Could you change the runtime image to something other than alpine? That was a bad idea, because whenever something needs a `libc`, there's only `libc.musl`.
What about `gcr.io/gcp-runtimes/ubuntu_16_0_4`?

ahmetb:
> What about `gcr.io/gcp-runtimes/ubuntu_16_0_4`?

We can do that. Is there a reason why we should still have `go` in this image?

corneliusweig:
> Is there a reason why we should still have `go` in this image?

The builder image needs it, but the final image with the krew binary inside has no `go` toolchain. Or do I misunderstand your question?

ahmetb:
I'm afraid you didn't read the PR description. (:

> * build krew binary outside container (do not duplicate build process)
>   
>   * allows binary to be refreshed without rebuilding container
>     this way, we preserve container session/state

we don't need Go inside the container.

corneliusweig:
> I'm afraid you didn't read the PR description.

Yes, that's right. I was only sharing the improved version of the former Dockerfile. With your improvements that will no longer be necessary.

ahmetb:
@corneliusweig are you actually using this script?
I'm planning on fixing the PR, but not sure about the use case.

corneliusweig:
@ahmetb Yes, whenever I want to review a new submission to krew-index, I try to install and run it inside docker. That's my main use-case.

ahmetb:
/assign @corneliusweig 
sorry just got to this!

corneliusweig:
Beautiful! Works like a charm. :unicorn:

/lgtm
/approve

corneliusweig:
/lgtm cancel

Can you change the PR title?

corneliusweig:
/lgtm
Prow won't merge anyway until the title is fixed.

PR #208, opened 2019-06-06
More integration tests

corneliusweig:
These are based on #203 and should be merged after that one. Reviewing before #203 is merged makes also no sense :-/

The tests currently cover the happy path of all subcommands excluding `krew upgrade`. Upgrade is a special case because it needs two versions of a valid plugin. That will be difficult to set up and may be better in a separate PR.

### DIFF
--- a/cmd/krew/cmd/list.go
+++ b/cmd/krew/cmd/list.go
@@ -51,7 +51,7 @@ Remarks:
 					names = append(names, name)
 				}
 				sort.Strings(names)
-				fmt.Fprint(os.Stdout, strings.Join(names, "\n"))
+				fmt.Fprintln(os.Stdout, strings.Join(names, "\n"))
 				return nil
 			}
 
--- a/cmd/krew/cmd/list.go
+++ b/cmd/krew/cmd/list.go
@@ -51,7 +51,7 @@ Remarks:
 					names = append(names, name)
 				}
 				sort.Strings(names)
-				fmt.Fprintln(os.Stdout, strings.Join(names, "\n"))
+				fmt.Fprint(os.Stdout, strings.Join(names, "\n"))
 				return nil
 			}
 
--- a/docs/CONTRIBUTOR_GUIDE.md
+++ b/docs/CONTRIBUTOR_GUIDE.md
@@ -59,16 +59,6 @@ hack/make-binaries.sh
 hack/run-integration-tests.sh
 ```
 
-The above builds binaries for all supported platforms.
-Building only for your platform produces a slightly different image but will
-work in most circumstances:
-
-```bash
-go build ./cmd/krew
-# you need to specify your local krew binary when using this method:
-hack/run-integration-tests.sh ./krew
-```
-
 ## Testing `krew` in a sandbox
 
 After making changes to krew, you should also check that it behaves as expected.
--- a/hack/run-integration-tests.sh
+++ b/hack/run-integration-tests.sh
@@ -44,8 +44,13 @@ install_kubectl_if_needed() {
   fi
 }
 
-install_kubectl_if_needed
 KREW_BINARY=$(readlink -f "${1:-$KREW_BINARY_DEFAULT}")  # needed for `kubectl krew` in tests
+if [[ ! -x "${KREW_BINARY}" ]]; then
+  echo "Did not find $KREW_BINARY. You need to build krew before running the integration tests"
+  exit 1
+fi
 export KREW_BINARY
+
+install_kubectl_if_needed
 
 go test -v ./...
--- a/hack/run-integration-tests.sh
+++ b/hack/run-integration-tests.sh
@@ -18,7 +18,9 @@ set -euo pipefail
 
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 BINDIR="${SCRIPTDIR}/../out/bin"
-KREW_BINARY_DEFAULT="$BINDIR/krew-linux_amd64"
+GOOS="$(go env GOOS)"
+GOARCH="$(go env GOARCH)"
+KREW_BINARY_DEFAULT="$BINDIR/krew-${GOOS}_${GOARCH}"
 
 if [[ "$#" -gt 0 && ( "$1" = '-h' || "$1" = '--help' ) ]]; then
   cat <<EOF
--- a/test/krew_test.go
+++ b/test/krew_test.go
@@ -162,6 +162,7 @@ func TestKrewUpdate(t *testing.T) {
 	test, cleanup := krew.NewTest(t)
 	defer cleanup()
 
+	// nb do not call WithIndex() here
 	test.Krew("update").RunOrFail()
 	plugins := lines(test.Krew("search").RunOrFailOutput())
 	if len(plugins) < 10 {
--- a/test/krew_test.go
+++ b/test/krew_test.go
@@ -16,7 +16,9 @@
 package test
 
 import (
+	"bytes"
 	"fmt"
+	"regexp"
 	"strings"
 	"testing"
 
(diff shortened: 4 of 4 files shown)


docs/CONTRIBUTOR_GUIDE.md
```diff
@@ -47,9 +47,28 @@ To run tests locally, the easiest way to get started is with
...
+hack/make-binaries.sh
+hack/run-integration-tests.sh
+```
+
+The above builds binaries for all supported platforms.
+Building only for your platform produces a slightly different image but will
+work in most circumstances:
+
+```bash
+go build ./cmd/krew
```
ahmetb: ehh, I'm a bit wary about `go build` as we might be doing interesting things like static linking that it doesn't  capture.

I would've preferred more if `run-integraton-tests` detected out/bin/krew-$OS_$ARCH, but we can keep it as is for now.

hack/run-integration-tests.sh
```diff
@@ -0,0 +1,49 @@
...
+if [[ "$#" -gt 0 && ( "$1" = '-h' || "$1" = '--help' ) ]]; then
+  cat <<EOF
+Usage:
+  $0 krew  # uses the given krew binary for running integration tests
+  $0       # assumes a pre-built krew binary at $KREW_BINARY_DEFAULT
+EOF
+  exit 0
+fi
+
+install_kubectl_if_needed() {
```
ahmetb: Big -1 for this. It's not clear from the name that this script installs kubectl –so let's not do that.

You don't want to do things like this on dev machines either. For travis CI we should just export this as a step there.

docs/CONTRIBUTOR_GUIDE.md
corneliusweig: Yeah, I wanted to outline an alternative to the current build script but that is probably not necessary.

Using the correct platform binary is absolutely necessary, I'lll add that.

hack/run-integration-tests.sh
corneliusweig: I think that concern is a bit exaggerated. After all, the binary gets placed in the `out/` directory which is to be considered ephemeral. However, I can make the downloading more visible and only enable it when passing a flag like `--download-kubectl-if-required` or so.

I would like to avoid extra preparatory steps in travis, that's why I put everything in that script. That also makes passing in the env variables more convenient which otherwise would clutter the travis steps.

ahmetb: Regardless I think it should be decoupled. 

e.g. travis can do: env KUBECTL=out/bin/kubectl hack/get-kubectl.sh
and in the tests we offer a $KUBECTL customization. 
This way the integration tests have no idea about what to do when kubectl is missing. The hidden/auto magic behavior tends to trip people occasionally. So let’s stick to what the script says it does in the filename.

corneliusweig: Sorry for my slow response on this topic. If you have such a strong opinion here, then let's do it your way.

As there is a separate issue for this, I'll leave that for another PR though.

docs/CONTRIBUTOR_GUIDE.md
corneliusweig: I added some logic to determine the correct krew binary for integration tests.

hack/run-integration-tests.sh
ahmetb: I opened #210 for  this.

cmd/krew/cmd/list.go
```diff
@@ -51,7 +51,7 @@ Remarks:
 					names = append(names, name)
 				}
 				sort.Strings(names)
-				fmt.Fprintln(os.Stdout, strings.Join(names, "\n"))
+				fmt.Fprint(os.Stdout, strings.Join(names, "\n"))
```
ahmetb: This command's output should end with a `\n`.  so I suspect fmt.Println was correct here.

ahmetb: see my comment  for  lines()

test/krew_test.go
```diff
@@ -51,3 +177,10 @@ func skipShort(t *testing.T) {
 		t.Skip("skipping integration test")
 	}
 }
+
+func lines(in []byte) []string {
+	if string(in) == "" {
+		return nil
+	}
+	return strings.Split(string(in), "\n")
```
ahmetb: ideally you should do `strings.TrimRight(string(in))` so that you avoid taking trailing `\n` which almost  all commands will have.

cmd/krew/cmd/list.go
corneliusweig: If that is a requirement, you are right. `Fprintln` only has the disadvantage that an empty list gets printed as "\n\n" instead of "\n" or "".

I'll think of a better solution then.

ahmetb: Hmm I don't see \n\n.

```
export KREW_ROOT=$(mktemp -d)
kubectl krew update
kubectl krew list
```

this command prints:

```
PLUGIN VERSION\n
```

when you redirect it  to a non-TTY stdout, it prints just `\n` which I think is ok?

test/krew_test.go
ahmetb: with this,  you don't   need the =="" check,

cmd/krew/cmd/list.go
corneliusweig: Hmm.. true. I don't remember how I encountered the double "\n\n" :/

test/krew_test.go
corneliusweig: Unfortunately, the `==""` is still required. For `krew list` I need the list of results to grow by one if a plugin is installed. However, when starting from none (resulting in "\n") to one plugin ("whatever\n") would result in 1 entry each.

ahmetb: hmm. I'm not sure if that's accurate.

Since in this case the stdout is not a tty, we don't print the tabular output.

I expect no output or at most `\n`. Both of which would return `[]string{}` (len=0).

When you install a plugin I expect this to return `pluginname\n`, if you split it by `\n` it should return  `[]string{"pluginnname"}` (len=1). 

Alternatively you can also test that case as `output.Contains("pluginname")`, which is also just fine.

test/krew_test.go
```diff
@@ -36,13 +58,117 @@ func TestKrewInstall(t *testing.T) {
...
+
+	output := test.Krew("version").RunOrFailOutput()
+
+	requiredSubstrings := []string{
+		"IsPlugin",
+		fmt.Sprintf("BasePath        %s", test.Root()),
+		"ExecutedVersion",
+		"GitTag",
+		"GitCommit",
+		"IndexURI        https://github.com/kubernetes-sigs/krew-index.git",
```
ahmetb: I feel like we'll easily break these spaces. :) 
I'm not sure if this whole thing is necessary. I currently only care about exitcode=0. :)

test/krew_test.go
```diff
@@ -36,13 +58,117 @@ func TestKrewInstall(t *testing.T) {
...
+	}
+}
+
+func TestKrewInfo(t *testing.T) {
+	skipShort(t)
+
+	test, cleanup := krew.NewTest(t)
+	defer cleanup()
+
+	output := test.WithIndex().Krew("info", validPlugin).RunOrFailOutput()
```
ahmetb: again, here we probably just need exitcode=0 (+ I realize we don't have tests for true negatives such as "info for nonexisting plugin fails"; we can do those later) as that's the behavior that this cmd offers (the NAME: foo) isn't part of a supported API.

test/krew_test.go
corneliusweig: > I expect no output or at most `\n`. Both of which would return `[]string{}` (len=0).

That is unfortunately not the case. If the input string is empty or "\n", it will result in `[]string{""}`.

test/krew_test.go
corneliusweig: I think it's worth a lot to know if something is present in the output. Do you like a regex-based assertion better?

test/krew_test.go
corneliusweig: Yeah right. For good measure, I already added the negative test case.

test/krew_test.go
ahmetb: I think anything is fine.

test/krew_test.go
```diff
@@ -36,13 +60,142 @@ func TestKrewInstall(t *testing.T) {
...
+
+func TestKrewInfo(t *testing.T) {
+	skipShort(t)
+
+	tests := []struct {
+		name      string
+		plugin    string
+		shouldErr bool
+	}{
+		{
```
ahmetb: please don't put negative cases to the same test case.
they are different test cases. and these aren't unit tests. tabular tests like []struct{}{} are good for tabular unit tests. but ideally each integration test case should be a separate method testing a particular scenario workflow.

e.g. try

TestKrewInfo
TestKrewInfoInvalidPlugin

ahmetb:
I recommend removing the WIP label so we can move forward with merging.

test/krew_test.go
corneliusweig: Makes sense. The test cases are not split into separate test functions.

corneliusweig:
> I recommend removing the WIP label so we can move forward with merging.

The WIP label should already have been gone for a while. At least I don't see it any longer.

PR #211, opened 2019-06-11
Upgrade golangci-lint to version v1.17.1

corneliusweig:
Also fix new reported issues


ahmetb:
Travis failure.

corneliusweig:
Interesting. The latest `golangci-lint` version does no longer run with go-1.10. Unfortunately, `gox` is still stuck on this version :(

I'll close this because it is blocked.

PR #215, opened 2019-06-19
Fix readme for wrong use of fish subshell in string

kirrmann:
The scripts fails and outputs an error `cd: The directory '(mktemp -d)' does not exist`. This error appears because `()` in strings aren't generally recognized as subshells in fish. To fix this I set a variable which we use to `cd` to. There seems to be no way to quote subshell commands in fish correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,13 +68,13 @@ Read the [User Guide](./docs/USER_GUIDE.md) for detailed documentation.
 
     ```fish
     begin
-      set -x; cd "(mktemp -d)" &&
+      set -x; set temp_dir (mktemp -d); cd "$temp_dir" &&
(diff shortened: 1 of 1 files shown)


kirrmann:
registered to CLA

kirrmann:
recheck CLA

kirrmann:
recheck CLA

README.md
```diff
@@ -68,7 +68,7 @@ Read the [User Guide](./docs/USER_GUIDE.md) for detailed documentation.
 
     ```fish
     begin
-      set -x; cd "(mktemp -d)" &&
+      set -x; set temp_dir (mktemp -d); cd "$temp_dir" &&
```
corneliusweig: Thanks for providing this fix!

Can you please unset the variable with `set -e tempdir` when it is no longer needed?

kirrmann: yes of course 👍

Issue #213, opened 2019-06-12
Remove obsolete TODO items


corneliusweig:
The code contains quite a few open TODO items. I think it would be worth going through that list and remove stale or completed items. Where appropriate, other items should IMHO be moved into github issues for better visibility.

That should ideally be done by @lbb or @ahmetb who know more about those TODO items.

----

Examples with probably pointless TODO items: `pkg/index/types.go`, `pkg/index/validate.go`

corneliusweig:
@ahmetb There are 6 todo items remaining. Would you agree if I opened issues for those instead of todo comments?

@lbb can you answer the question in `pkg/installation/upgrade.go:109`:

> TODO(ahmetb,lbb) is this part implemented???

ahmetb:
I don't think so, it seems to be a dead code. :|
Hopefully actually using the receipts will help us track down these pieces more.

I wouldn't focus on closing the TODOs, I think most of them are left in the code as a potential future improvement that isn't critical atm.

PR #218, opened 2019-06-19
Cleanup some TODOs

ahmetb:
- cleaned up some obsolete todos (ref: #213), some still relevant
- removed PluginIndex type, as it wasn't used for any reason

### DIFF
--- a/pkg/index/indexscanner/scanner.go
+++ b/pkg/index/indexscanner/scanner.go
@@ -31,25 +31,24 @@ import (
 )
 
 // LoadPluginListFromFS will parse and retrieve all plugin files.
-func LoadPluginListFromFS(indexDir string) (index.PluginList, error) {
-	var indexList index.PluginList
+func LoadPluginListFromFS(indexDir string) ([]index.Plugin, error) {
 	indexDir, err := filepath.EvalSymlinks(indexDir)
 	if err != nil {
-		return indexList, err
+		return nil, err
 	}
 
 	files, err := ioutil.ReadDir(filepath.Join(indexDir, "plugins"))
 	if err != nil {
-		return indexList, errors.Wrap(err, "failed to open index dir")
+		return nil, errors.Wrap(err, "failed to open index dir")
 	}
 
+	list := make([]index.Plugin, 0)
 	for _, f := range files {
 		if f.IsDir() {
 			continue
 		}
 
 		pluginName := strings.TrimSuffix(f.Name(), filepath.Ext(f.Name()))
-		// TODO(lbb): Use go routines to speed up slow FS operations.
 		p, err := LoadPluginFileFromFS(indexDir, pluginName)
 		if err != nil {
 			// Index loading shouldn't fail because of one plugin.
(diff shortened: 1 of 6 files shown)


pkg/index/indexscanner/scanner.go
```diff
@@ -31,37 +31,35 @@ import (
...
+		return nil, err
 	}
 
 	files, err := ioutil.ReadDir(filepath.Join(indexDir, "plugins"))
 	if err != nil {
-		return indexList, errors.Wrap(err, "failed to open index dir")
+		return nil, errors.Wrap(err, "failed to open index dir")
 	}
 
+	list := make([]index.Plugin, 0)
```
corneliusweig: I find it good practice to pre-allocate the required array like that:
```suggestion
	list := make([]index.Plugin, 0, len(files))
```
Granted, it's a bit too large but a good heuristic.

pkg/index/indexscanner/scanner.go
```diff
@@ -111,7 +109,10 @@ func DecodePluginFile(r io.Reader) (index.Plugin, error) {
 	}
 	decoder := json.NewDecoder(bytes.NewReader(jsonRaw))
 
-	// TODO(lbb): Enable strict visioning.
+	// TODO(ahmetb): when we have a stable API that don't add new fields,
```
corneliusweig:
```suggestion
	// TODO(ahmetb): when we have a stable API that won't add new fields,
```

pkg/index/types.go
```diff
@@ -18,16 +18,15 @@ import (
...
+// Plugin describes a plugin manifest file.
 type Plugin struct {
 	metav1.TypeMeta   `json:",inline" yaml:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty" yaml:"metadata"`
 
 	Spec PluginSpec `json:"spec"`
 }
 
-// PluginSpec TODO(lbb)
+// PluginSpec describes a plugin specification.
```
corneliusweig:
```suggestion
// PluginSpec is the plugin specification.
```

pkg/index/types.go
```diff
@@ -38,7 +37,8 @@ type PluginSpec struct {
 	Platforms []Platform `json:"platforms,omitempty"`
 }
 
-// Platform TODO(lbb)
+// Platform describes the how to match to a particular platform (os, arch) and
```
corneliusweig: I find that comment easier to read, if it is reversed, like so:
```
// Platform describes how to perform an installation on a specific platform
// and how to match the target platform (os, arch).
```

pkg/index/types.go
```diff
@@ -52,16 +52,9 @@ type Platform struct {
 	Bin string `json:"bin"`
 }
 
-// FileOperation TODO(lbb)
+// FileOperation explains a file copying operation from plugin archive to the
```
corneliusweig:
```suggestion
// FileOperation specifies a file copying operation from plugin archive to the
```

pkg/index/indexscanner/scanner.go
ahmetb: I honestly don't like the static analysis tool that made me do this.
I want to just do `var list []type` what's wrong with that if I don't know the exact size? Any ideas why it's complaining?

corneliusweig: Not 100% sure here. But I think it's because the linter assumes the `list` ought to be approximately the size of `files`. And in that case it makes perfect sense to pre-allocate.

ahmetb:
Merging manually since approved manually.

PR #220, opened 2019-06-19
hack: move out install_kubectl_if_needed

### LINKED ISSUE #210: hack/run-integration-tests.sh: move install_kubectl_if_needed out
This script shouldn't be downloading `kubectl`. (It may however offer a $KUBECTL to customize the used kubectl path).

Right now `install_kubectl_if_needed`  function in there helps us install kubectl on Travis CI, but ideally this should be a separate script, and since the name of `run-integration-tests.sh` doesn't make it clear that there's kubectl installation in there, it shouldn't install kubectl.

ahmetb:
- install appropriate kubectl (goos/goarch)
- save it to $GOPATH/bin since "export $PATH" won't work outside a particular
  script
- lowercase naming for local variables

Fixes #210.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ script:
 - hack/run-lint.sh
 - go test -v -short -coverprofile=coverage.txt -covermode=atomic ./...
 - hack/make-all.sh
+- hack/ensure-kubectl-installed.sh
 - hack/run-integration-tests.sh
 after_success:
 - bash <(curl -s https://codecov.io/bash)
--- a/hack/run-integration-tests.sh
+++ b/hack/run-integration-tests.sh
@@ -31,26 +31,11 @@ EOF
   exit 0
 fi
 
-install_kubectl_if_needed() {
-  if hash kubectl 2>/dev/null; then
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -9,6 +9,7 @@ notifications:
 before_install:
 - go get github.com/mitchellh/gox
 script:
+- hack/ensure-kubectl-installed.sh
```
corneliusweig: Can you move this down right before `run-integration-tests.sh`? That would speed up failing runs a little.

hack/run-integration-tests.sh
```diff
@@ -31,19 +31,6 @@ EOF
   exit 0
 fi
 
-install_kubectl_if_needed() {
```
corneliusweig: This function is still referenced in https://github.com/kubernetes-sigs/krew/blob/e643ab39cfacdf6f30943bcc4f27d65cbf142553/hack/run-integration-tests.sh#L41
That's also why travis failed.

ahmetb:
I'll go ahead and merge this since @corneliusweig approved.

PR #221, opened 2019-06-24
hack: update verify-code-patterns.sh

### LINKED ISSUE #217: hack/verify-code-patterns.sh: fails locally
When I run `hack/verify-code-patterns.sh` from the root (at d6629aa) it fails with exitcode=1 without any output. This looks bad in `hack/run-tests.sh`.

Throwing a `set -x` shows some stuff:

```
hack/verify-code-patterns.sh
++ find . -name '*_test.go' -not -path './vendor/*' -print0
++ xargs -0 grep -En 'ioutil\.TempDir'
++ [[ 1 == 123 ]]
+ out=
```

So it looks like for some reason on my shell, exitcode is 1, and not 123.

Although, when I run it on zsh directly, it works.

```
$ find . -name '*_test.go' -not -path './vendor/*' -print0 | xargs -0 grep -En 'ioutil\.TempDir' || [[ $? == 123 ]]; echo $?
0
```

I think we need a cleaner solution that doesn't do `[[ $? == 123 ]]`.

ahmetb:
This fixes an issue where sometimes bash returns 1 instead of 123
in the exitcode of the pipe. This simplifies the find|xargs|grep with
just grep.

Fixes #217
/assign @corneliusweig

### DIFF
--- a/hack/verify-code-patterns.sh
+++ b/hack/verify-code-patterns.sh
@@ -16,7 +16,8 @@
 
 set -euo pipefail
 
-out="$( find . -name '*_test.go' -not -path './vendor/*' -print0 | xargs -0 grep -En 'ioutil\.TempDir' || [[ $? == 123 ]] )"
+# Disallow usage of ioutil.TempDir in tests in favor of testutil.
+out="$(grep --include '*_test.go' --exclude-dir 'vendor/' -EIrn 'ioutil.\TempDir' || true)"
 if [[ -n "$out" ]]; then
   echo >&2 "You used ioutil.TempDir in tests, use 'testutil.NewTempDir()' instead:"
   echo >&2 "$out"


hack/verify-code-patterns.sh
```diff
@@ -16,7 +16,8 @@
 
 set -euo pipefail
 
-out="$( find . -name '*_test.go' -not -path './vendor/*' -print0 | xargs -0 grep -En 'ioutil\.TempDir' || [[ $? == 123 ]] )"
+# Disallow usage of ioutil.TempDir in tests in favor of testutil.
+out="$(grep --include '*_test.go' --exclude 'vendor/' -EIrn 'ioutil.\TempDir' || true)"
```
corneliusweig: This does not yet exclude the vendor directory as desired. This does for me:
```suggestion
out="$(grep --include '*_test.go' --exclude-dir 'vendor/' -EIrn 'ioutil.\TempDir' || true)"
```

ahmetb:
> One of the options needs to be adjusted (at least on my system).

Does it work if I say `--exclude 'vendor/*'`? I hope it's not BSD grep vs GNU grep difference?

corneliusweig:
No, `--exclude 'vendor/*'` or similar does not work. My man page says:

> --exclude=GLOB
              Skip any command-line file with a name suffix that matches the pattern GLOB, using
              wildcard  matching; a name suffix is either the whole name, or any suffix starting
              after a / and before a +non-/.  When searching recursively, skip any subfile whose
              base name matches GLOB; the base name is the part after the last /.  A pattern can
              use *, ?, and [...]  as  wildcards,  and  \  to  quote  a  wildcard  or  backslash
              character literally.

So it's clearly limited to the basename.

---
For completeness:

> --exclude-dir=GLOB
              Skip any command-line directory with a name suffix that matches the pattern  GLOB.
              When  searching  recursively,  skip any subdirectory whose base name matches GLOB.
              Ignore any redundant trailing slashes in GLOB.

ahmetb:
Yeah makes sense. I'll fix.

PR #223, opened 2019-06-25
Small refactorings for better code structure

corneliusweig:
This PR includes some small refactorings which originally were part of #195 . In particular:

* move the expected ending for manifests into a code constant
* check that this code constant is not used outside of `constants.go`
* move the logic to match a target platform out of `installation` and into the `index` package.

### DIFF
--- a/cmd/krew/cmd/search.go
+++ b/cmd/krew/cmd/search.go
@@ -79,7 +79,7 @@ Examples:
 			var status string
 			if _, ok := installed[name]; ok {
 				status = "installed"
-			} else if _, ok, err := installation.GetMatchingPlatform(plugin); err != nil {
+			} else if _, ok, err := plugin.Spec.GetMatchingPlatform(); err != nil {
 				return errors.Wrapf(err, "failed to get the matching platform for plugin %s", name)
 			} else if ok {
 				status = "available"
--- a/hack/verify-code-patterns.sh
+++ b/hack/verify-code-patterns.sh
@@ -23,7 +23,11 @@ if [[ -n "$out" ]]; then
   exit 1
 fi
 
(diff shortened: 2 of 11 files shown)


cmd/krew/cmd/search.go
```diff
@@ -19,10 +19,12 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+
 	"sigs.k8s.io/krew/pkg/index/indexscanner"
```
ahmetb: seems like this import should be on the next block.

hack/verify-code-patterns.sh
```diff
@@ -22,3 +22,10 @@ if [[ -n "$out" ]]; then
   echo >&2 "$out"
   exit 1
 fi
+
+out="$(grep --include '*.go' --exclude "*_test.go" --exclude 'constants.go' --exclude-dir 'vendor/' -EIrn '\.yaml"' || true)"
```
ahmetb: this is a long line now, let's split into multiple with `\`

also `--exclude 'vendor/*'` should be working here.

ahmetb:
Small comments.
/approve

hack/verify-code-patterns.sh
corneliusweig: I improved the legibility as suggested.

For my GNU sed, the `--exclude 'vendor/*'` command does not work. But I think the `--exclude-dir` version also works for BSD sed. At least https://www.freebsd.org/cgi/man.cgi?query=grep&sektion=&n=1 sounds alike.

ahmetb: You're totally right, something’s up with this --exclude.
I think we can keep this.

PR #228, opened 2019-06-27
Various fixes to integration tests

### LINKED ISSUE #224: hack/run-integration-tests: runs all tests
hack/run-integration-tests.sh hsa `go test -v ./...`, that's why it's running all  tests.

This must use `${SCRIPTDIR}/../test`. And ideally we should rename `./test` to `/.integration_tests`.

### LINKED ISSUE #225: hack/run-integration-tests: uses installed krew
hack/run-integration-tests.sh does
```

KREW_BINARY=$(readlink -f "${1:-$KREW_BINARY_DEFAULT}")  # needed for `kubectl krew` in tests
if [[ ! -x "${KREW_BINARY}" ]]; then
  echo "Did not find $KREW_BINARY. You need to build krew before running the integration tests"
  exit 1
fi
export KREW_BINARY
```

however it still calls `kubectl krew` under the covers, which for some reason isn't picking up `$KREW_BINARY`.

For example, when I have krew installed, an run this script, I see tests fail:

```
--- FAIL: TestKrewUninstall (1.34s)
    krew_test.go:70: krew [uninstall konfig]: exit status 255, F0625 16:52:31.788014   31000 root.go:52] unknown command "uninstall" for "krew"

        Did you mean this?
        	install
```

### LINKED ISSUE #227: hack/run-integration-tests: should not actually run plugin
Since we're installing 3rd party plugins in the test suite, there is a risk of
malicious code execution when the plugin runs in test suite. Also we're not
testing kubectl's behavior of executing a plugin when it is in PATH.

We can simply just do a kubectl-$PLUGIN lookup in PATH.

ahmetb:
I was hoping to write more integration tets, but along the way I noticed some
issues with how we run integration tests.
This PR addressses few things about the integration tests that bugged me.

- do not actually run the test plugin, it's out of our tool's scope
- run only integration tests from the script (previously ./...)
- DO NOT use the host-krew, require KREW_BINARY to be set (we almost never want to use host-krew)

Fixes #224
Fixes #225
Fixes #227

### DIFF
--- a/hack/run-integration-tests.sh
+++ b/hack/run-integration-tests.sh
@@ -38,5 +38,4 @@ if [[ ! -x "${KREW_BINARY}" ]]; then
 fi
 export KREW_BINARY
 
-"${SCRIPTDIR}/ensure-kubectl-installed.sh"
 go test -test.v sigs.k8s.io/krew/test
--- a/hack/run-integration-tests.sh
+++ b/hack/run-integration-tests.sh
@@ -18,9 +18,9 @@ set -euo pipefail
 
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 BINDIR="${SCRIPTDIR}/../out/bin"
-GOOS="$(go env GOOS)"
-GOARCH="$(go env GOARCH)"
-KREW_BINARY_DEFAULT="$BINDIR/krew-${GOOS}_${GOARCH}"
+goos="$(go env GOOS)"
+goarch="$(go env GOARCH)"
+KREW_BINARY_DEFAULT="${BINDIR}/krew-${goos}_${goarch}"
 
 if [[ "$#" -gt 0 && ( "$1" = '-h' || "$1" = '--help' ) ]]; then
   cat <<EOF
@@ -33,9 +33,10 @@ fi
 
 KREW_BINARY=$(readlink -f "${1:-$KREW_BINARY_DEFAULT}")  # needed for `kubectl krew` in tests
 if [[ ! -x "${KREW_BINARY}" ]]; then
-  echo "Did not find $KREW_BINARY. You need to build krew before running the integration tests"
+  echo "Did not find $KREW_BINARY. You need to build krew FOR ${goos}/${goarch} before running the integration tests."
   exit 1
 fi
 export KREW_BINARY
 
-go test -v ./...
+"${SCRIPTDIR}/ensure-kubectl-installed.sh"
+go test -test.v sigs.k8s.io/krew/test
--- a/test/krew/krew.go
+++ b/test/krew/krew.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"testing"
 	"time"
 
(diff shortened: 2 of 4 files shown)


hack/run-integration-tests.sh
```diff
@@ -33,9 +33,10 @@ fi
...
 KREW_BINARY=$(readlink -f "${1:-$KREW_BINARY_DEFAULT}")  # needed for `kubectl krew` in tests
 if [[ ! -x "${KREW_BINARY}" ]]; then
-  echo "Did not find $KREW_BINARY. You need to build krew before running the integration tests"
+  echo "Did not find $KREW_BINARY. You need to build krew FOR ${goos}/${goarch} before running the integration tests."
   exit 1
 fi
 export KREW_BINARY
 
-go test -v ./...
+"${SCRIPTDIR}/ensure-kubectl-installed.sh"
```
corneliusweig: That one puzzles me. You previously made a strong point against installing `kubectl` in this script. Quoting https://github.com/kubernetes-sigs/krew/issues/210#issue-454342289:

> (...)  since the name of run-integration-tests.sh doesn't make it clear that there's kubectl installation in there, it shouldn't install kubectl.

Calling that script here somehow defeats the purpose of the script, don't you think?

test/krew/krew.go
```diff
@@ -155,8 +187,7 @@ func (it *ITest) cmd(ctx context.Context) *exec.Cmd {
 	args = append(args, it.args...)
 
 	cmd := exec.CommandContext(ctx, "kubectl", args...)
-	cmd.Env = append(os.Environ(), it.env...)
-
+	cmd.Env = it.env // clear env, do not inherit from system
```
corneliusweig: Yeah, that makes much more sense!

test/krew/krew.go
```diff
@@ -32,51 +33,82 @@ const krewBinaryEnv = "KREW_BINARY"
...
-	return fmt.Sprintf("PATH=%s:%s", krewBinPath, path)
+	return v + string(os.PathListSeparator) + curPath
 }
 
-// Call configures the runner to call plugin with arguments args.
-func (it *ITest) Call(plugin string, args ...string) *ITest {
-	it.plugin = plugin
-	it.args = args
-	return it
+func (it *ITest) lookExecutable(file string) error {
```
corneliusweig: I know, you named that function like `exec.LookPath`. But it doesn't read well. How about

```suggestion
func (it *ITest) lookupExecutable(file string) error {
```

test/krew/krew.go
```diff
@@ -32,51 +33,82 @@ const krewBinaryEnv = "KREW_BINARY"
...
+	orig := os.Getenv("PATH")
+	defer func() { os.Setenv("PATH", orig) }()
+
+	binPath := filepath.Join(it.Root(), "bin")
+	os.Setenv("PATH", binPath+string(os.PathListSeparator)+binPath)
```
corneliusweig:
```suggestion
	os.Setenv("PATH", binPath)
```

test/krew/krew.go
```diff
@@ -32,51 +33,82 @@ const krewBinaryEnv = "KREW_BINARY"
...
-	if err := os.MkdirAll(krewBinPath, os.ModePerm); err != nil {
+// setupKrewBin symlinks the $KREW_BINARY to $tempDir/bin and returns the path
+// to this directory.
+func setupKrewBin(t *testing.T, tempDir *testutil.TempDir) string {
+	krewBinary, found := os.LookupEnv(krewBinaryEnv)
+	if !found {
+		t.Fatalf("%s environment variable pointing to krew binary not set", krewBinaryEnv)
+	}
+	binPath := tempDir.Path("bin")
+	if err := os.MkdirAll(binPath, 0755); err != nil {
```
corneliusweig: This `mkdir` functionality could also be moved to `TempDir`. Will avoid further repetition in the future.

test/krew/krew.go
```diff
@@ -32,51 +33,82 @@ const krewBinaryEnv = "KREW_BINARY"
...
+	binPath := filepath.Join(it.Root(), "bin")
+	os.Setenv("PATH", binPath+string(os.PathListSeparator)+binPath)
+
+	_, err := exec.LookPath(file)
+	return err
+}
+
+// AssertExecutableInPATH asserts that the executable file is in bin path.
+func (it *ITest) AssertExecutableInPATH(file string) {
+	if err := it.lookExecutable(file); err != nil {
```
corneliusweig: That would be a good candidate for `t.Helper()`

test/krew/krew.go
```diff
@@ -32,51 +33,82 @@ const krewBinaryEnv = "KREW_BINARY"
...
+		it.t.Fatalf("executable %s not in PATH: %+v", file, err)
+	}
+}
+
+// AssertExecutableNotInPATH asserts that the executable file is not in bin
+// path.
+func (it *ITest) AssertExecutableNotInPATH(file string) {
```
corneliusweig: That would be a good candidate for `t.Helper()`

test/krew_test.go
```diff
@@ -57,7 +57,7 @@ func TestKrewInstall(t *testing.T) {
 	defer cleanup()
 
 	test.WithIndex().Krew("install", validPlugin).RunOrFailOutput()
-	test.Call(validPlugin, "--help").RunOrFail()
+	test.AssertExecutableInPATH("kubectl-" + validPlugin)
```
corneliusweig: Big :+1: for this!

test/krew/krew.go
ahmetb: I think it's ok, we might introduce too much magic if we move. It's not too repetitive for now.

test/krew/krew.go
ahmetb: nice catch.

hack/run-integration-tests.sh
ahmetb: You're right, this is me being an idiot and accepting both commits in merge-conflict.

ahmetb:
Addressed code review feedback. 👍 Thanks!

ahmetb:
🔜

PR #235, opened 2019-07-02
Move to go modules (vgo)

### LINKED ISSUE #234: Migrate to go modules
Now that kubernetes client-go actually supports Go modules, it should be possible and it can simplify the dependency management (i.e. adding new deps) just a little bit.

That said, it complicates the Travis CI builds, as travis still sets GOPATH and requires us to update our build command to make sure it uses vgo.

/kind cleanup

ahmetb:
Fixes #234.
/assign @corneliusweig


corneliusweig:
This PR is too big, I can't add inline comment :laughing:

- `.travis.yml/GOPROXY`: I can't find that in the travis guide for building go projects (https://docs.travis-ci.com/user/languages/go/). Why do you set that here?
- PR #195 added github.com/google/go-cmp v0.3.0. That's what causes the conflicts.
- The `gox` install hack really looks nasty, as it is even inlined in `.travis.yml`. If you don't see the chance we can get rid of that soonish, I'd prefer a proper script there.

ahmetb:
> * `.travis.yml/GOPROXY`: I can't find that in the travis guide for building go projects ([docs.travis-ci.com/user/languages/go](https://docs.travis-ci.com/user/languages/go/)). Why do you set that here?

Anytime you fetch go modules, using GOPPROXY is much faster. See http://proxy.golang.org/. This speeds up the test suite. e.g.

```
time go mod download => 27s
go clean -modcache -cache
time env GOPROXY=https://proxy.golang.org go mod download => 4s.
```

> That's what causes the conflicts.

Addressed merge conflicts.

> * The `gox` install hack really looks nasty, as it is even inlined in `.travis.yml`. If you don't see the chance we can get rid of that soonish, I'd prefer a proper script there.

Solved this by introducing a `hack/install-gox.sh`.

PR #237, opened 2019-07-02
pkg/installation: reorder receipt save & install()

ahmetb:
Looks like we didn't address the code review comment at
https://github.com/kubernetes-sigs/krew/pull/195#discussion_r296915679.

Copying the conversation from there:

> This is better because:
>
> - actually installing a plugin (extract files, symlink binary) but then failing
>   to store the receipt is okay (at most you'd be leaking some files)
> - claiming the plugin is installed (i.e. receipt exists) while it's not
>   installed is not okay.
> - tens of things can go wrong in `install()`, much fewer things can go wrong
>  in `receipt.Store()`, so makes sense to do it last.

/assign @corneliusweig
/kind cleanup

### DIFF
--- a/pkg/installation/install.go
+++ b/pkg/installation/install.go
@@ -78,16 +78,15 @@ func Install(p environment.Paths, plugin index.Plugin, forceDownloadFile string)
 		return err
 	}
 
-	glog.V(2).Infof("Storing install receipt for plugin %s", plugin.Name)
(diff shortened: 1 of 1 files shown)


corneliusweig:
Woops, sorry. Looks like I was reordering the uninstall procedure instead of the install procedure (7e7cf0b455729e9ffeaf7a9273879dfb802925b9). Do you want to swap the uninstall order back to original?

ahmetb:
> Do you want to swap the uninstall order back to original?

Done. PTAL.

PR #242, opened 2019-07-02
Add integration test for `krew upgrade`

### LINKED ISSUE #233: Krew upgrade needs integration test
Krew upgrade is currently not covered by integration tests, because it requires a more advanced test setup. What the integration test should do:

- Install an old plugin version
- Verify that the old version is installed
- Run `krew upgrade`
- Verify that the new version is installed

The tricky part is to reliably get two plugin versions in the test. One way to do this, could be to install a fixture version with the `--manifest` option. To verify that the original version and the eventual version are actually different, one could take the hash of a file that must change during upgrade.

corneliusweig:
Only happy path so far.

Close #233

### DIFF
--- a/integration_test/testutil_test.go
+++ b/integration_test/testutil_test.go
@@ -117,30 +117,31 @@ func lines(in []byte) []string {
 	return strings.Split(trimmed, "\n")
 }
 
-func (it *ITest) lookupExecutable(file string) error {
+func (it *ITest) lookupExecutable(file string) (string, error) {
 	orig := os.Getenv("PATH")
 	defer func() { os.Setenv("PATH", orig) }()
 
 	binPath := filepath.Join(it.Root(), "bin")
 	os.Setenv("PATH", binPath)
 
-	_, err := exec.LookPath(file)
-	return err
+	return exec.LookPath(file)
 }
 
-// AssertExecutableInPATH asserts that the executable file is in bin path.
-func (it *ITest) AssertExecutableInPATH(file string) {
+// AssertExecutableInPATH asserts that the executable file is in bin path and returns its location.
+func (it *ITest) AssertExecutableInPATH(file string) string {
 	it.t.Helper()
-	if err := it.lookupExecutable(file); err != nil {
+	path, err := it.lookupExecutable(file)
+	if err != nil {
 		it.t.Fatalf("executable %s not in PATH: %+v", file, err)
 	}
+	return path
 }
 
 // AssertExecutableNotInPATH asserts that the executable file is not in bin
 // path.
 func (it *ITest) AssertExecutableNotInPATH(file string) {
 	it.t.Helper()
-	if err := it.lookupExecutable(file); err == nil {
+	if _, err := it.lookupExecutable(file); err == nil {
 		it.t.Fatalf("executable %s still exists in PATH", file)
 	}
 }
--- a/integration_test/testutil_test.go
+++ b/integration_test/testutil_test.go
@@ -117,7 +117,7 @@ func lines(in []byte) []string {
 	return strings.Split(trimmed, "\n")
 }
 
-func (it *ITest) lookupExecutable(file string) (string, error) {
+func (it *ITest) LookupExecutable(file string) (string, error) {
 	orig := os.Getenv("PATH")
 	defer func() { os.Setenv("PATH", orig) }()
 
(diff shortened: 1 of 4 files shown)


integration_test/testutil_test.go
```diff
@@ -117,30 +117,31 @@ func lines(in []byte) []string {
...
 
-	_, err := exec.LookPath(file)
-	return err
+	return exec.LookPath(file)
 }
 
-// AssertExecutableInPATH asserts that the executable file is in bin path.
-func (it *ITest) AssertExecutableInPATH(file string) {
+// AssertExecutableInPATH asserts that the executable file is in bin path and returns its location.
+func (it *ITest) AssertExecutableInPATH(file string) string {
```
corneliusweig: I'm not 100% happy that this function returns something, on the other hand it should be quite clear what it is. WDYT?

integration_test/upgrade_test.go
```diff
@@ -14,4 +14,50 @@
...
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	cwd, err := os.Getwd()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	test.WithIndex().
+		Krew("install", "--manifest", filepath.Join(cwd, "testdata", "konfig.yaml")).
```
ahmetb: maybe use the const here for plugin name.

integration_test/upgrade_test.go
```diff
@@ -14,4 +14,50 @@
...
+		RunOrFail()
+	location := test.AssertExecutableInPATH("kubectl-" + validPlugin)
+	initialHash := hashFile(t, location)
```
ahmetb: I’m -1 for reading file contents. Even if you did, you don't need a hash like sha256, you can just `reflect.DeepEquals` the bytes –or even simply compare `len(bytes)`. :)

Also I don't want to rely on os.Open behavior on symlinks, though I know open() follows symlinks. 

What I would do here is a simple `readlink()` and see if the symlink paths are the same. We know that upon upgrade, they would change.

(I realize that you can make an argument here about testing the behavior; not the implementation. And comparing bytes directly is perhaps a better idea.)

ahmetb: Another option is this:

```
fi1 := os.Lstat("kubectl-foo")
...
fi2 := os.Lstat("kubectl-bar")
if os.SameFile(fi1,fi2) {...
```

integration_test/testutil_test.go
ahmetb: I think I wouldn't change the signature here.  You can just drop the return value. Assertion methods typically don't return things.

integration_test/upgrade_test.go
```diff
@@ -14,4 +14,50 @@
...
+	"testing"
+)
+
+func TestKrewUpgrade(t *testing.T) {
+	skipShort(t)
+
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	cwd, err := os.Getwd()
```
ahmetb: you don't need CWD, you can just do `filepath.Join("testdata","konfig.yaml")` –although I’m not sure if `go test` sets any CWD. If we do this:


    cd hack
    run-integration-tests.sh

this might not work.

corneliusweig: Indeed, `cwd` is obsolete here. I checked that it works when running the tests from different dirs as well.

integration_test/upgrade_test.go
corneliusweig: I went with the readlink approach because `Lstat` may have problems when one of the files has been deleted in the meantime.

integration_test/testutil_test.go
corneliusweig: I totally agree. I exposed `LookupExecutable` instead.

corneliusweig:
Thanks for your suggestions, looks much better now. PTAL

integration_test/upgrade_test.go
```diff
@@ -14,4 +14,44 @@
...
+
+	test.Krew("upgrade").RunOrFail()
+	eventualLocation := realLocation(test, validPlugin)
+
+	if initialLocation == eventualLocation {
+		t.Errorf("Expecting the plugin path to change but was the same.")
+	}
+}
+
+func realLocation(test *ITest, plugin string) string {
```
ahmetb: resolvePluginSymlink is a better name.
"real" has different connotations like realpath(2).

integration_test/testutil_test.go
```diff
@@ -117,21 +117,20 @@ func lines(in []byte) []string {
 	return strings.Split(trimmed, "\n")
 }
 
-func (it *ITest) lookupExecutable(file string) error {
+func (it *ITest) LookupExecutable(file string) (string, error) {
```
ahmetb: is there a particular reason to export this method? it's all in the same package.

ahmetb:
/approve
sorry for back-and-forth, will /lgtm once the comments are addressed.

integration_test/testutil_test.go
corneliusweig: No particular reason. I found it more consistent with the other methods on `ITest`, which are all exposed when used outside of `testutil_test.go`.

integration_test/upgrade_test.go
corneliusweig: Yes, so true..

integration_test/testutil_test.go
ahmetb: OK it's not a big issue.

corneliusweig:
> sorry for back-and-forth

Don't be! Your reviews are very valuable.

PR #243, opened 2019-07-03
Ensure that go modules are enabled in all helper scripts

corneliusweig:
Before, these scripts failed on my system when run individually.


hack/run-lint.sh
```diff
@@ -16,6 +16,10 @@
 
 set -euo pipefail
 
+[[ -n "${DEBUG:-}" ]] && set -x
+
+export GO111MODULE=on
```
ahmetb: I personally prefer 

env a=b cmd args...

syntax as it restricts the scope of the variable and avoids other side effects. 
For example if we called install-gox.sh from one of these scripts, it would inherit this variable, and it would be potentially an issue.

corneliusweig: Fair point. I limited the scope to the essential parts.

hack/run-integration-tests.sh
```diff
@@ -47,4 +47,4 @@ fi
 KREW_BINARY="${krew_binary_realpath}"
 export KREW_BINARY
 
-go test -test.v sigs.k8s.io/krew/integration_test
+GO111MODULE=on go test -test.v sigs.k8s.io/krew/integration_test
```
ahmetb: I’d use “env A=B cmd” but in this case it makes no difference.

Issue #131, opened 2018-11-07
Question on bin files without .exe suffix


ahmetb:
If a plugin's windows `platform` spec specifies something like `bin: foo`, we:

1. copy the executable to plugin's installation dir as `foo`
2. but we symlink it as kubectl-foo.exe on windows.

I am not at all sure that `kubectl-foo.exe` which points to `foo` (that has no .exe extension) can be executed. Can someone verify if this works or not?

Depending on that, (1) we need to make sure if a platform spec matches `windows`, its `bin` must end with ".exe" (2) we need to update existing plugin specs in krew-index.

juanvallejo:
I think for windows, having a symlink that ends in ".exe" should be fine. We don't follow symlinks to check the original file's extension. This being said, I'll try to double check that this works on windows

ahmetb:
Yeah the part I am not sure is whether execve equivalent of Windows actually runs files without an executable suffix, my guess would be not.

ahmetb:
/remove-lifecycle stale
I plan on checking this + krew self-upgrade code-path on windows sometime.

ahmetb:
yeah the /label is failing us on default labels that aren't tracked by the test-infra here.

ahmetb:
Just had the time to verify this on windows:

The target of the symlink (the file it's pointing to) does not have to have an executable suffix (like `.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.PSC1`).

So the value in the `bin` field doesn't have to have an .exe suffix, because our symlink in %HOME%\.krew\bin will already have an .exe suffix.

That said, since we add .exe to every executable, if your plugin is actually not a win32 application, it will not work. For example, plugins that have a .ps1 or .bat extension currently don't work on Krew.

I'm thinking possibly we need to preserve the file extension based on the `platform.bin` field field on Windows. 
/cc @corneliusweig

ahmetb:
The only clean solution that comes to my mind is that

- `bin` field must contain a file extension if platform selector matches `os:windows` (enforce this only in cosmetic validation, not parsing, so that existing plugins continue to work, and any modified plugin spec must comply)

[...]

PR #245, opened 2019-07-04
Add integration tests for `krew install` with --archive and --manifest options

### LINKED ISSUE #244: integration_test: add test cases for install --manifest/--archive options
/kind cleanup
/priority important-soon

corneliusweig:
Fix #244


integration_test/testdata/test-foo.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
```
ahmetb: 🤔 can we symlink these from existing testdata dir?

integration_test/testdata/konfig.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
```
ahmetb: I wonder if we can use the index prep  on each test and use this file from there? Or http.Get this from github?
I am not against this but the amount of testdata files might soon get out of hand.

integration_test/install_test.go
```diff
@@ -58,7 +58,7 @@ func TestKrewInstall_ManifestAndArchive(t *testing.T) {
 		"--manifest", filepath.Join("testdata", fooPlugin+constants.ManifestExtension),
 		"--archive", filepath.Join("testdata", fooPlugin+".tar.gz")).
 		RunOrFail()
-	test.AssertExecutableInPATH("kubectl-" + strings.ReplaceAll(fooPlugin, "-", "_"))
```
ahmetb: we can bump to 1.12 if necessary.

corneliusweig: That's a good idea anyway. But I would not do it for such a small inconvenience.

integration_test/testdata/test-foo.yaml
corneliusweig: So I looked at the existing test-data we have and I could find nothing with valid yaml+archive. I think it would be possible without having these testdata in the repo, but the cost will be a more complex test setup. So my current gut feeling is that these new testdata will cause the smallest pain in the future.

ahmetb: You're right, it seems like all testdata dirs are unit-test oriented and incomplete.

I think you can create a very simple "foo" plugin in `integration_test/testdata` here with a tarballed bash script, and a manifest for it, and just use it to test.

integration_test/testdata/konfig.yaml
corneliusweig: In principle, I like the idea to take this from the index prep. However, this particular yaml is an old version which is also needed in the `krew upgrade` test #242 (I made sure it's the same). So I'm reusing this file from there. Besides, having this yaml in the repo makes it more visible which will make test maintenance simpler. So again, I'm still inclined to keep the testdata as-is.

integration_test/testdata/test-foo.yaml
corneliusweig: Yes, I called it `test-foo`, but I can rename to `foo` as well.

integration_test/install_test.go
corneliusweig: Renaming from `test-foo` to `foo` even made this `ReplaceAll` obsolete.

integration_test/testdata/konfig.yaml
ahmetb: OK SG.

integration_test/install_test.go
```diff
@@ -25,3 +34,42 @@ func TestKrewInstall(t *testing.T) {
...
+}
+
+func TestKrewInstall_ManifestAndArchive(t *testing.T) {
+	skipShort(t)
+
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	test.Krew("install",
+		"--manifest", filepath.Join("testdata", fooPlugin+constants.ManifestExtension),
```
ahmetb: You don't need to address this but: I feel like these small constants are hurting the readability quite a bit, unless needed, we can probably avoid in the future.

integration_test/testdata/foo.yaml
```diff
@@ -0,0 +1,28 @@
...
+      bin: foo
+      files:
+        - from: ./foo
+          to: "."
+      selector:
+        matchExpressions:
+        - key: os
+          operator: In
+          values: ["darwin", "linux"]
+    - uri: https://foo.bar/foo.tar.gz
```
ahmetb: I recommend removing the windows section. (We strive to keep testdata manifests to only what's being tested.)

You can make a select-all OS selector like:

```
matchLabels:
  os: {}
```

this should match all OSes.

corneliusweig: Yes, you are right. My thinking was to make sure that the test would run on Windows (also see #131), but that does not depend on the file ending.

ahmetb: I think regardless we make a symlink called kubectl-foo.exe anyway. That said I'm not sure if that would actually work when the target file is not named .exe/.com/.bat/.ps1 etc. If you create a `foo.tar.gz` with foo.sh and foo.bat, this would probably be more straightforward.

corneliusweig: Ok, so I added a windows version of the test-foo plugin. I think that should do it.

PR #246, opened 2019-07-05
Added glog.v=4 in the test script

### LINKED ISSUE #226: hack/run-integration-tests: how to use glog
I see some glog.V() statements in the integration tests but they don't show up in the output. Should we set some -glog.v=4 in the script?

hpandeycodeit:
Fixes #226


hack/run-integration-tests.sh
```diff
@@ -47,4 +47,5 @@ fi
 KREW_BINARY="${krew_binary_realpath}"
 export KREW_BINARY
 
+-glog.v=4
```
ahmetb: You sure you added it to the right place?

hpandeycodeit: I think. Maybe the other option is to add with the test command, something like this:

```
go test -test.v sigs.k8s.io/krew/integration_test v 4
```

ahmetb: I feel like you aren’t sure :)
You just did what I described in the issue, so you just typed -glog.v=4 in the script on a new line, which won’t work. 
I don’t think your suggestion of adding `v 4`  to the go test as positional arguments  will work either.

Clearly the tests are also failing. Please test this locally by running the command.

ahmetb:
Thanks for the -args change. It doesn’t seem like it’s printing any of the glog statements from the test code according to Travis CI output.  It seems like it’s still not working. Maybe the arg name is -v and not -glog.v? Please try locally. 

@corneliusweig maybe we should purge the usage of glog in _test files? I feel like t.Logf is sufficient (and Go already knows how to suppress it when tests are sucessful).

corneliusweig:
Yes, purging glog from the tests sounds like the right thing to do. I think there would be very little use for different logging levels other than `on` and `off`.

ahmetb:
/close
yeah let's remove/prevent glog usage in test files instead. opened #247 to track.

PR #248, opened 2019-07-08
Use `t.Logf` instead of `glog` in tests

### LINKED ISSUE #247: integration_test: prevent usage of glog, move to t.Logf
Transfer from #226 per https://github.com/kubernetes-sigs/krew/pull/246#issuecomment-509025098

1. we should remove glog usages from test files
2. we should add a check to `hack/verify-code-patterns.sh` to find and disallow usage of `[gk]log\.` and ` in *_test.go files.

corneliusweig:
Fix #247

### DIFF
--- a/hack/verify-code-patterns.sh
+++ b/hack/verify-code-patterns.sh
@@ -24,6 +24,7 @@ if [[ -n "$out" ]]; then
   exit 1
 fi
 
+# use code constant for ".yaml"
 out="$(grep --include '*.go' \
             --exclude "*_test.go" \
             --exclude 'constants.go' \
@@ -31,6 +32,14 @@ out="$(grep --include '*.go' \
             -EIrn '\.yaml"' || true)"
 if [[ -n "$out" ]]; then
   echo >&2 'You used ".yaml" in production, use constants.ManifestExtension instead:'
+  echo >&2 "$out"
+  exit 1
+fi
+
+# Do not use glog in test code
+out="$(grep --include '*_test.go' --exclude-dir 'vendor/' -EIrn 'github.com/golang/glog' || true)"
+if [[ -n "$out" ]]; then
+  echo >&2 "You used glog in tests, use 't.Logf' instead:"
   echo >&2 "$out"
   exit 1
 fi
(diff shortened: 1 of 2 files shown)


hack/verify-code-patterns.sh
```diff
@@ -34,3 +35,11 @@ if [[ -n "$out" ]]; then
   echo >&2 "$out"
   exit 1
 fi
+
+# Do not use glog in test code
+out="$(grep --include '*_test.go' --exclude-dir 'vendor/' -EIrn 'github.com/golang/glog' || true)"
```
ahmetb: I recommend making the pattern based on invocation `[kg]log\.` as we might move to klog at some point (in fact we probably should). Similarly, import paths are more likely to change than the usage (e.g. this could change to `golang.org/glog` someday).

integration_test/install_test.go
```diff
@@ -25,3 +34,42 @@ func TestKrewInstall(t *testing.T) {
 	test.WithIndex().Krew("install", validPlugin).RunOrFailOutput()
 	test.AssertExecutableInPATH("kubectl-" + validPlugin)
 }
+
+func TestKrewInstall_Manifest(t *testing.T) {
```
ahmetb: I think the other PR got mixed in here :)

integration_test/testutil_test.go
```diff
@@ -166,7 +165,7 @@ func (it *ITest) WithIndex() *ITest {
 // WithEnv sets an environment variable for the krew run.
 func (it *ITest) WithEnv(key string, value interface{}) *ITest {
 	if key == "KREW_ROOT" {
-		glog.V(1).Infoln("Overriding KREW_ROOT in tests is forbidden")
+		it.t.Log("Overriding KREW_ROOT in tests is forbidden")
```
ahmetb: out of scope but I think this is more suitable for a `t.Fatal`

corneliusweig: It's a small but important thing, so I changed that.

integration_test/install_test.go
corneliusweig: gosh, how embarrassing :roll_eyes:

hack/verify-code-patterns.sh
corneliusweig: My thought was to avoid false positives as `glog` is rather short. But for now there are no collisions and we can think of a better way if a collision should occur. :+1:

PR #249, opened 2019-07-09
Add subcommand for migrating the krew home

### LINKED ISSUE #232: Migrate krew home to add install receipts for installed plugins
After #195, `krew` adds install receipts when installing new plugins. After that, the subcommands `list`, `uninstall`, and `upgrade` should make use of the install receipts instead of scanning the `$KREW_ROOT/bin` folder.

This issue is to discuss the logic how the migration can be done.

---

This could work:

1. Detect if the krew home was migrated before. This can be done by 
    - _not_ creating the `$KREW_ROOT/receipt` folder by default and see if it exists
    - look for a special file in `$KREW_ROOT/layout_version` or similar
2. If the krew home was migrated before, do nothing.
    If the krew home has not been migrated, print a warning to run `krew upgrade` and quit.
3. When `krew upgrade` is run, upgrade all plugins. For plugins that don't need upgrade, save the manifest in `$KREW_ROOT/receipt`.

After that, the receipt folder will be up to date.

The command `krew upgrade` is necessary, as it is impossible to find out the plugin manifest when the installed version lags behind the current manifest version. This as a big downside: the logic for `krew upgrade` cannot be migrated to work with install receipts anytime soon. Or at least `krew upgrade` needs to preserve the scanning mode for some time.

WDYT?

corneliusweig:
This PR adds a new subcommand `krew system` following the logic outlined in https://github.com/kubernetes-sigs/krew/issues/232#issuecomment-507481767.

The basic steps are:

* add a check to detect if the krew home needs to be migrated. This check runs before every command.
* add migration logic
  - build list of plugins by looking at the folders in `$KREW_ROOT/store` and perform some validation on them
  - remove the link in `$KREW_ROOT/bin` and the folder in `$KREW_ROOT/store/<plugin>`.
  - call `kubectl krew install <plugin>` if the previous steps were successful
* manually copy over the `krew.yaml` manifest, because that cannot be re-installed (and we know that this is the correct manifest).

**Notes**:

1. I'm not 100% happy with the method how we detect an outdated krew home. One major drawback is that new users will have to run the migration as the very first thing.

2. So far, I didn't implement as many component tests as I normally would. However this is throw-away code and covered by integration tests. So I think this is acceptable.

3. Because we want to make the migration code independent from the main logic, I duplicated a lot of code (mainly from `pkg/installation/install.go`). This was intentional.

Close #232


ahmetb:
I'll try to find time to review this.


> * I'm not 100% happy with the method how we detect an outdated krew home. One major drawback is that new users will have to run the migration as the very first thing.

this is ok

> * So far, I didn't implement as many component tests as I normally would. However this is throw-away code and covered by integration tests. So I think this is acceptable.

we don't need tests for this, but we have to do quite a bit of manual verification.

> * add migration logic
>   
>   * build list of plugins by looking at the folders in `$KREW_ROOT/store` and perform some validation on them
>   * remove the link in `$KREW_ROOT/bin` and the folder in `$KREW_ROOT/store/<plugin>`.
>   * call `kubectl krew install <plugin>` if the previous steps were successful

this SGTM. I was assuming we can do `krew uninstall` but we can’t. Makes sense.



> * manually copy over the `krew.yaml` manifest, because that cannot be re-installed (and we know that this is the correct manifest).

hmm this feels hacky. but might be OK.

cmd/krew/cmd/root.go
```diff
@@ -69,7 +79,9 @@ func init() {
 		paths.DownloadPath(),
 		paths.InstallPath(),
 		paths.BinPath(),
-		paths.InstallReceiptPath()); err != nil {
+	// TODO(corneliusweig): re-enable when migration code is removed:
+	//  paths.InstallReceiptPath(),
```
ahmetb: I think you can still enable it.
We just need to find a different way to detect. An example could be: :the number of directories in `store/` are non-zero but `*.yaml`s in `receipts/` are zero" is cleaner and doesn't require you to do this.

cmd/krew/cmd/system.go
```diff
@@ -0,0 +1,46 @@
+// Copyright © 2019 NAME HERE <EMAIL ADDRESS>
```
ahmetb: fix.

cmd/krew/cmd/system.go
```diff
@@ -0,0 +1,46 @@
...
+	"fmt"
+
+	"sigs.k8s.io/krew/pkg/migration"
+
+	"github.com/spf13/cobra"
+)
+
+// systemCmd represents the system command
+var systemCmd = &cobra.Command{
+	Use:   "system receipts-upgrade",
```
ahmetb: can't we nest subcommands? this seems a bit hacky. we should let cobra handle non-existing subcmds etc for us.

cmd/krew/cmd/system.go
```diff
@@ -0,0 +1,46 @@
...
+	Long: `Perform krew maintenance tasks such as migrating to a new krew-home layout.
+
+This command will be removed without further notice from future versions of krew.
+`,
+	Args: cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if args[0] != "receipts-upgrade" {
+			return fmt.Errorf("Only subcommand `receipts-upgrade` is supported.")
+		}
+		return migration.DoMigration(paths)
```
ahmetb: DoReceiptsMigration is better.

integration_test/system_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+	test.WithIndex().Krew("install", validPlugin).RunOrFail()
+
+	// needs to be after initial installation
+	prepareOldKrewRoot(test)
+
+	test.Krew("system", "receipts-upgrade").RunOrFailOutput()
+	test.AssertExecutableInPATH("kubectl-" + validPlugin)
+
+	assertReceiptExistsFor(test, validPlugin)
```
ahmetb: nit: should this be part of ITest? this way you can do: 

`test.assertReceiptExistsFor(test, validPlugin)` just like  `test.AssertExecutableInPATH("kubectl-" + validPlugin)` 

also note that it's inconsistent one is exported, the other one isn't.

pkg/migration/migration.go
```diff
@@ -0,0 +1,212 @@
...
+	"sigs.k8s.io/krew/pkg/environment"
+	"sigs.k8s.io/krew/pkg/index"
+	"sigs.k8s.io/krew/pkg/migration/oldenvironment"
+)
+
+const (
+	krewPluginName = "krew"
+)
+
+func IsMigrated(newPaths environment.Paths) bool {
```
ahmetb: Consider making `pkg/migration/receiptsmigration` OR name every method like `DoneReceiptsMigration()` to clarify _which_ migration it refers to.

ditto below.

pkg/migration/migration.go
```diff
@@ -0,0 +1,212 @@
...
+	_, err := os.Lstat(newPaths.InstallReceiptPath())
+	return err == nil
```
ahmetb: I think some errors can be legit FS errors, although they're rare. using `os.IsNotExist(err)` is often better in these situations.

But since there's a better logic to detect this, don't worry.

pkg/migration/migration.go
```diff
@@ -0,0 +1,212 @@
...
+
+// DoMigration searches for installed plugins, removes each plugin and reinstalls afterwards.
+// Once started, it keeps going even if there are errors.
+func DoMigration(newPaths environment.Paths) error {
+	if IsMigrated(newPaths) {
+		glog.Infoln("Already migrated")
+		return nil
+	}
+
+	oldPaths := oldenvironment.MustGetKrewPaths()
```
ahmetb: why do we need `oldenvironment` package?

Can't we just get KREW_ROOT from the current `environment` pkg and do `filepath.Join(root, "store")` etc? This `oldenvironment` adds a lot of code, I recommend just going for straight brain surgery on version==0.2.x assumption here and "assume" paths.

cmd/krew/cmd/root.go
corneliusweig: Yes, perfect. That'll do.

integration_test/system_test.go
corneliusweig: That depends on whether we want to keep that assertion method. Also, I would consider receipts more of an implementation detail. It's necessary to assert on them in the current test-case, but should we encourage doing so in other integration tests?

pkg/migration/migration.go
corneliusweig: Changing the package sounds good.

pkg/migration/migration.go
corneliusweig: My thinking was
- it eases review, because one can do file-diff.
- I don't have to bother about correctness wrt `KREW_ROOT` or so, because it's all tried and tested.
- it raises the abstraction level of the migration code.

Granted, it adds more code than necessary, but it's going to be deleted anyways.

cmd/krew/cmd/system.go
corneliusweig: I went with your other recommendation and renamed the package. Now it is `receiptsmigration.Migrate`.

cmd/krew/cmd/system.go
corneliusweig: Yes, good idea.

pkg/migration/migration.go
ahmetb: I'm OK with it. Don't forget to add TODOs to the methods and packages we must delete. Granted, we'll probably delete this code in v0.4.

pkg/testutil/tempdir.go
```diff
@@ -71,3 +71,8 @@ func (td *TempDir) Write(file string, content []byte) *TempDir {
 	}
 	return td
 }
+
+// Touch creates a file without content in the temporary directory.
+func (td *TempDir) Touch(file string) *TempDir {
```
ahmetb: let's not add this here just because you're using it from migration code.

cmd/krew/cmd/system.go
ahmetb: don't forget to change the string :P

integration_test/system_test.go
ahmetb: in that case  we should add a TODO to all these tests + helpers to remove them after the migration code is scrubbed.

pkg/receiptsmigration/migration.go
```diff
@@ -0,0 +1,236 @@
...
+		return errors.Wrapf(err, "failed to create directory %q", newPaths.InstallReceiptPath())
+	}
+
+	// krew must be skipped by the normal migration logic
+	if err := copyKrewManifest(newPaths.IndexPluginsPath(), newPaths.InstallReceiptPath()); err != nil {
+		return errors.Wrapf(err, "failed to copy krew manifest")
+	}
+
+	for _, plugin := range installed {
+		if consistent := isInstallationConsistent(oldPaths.BinPath(), plugin); !consistent {
```
ahmetb: why would this case happen?

pkg/receiptsmigration/migration.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	oldPaths := oldenvironment.MustGetKrewPaths()
+	installed, err := getPluginsToReinstall(oldPaths, newPaths)
+	if err != nil {
+		return errors.Wrapf(err, "failed to build list of plugins")
+	}
+
+	glog.Infoln("These plugins will be reinstalled: ", installed)
+	if err := os.MkdirAll(newPaths.InstallReceiptPath(), 0755); err != nil {
```
ahmetb: you shouldn't need to Mkdir here if you install other plugins first (which would do `krew install` that creates this directory), and then copyKrewManifest over.

That way you're not duplicating any logic like the chmod of this directory.

ahmetb:
I think at this point we should probably just do manual testing.

cmd/krew/cmd/system.go
corneliusweig: :man_facepalming:

pkg/migration/migration.go
corneliusweig: My hope was that we simply revert the whole commit. I find todos on every method a little excessive (especially since I'm not a big fan of todos). Can we take a middle ground where there's only one todo for a whole package, when the package should be deleted, and one todo for a function outside of such packages?

pkg/testutil/tempdir.go
corneliusweig: Agreed. I imagine we will have other uses for this method, but this PR is not a good place to add it.

pkg/receiptsmigration/migration.go
corneliusweig: I wanted to guard against the case where a user has added other files in `store` or `bin`. But probably the check that the plugin is available in the index already covers that well enough. I'll remove the check.

pkg/receiptsmigration/migration.go
corneliusweig: Oh, this was a leftover of my flawed migration indicator. `cmd/krew/cmd/root.go:70` will already ensure that all environment paths are created, so this is obsolete.  Thanks for spotting this!

ahmetb:
/approve
I added #255 that addresses the need for a script that ensures as we refactor more parts, the migration works. PTAL.

PR #252, opened 2019-07-12
krew v1.11 kubectl usage

kenichi-shibata:
Documentation add


corneliusweig:
I'm currently -1 for this change. It does not say that every plugin installed via `krew` then needs to be called as `kubectl-plugin` instead of `kubectl plugin` as well. And even if it would say so, this would make for a quite bad user experience.

ahmetb:
-2
krew doesn’t do anything for kubectl 1.11. Installed plugins are supported only on 1.12 or after because the plugin system is completely different in 1.11 or earlier.

PR #255, opened 2019-07-12
hack: add script to verify receipts migration

ahmetb:
This script makes sure as long as we have the 'system receipts-upgrade' command
upgrades from krew v0.2.1 to the currently built version works.

This should be deleted once the 'system receipts-upgrade' command is gone.

/hold
rebase + merge after #249 (tests expected to fail until then)

### DIFF
--- a/hack/verify-receipts-upgrade-migration.sh
+++ b/hack/verify-receipts-upgrade-migration.sh
@@ -1,0 +1,147 @@
+#!/usr/bin/env bash
+
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script tests "krew system receipts-migration" which was introduced for
+# migrating krew 0.2.x to krew 0.3.x.
+#
+# TODO(ahmmetb,corneliusweig) remove at/after krew 0.4.x when receipts-migration
+# is no longer supported.
+
+set -euo pipefail
+
+[[ -n "${DEBUG:-}" ]] && set -x
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+BINDIR="${SCRIPTDIR}/../out/bin"
+goos="$(go env GOOS)"
+goarch="$(go env GOARCH)"
+
+install_krew_0_2_1() {
+    # from https://github.com/kubernetes-sigs/krew/blob/v0.2.1/README.md
+    krew_root="${1}"
+    temp_dir="$(mktemp -d)"
+    trap 'rm -rf "${temp_dir}"' RETURN
+    (
+        cd "${temp_dir}"
+        curl -fsSLO "https://storage.googleapis.com/krew/v0.2.1/krew.{tar.gz,yaml}"
+        tar zxf krew.tar.gz
+        env KREW_ROOT="${krew_root}" ./krew-"${goos}_amd64" install \
+            --manifest="./krew.yaml" --archive="./krew.tar.gz"
+    )
+}
+
+install_plugin() {
+    krew_root="${1}"
+    plugin="${2}"
+
+    run_krew "${1}" install "${plugin}" 1>/dev/null
+}
+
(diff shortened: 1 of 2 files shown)


hack/verify-receipts-upgrade-migration.sh
```diff
@@ -0,0 +1,132 @@
...
+    env KREW_ROOT="${krew_root}" \
+        PATH="${krew_root}/bin:$PATH" \
+        kubectl krew "$@"
+}
+
+verify_plugin_installed() {
+    krew_root="${1}"
+    plugin="${2}"
+    local plugin_bin_name
+    plugin_bin_name="kubectl-${plugin/-/_}"
```
corneliusweig:
```suggestion
    plugin_bin_name="kubectl-${plugin//-/_}"
```
Otherwise only the first occurrence of `-` will be replaced

hack/verify-receipts-upgrade-migration.sh
```diff
@@ -0,0 +1,132 @@
...
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+BINDIR="${SCRIPTDIR}/../out/bin"
+goos="$(go env GOOS)"
+goarch="$(go env GOARCH)"
+
+install_krew_0_2_1() {
+    # from https://github.com/kubernetes-sigs/krew/blob/v0.2.1/README.md
+    krew_root="${1}"
+    cd "$(mktemp -d)"
```
corneliusweig: Can you do `pushd "$(mktemp -d)"` and `popd` instead?

Also, it would be nicer to clean up the temporary directory as you do it below
```
trap 'rm -rf "${temp_dir}"' RETURN
```

hack/verify-receipts-upgrade-migration.sh
```diff
@@ -0,0 +1,132 @@
...
+    curl -fsSLO "https://storage.googleapis.com/krew/v0.2.1/krew.{tar.gz,yaml}"
+    tar zxf krew.tar.gz
+    env KREW_ROOT="${krew_root}" ./krew-"${goos}_amd64" install \
+        --manifest="./krew.yaml" --archive="./krew.tar.gz"
```
corneliusweig: nit: it would work with just the `--manifest` option, but no need to change.

ahmetb: I could just wget the manifest from krew-index repo and install with that but wanted to stick to what users do.

hack/verify-receipts-upgrade-migration.sh
ahmetb: pushd/popd shouldn't really be necessary since "cd" is function-scoped.
adding the trap.

ahmetb:
I couldn't think of any additional test cases, so I threw `krew search` and `krew list` in there just to make sure we can read manifest files and read receipts.

hack/verify-receipts-upgrade-migration.sh
corneliusweig: If the function is called in a sub-shell, `cd` wont have global effects. But here, `cd` is not confined to the function. For example, this should be technically the same:
```bash
#!/bin/bash

ls

do_sth() {
	echo "do_sth called"
	cd /
	ls
}

do_sth
ls
```
Here, the third `ls` prints the same as the second (with bash-4).

It doesn't really matter for the correctness of this script, but it's still something to consider.

hack/verify-receipts-upgrade-migration.sh
```diff
@@ -0,0 +1,146 @@
...
+        exit 1
+    fi
+
+    krew_root="$(mktemp -d)"
+    trap 'rm -rf "${krew_root}"' RETURN
+
+    echo >&2 "Test directory: ${krew_root}"
+    install_krew_0_2_1 "${krew_root}"
+    install_plugin "${krew_root}" "get-all"
+    install_plugin "${krew_root}" "get-all"
```
corneliusweig: Is it intentional, that this line is duplicated?

hack/verify-receipts-upgrade-migration.sh
```diff
@@ -0,0 +1,146 @@
...
+    )
+    run_krew "${krew_root}" list || (
+        echo >&2 "krew list is failing"
+        exit 1
+    )
+    run_krew "${krew_root}" search || (
+        echo >&2 "krew search is failing"
+        exit 1
+    )
+    run_krew "${krew_root}" uninstall get-all || (
```
corneliusweig: :+1: Good idea to test other sub-commands as well.

hack/verify-receipts-upgrade-migration.sh
ahmetb: not intentional.

ahmetb:
Looks like doing `cd` inside a function is actually changing the $CWD of the script (since we define function inside a { } which is not a subshell). So started a subshell in install_krew_0_2_1 function now to not put up with side-effects of `cd`.

PR #259, opened 2019-07-16
hack: add make-binary.sh for local builds

ahmetb:
I've been finding myself doing a lot of:

    OSARCH=darwin/amd64 hack/make-binaries.sh

this should help with that. I'm mostly making this change for myself.

/kind proposal
/assign @corneliusweig


corneliusweig:
/retitle hack: add make-binary.sh for local builds

PR #260, opened 2019-07-16
Add script to generate plugin overview page

### LINKED ISSUE #258: Proposal: Showcase page for existing plugins
Currently, there is no web-compatible way to show a nicely formatted list of plugins in krew-index.

@ahmetb Suggested to have a markdown page with a list of plugins to fill that gap. This page has to be generated by a script. At the bare minimum, this plugins.md overview should show a table of plugin names + shortDescription. It would be nice to also show additional information, such as github stars via https://shields.io.

See also kubernetes-sigs/krew#192
Migrated from https://github.com/kubernetes-sigs/krew-index/issues/179

corneliusweig:
Close #258


hack/generate-plugin-overview.go
```diff
@@ -0,0 +1,106 @@
...
+	_, _ = fmt.Fprintln(out, pageHeader)
+
+	printTableHeader(out)
+	for _, p := range plugins {
+		printTableRowForPlugin(out, &p)
+	}
+}
+
+func printTableHeader(out io.Writer) {
+	printRow(out, "Name", "Description", "GitHub Stars")
```
ahmetb: Can I recommend actually spitting out html tables?
It's often much better to have html tables in .md files as they're more reliable and bring more freedom that probably will be necessary in the future.

ahmetb:
Do you think we can have  this in cmd/[name]/ instead ?

hack/generate-plugin-overview.go
corneliusweig: Sure. However, this also means that we should also translate the markdown used in some of the short-descriptions into html. I think a common choice for that task is blackfriday (used by hugo for example).

corneliusweig:
> Do you think we can have this in cmd/[name]/ instead ?

Yeah why not.

hack/generate-plugin-overview.go
ahmetb: I think this cmd/foo tool can even have its own go.mod. :)

ahmetb:
Let's move this PR forward by moving the file to cmd/gen-plugin-list/main.go

corneliusweig:
> Let's move this PR forward by moving the file to cmd/gen-plugin-list/main.go

Moving the file was not the problem, converting the table to html was. I didn't really find a clean solution here to prevent `blackfriday` from wrapping its output in `<p>`-tags. I ended up replacing those via a regex.

ahmetb:
🤔 lets keep .md, then.

corneliusweig:
Ok, I reverted to the pure markdown version again.

cmd/plugin-overview/plugin-overview.go
```diff
@@ -0,0 +1,109 @@
...
+
+`
+)
+
+var (
+	githubRepoPattern = regexp.MustCompile(`.*github.com/([^/]+/[^/#]+)`)
+)
+
+func main() {
+	pluginsDir := pflag.String("plugins-dir", "", "The directory containing the plugin manifests")
```
ahmetb: do you get much out of pflag vs stdlib pkg/flags here?

cmd/plugin-overview/go.mod
```diff
@@ -0,0 +1,12 @@
...
+go 1.12
+
+require (
+	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
+	github.com/pkg/errors v0.8.1 // indirect
+	github.com/spf13/pflag v1.0.3
+	sigs.k8s.io/krew v0.2.1
+)
+
+replace sigs.k8s.io/krew => ../..
```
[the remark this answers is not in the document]
corneliusweig: Yes, probably. The problem is that I want to use the homepage field, which was added after the last krew release.
I pinned the import to the current master HEAD instead, to solve the issue you raised.

cmd/plugin-overview/plugin-overview.go
corneliusweig: No. This is a leftover from when this tool was not its own module.
 Only the usage output looks a bit weird when using vanilla `flags`, though.

ahmetb: I wouldn't worry about usage as it's not a primary tool.

cmd/generate-plugin-overview/go.mod
```diff
@@ -0,0 +1,9 @@
+module sigs.k8s.io/krew/cmd/generate-plugin-overview
+
+go 1.12
+
+require (
+	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
+	github.com/pkg/errors v0.8.1 // indirect
+	sigs.k8s.io/krew v0.2.2-0.20190724210953-8867a94acd6d
```
ahmetb: funny enough this doesn't build from root

```
$ go build ./cmd/generate-plugin-overview

go: finding sigs.k8s.io/krew/cmd/generate-plugin-overview latest
go: finding sigs.k8s.io/krew/cmd latest
can't load package: package sigs.k8s.io/krew/cmd/generate-plugin-overview: unknown import path "sigs.k8s.io/krew/cmd/generate-plugin-overview": cannot find module providing package sigs.k8s.io/krew/cmd/generate-plugin-overview
```

but if you `./cmd/generate-plugin-overview` then `go build .` it works haha.

cmd/plugin-overview/plugin-overview.go
ahmetb: I see your point now. `glog` hijacks the flags smh.

cmd/generate-plugin-overview/generate-plugin-overview.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"github.com/golang/glog"
+
+	"sigs.k8s.io/krew/pkg/index"
+	"sigs.k8s.io/krew/pkg/index/indexscanner"
+)
+
+const (
+	separator  = " | "
+	pageHeader = `# Plugin overview
```
ahmetb: Let's do something bit more useful as we'll likely link to this
document from Krew's main repo readme.

```
## Available kubectl plugins

The following kubectl plugins are currently available
on [Krew plugin index](https://sigs.k8s.io/krew-index).

To install these kubectl plugins:

1. [Install Krew](link)
2. Run `kubectl krew install PLUGIN_NAME` to install them via Krew.


```

cmd/generate-plugin-overview/generate-plugin-overview.go
```diff
@@ -0,0 +1,109 @@
...
+	} else if homepage == `https://sigs.k8s.io/krew` {
+		repo = "kubernetes-sigs/krew"
+	} else if homepage == `https://kubernetes.github.io/ingress-nginx/kubectl-plugin/` {
+		repo = "kubernetes/ingress-nginx"
+	}
+
+	if repo == "" {
+		return ""
+	}
+	return "![GitHub stars](https://img.shields.io/github/stars/" + repo + ".svg?label=github%20stars&logo=github)"
```
ahmetb: let's just do label=stars

cmd/generate-plugin-overview/generate-plugin-overview.go
```diff
@@ -0,0 +1,109 @@
...
+
+import (
+	"flag"
+	"fmt"
+	"io"
+	"os"
+	"regexp"
+	"strings"
+
+	"github.com/golang/glog"
```
ahmetb: maybe don't import glog (and just use log.Printf+os.Exit) if it helps you avoid glog hijacking flags

ahmetb:
I just tried and I like it!

Do you envision we would require the output file on krew-index repository to be up to date (as a travis check) or should we update it ourselves periodically? Earlier option sounds like it's a burden during onboarding new plugins.

cmd/generate-plugin-overview/generate-plugin-overview.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	out := os.Stdout
+
+	_, _ = fmt.Fprintln(out, pageHeader)
+
+	printTableHeader(out)
+	for _, p := range plugins {
+		printTableRowForPlugin(out, &p)
+	}
```
ahmetb: can we add a footer like 

```
-----

_This page is generated by running
[generate-plugin-overview](http://sigs.k8s.io/krew/cmd/generate-plugin-overview)
tool._
```

so users can self-serve?

cmd/generate-plugin-overview/generate-plugin-overview.go
corneliusweig: Yes, that makes sense.

cmd/generate-plugin-overview/generate-plugin-overview.go
corneliusweig: I tried, but it didn't really work. It's fine either way. As you said, the tool is just a little helper.

corneliusweig:
> Do you envision we would require the output file on krew-index repository to be up to date

I wouldn't be so strict about the overview. Initially, manual sounds right, as there is

a) not so many new plugins
b) we may not keep the page

In the long run, my favorite solution would be to let a hook add additional commits, when the overview has changed. Or at least, a PR for the change is created. But that sounds like more work than we are saving for a long time.

ahmetb:
Thanks for developing this.
/lgtm
/approve

I'm thinking we host this on `krew` main repo `docs/plugins.md` and link from the main README.md, but we can also go with `krew-index` (just that it ships to everyone’s machine is my only concern, but this is a small file anyway).

PR #261, opened 2019-07-16
Make files in `spec.platforms[]` optional

### LINKED ISSUE #135: Proposal: make spec.platforms[].files optional
If no `files` field specified, we may consider assuming:

```
files
- from: '*'
  to: '.'
```

then developers can just point to the entrypoint within the extracted archive through `bin:` field.

This already started to become a common pattern in several plugins.

corneliusweig:
Many plugins use the pattern to copy all files from the archive into the installation directory. Those plugins always have to specify a file copy operation such as
```yaml
files:
- from: "*"
  to: "."
```

This PR makes the files field completely optional and defaults to the above copy specifier. As discussed in #135, the default copy spec may install too many files. The plugin developer docs therefore explicitly show such an antipattern example.

Close #135

### DIFF
--- a/docs/DEVELOPER_GUIDE.md
+++ b/docs/DEVELOPER_GUIDE.md
@@ -187,14 +187,14 @@ wildcard may install more files than desired.
 ```
 
 - To copy all files in the `bin/` directory of the extracted archive to
-  the root of your plugin, use the default:
+  the root of your plugin, leave out the `files:` field, which is equivalent to
 
   ```yaml
-      files:   # unspecified, equivalent to [{from: "*", to: "."}]
+      files: [{from: "*", to: "."}]
   ```
   
-  This would copy out binaries for both platforms to the installation
-  directory on user’s machine, despite only one of them will be used:
+  This copies out binaries for both platforms to the installation directory
+  onto the user’s machine, despite only one of them will be used:
   
   ```text
   .
--- a/docs/DEVELOPER_GUIDE.md
+++ b/docs/DEVELOPER_GUIDE.md
@@ -104,7 +104,8 @@ spec:
     uri: https://github.com/example/foo/releases/v1.0.zip
     # sha256sum of the file downloaded above:
     sha256: "208fde0b9f42ef71f79864b1ce594a70832c47dc5426e10ca73bf02e54d499d0"
-    files:                     # copy the used files out of the zip archive
+    # copy the used files out of the zip archive, defaults to `[{from: "*", to: "."}]`
+    files:
     - from: "/foo-*/unix/*.sh" # path to the files extracted from archive
       to: "."               # '.' refers to the root of plugin install directory
     bin: "./kubectl-foo"  # path to the plugin executable after copying files above
@@ -169,20 +170,12 @@ be installed. You can use the `files` field in the plugin manifest to specify
 which files should be copied into the plugin directory after extracting the
 plugin archive.
 
-The `file:` list specifies the copy operations (like `mv(1) <from> <to>`) to
-the files `from` the archive `to` the installation destination.
+The `files:` list specifies the copy operations (like `mv(1) <from> <to>`) to
+the files `from` the archive `to` the installation destination. If the `files`
+list is unspecified, it defaults to `[{from: "*", to: "."}]`. Mind that a
+wildcard may install more files than desired.
 
-**Example:** Copy all files in the `bin/` directory of the extracted archive to
-the root of your plugin:
-
-```yaml
-    files:
-    - from: bin/*.exe
-      to: .
-```
-
-Given the file operation above, if the extracted plugin archive looked like
-this:
+**Example:** Consider a plugin whose extracted archive looks like
 
 ```text
 .
@@ -193,12 +186,36 @@ this:
     └── kubectl-foo-windows.exe
 ```
 
-The resulting installation directory would up just with:
+- To copy all files in the `bin/` directory of the extracted archive to
+  the root of your plugin, use the default:
 
-```text
-.
-└── krew-foo-windows.exe
-```
+  ```yaml
+      files:   # unspecified, equivalent to [{from: "*", to: "."}]
+  ```
+  
+  This would copy out binaries for both platforms to the installation
+  directory on user’s machine, despite only one of them will be used:
+  
+  ```text
+  .
+  └── krew-foo-windows.exe
+  └── krew-foo-linux
+  ```
+
+- For a more fine-grained control, specify each copy operation:
+
+  ```yaml
+      files:
+      - from: bin/*.exe
+        to: .
+  ```
+
+  This will result in the following files in the installation directory:
+
+  ```text
+  .
+  └── krew-foo-windows.exe
+  ```
 
 #### Specifying plugin executable
 
--- a/pkg/index/validation/validate.go
+++ b/pkg/index/validation/validate.go
@@ -99,19 +99,19 @@ func ValidatePlugin(name string, p index.Plugin) error {
 // validatePlatform checks Platform for structural validity.
 func validatePlatform(p index.Platform) error {
 	if p.URI == "" {
-		return errors.New("URI has to be set")
+		return errors.New("`uri` has to be set")
 	}
 	if p.Sha256 == "" {
-		return errors.New("sha256 sum has to be set")
+		return errors.New("`sha256` sum has to be set")
 	}
 	if !isValidSHA256(p.Sha256) {
-		return errors.Errorf("sha256 value %s is not valid, must match pattern %s", p.Sha256, sha256Pattern)
+		return errors.Errorf("`sha256` value %s is not valid, must match pattern %s", p.Sha256, sha256Pattern)
 	}
 	if p.Bin == "" {
-		return errors.New("bin has to be set")
+		return errors.New("`bin` has to be set")
 	}
 	if err := validateFiles(p.Files); err != nil {
-		return errors.Wrap(err, "files is invalid")
+		return errors.Wrap(err, "`files` is invalid")
 	}
 	if err := validateSelector(p.Selector); err != nil {
 		return errors.Wrap(err, "invalid platform selector")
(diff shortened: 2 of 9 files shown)


docs/DEVELOPER_GUIDE.md
```diff
@@ -193,12 +186,35 @@ this:
...
-
-```text
-.
-└── krew-foo-windows.exe
-```
+- To copy all files in the `bin/` directory of the extracted archive to
+  the root of your plugin, use the default:
+
+  ```yaml
+      files: []  # same as [{from: "*", to: "."}]
```
ahmetb: Hmm. I think empty list should _not_ be treated as unspecified (`nil`) field.

Empty means do nothing, which a valid operation, but it shouldn't be allowed since it does not make sense.

docs/DEVELOPER_GUIDE.md
```diff
@@ -193,12 +186,35 @@ this:
...
+  ```
+  
+  The resulting installation directory would eventually contain:
```
ahmetb: This would copy out binaries for both platforms to the installation directory on user’s machine, despite only one of them will be used.

pkg/index/validate.go
```diff
@@ -90,8 +90,13 @@ func (p Platform) Validate() error {
 	if p.Bin == "" {
 		return errors.New("bin has to be set")
 	}
-	if len(p.Files) == 0 {
-		return errors.New("can't have a plugin without specifying file operations")
+	for _, fo := range p.Files {
```
ahmetb: Let's extract this to a new method that we can unit-test, just like ValidatePlatform above.

ahmetb: Let’s add a check that verifies `nil` is ok for p.Files (you should make this `Files` field a pointer like *[]FileOperaton first I think to distinguish between `nil` and empty `[]`) AND validate that empty is not allowed and should not be defaulted.

pkg/installation/util.go
```diff
@@ -76,7 +76,12 @@ func getDownloadTarget(index index.Plugin) (version, uri string, fos []index.Fil
 	version, uri = getPluginVersion(p)
 	glog.V(4).Infof("Matching plugin version is %s", version)
 
-	return version, uri, p.Files, p.Bin, nil
+	files := p.Files
+	if len(files) == 0 {
+		files = []index.FileOperation{{From: "*", To: "."}}
```
ahmetb: worth adding a glog.V(4) here saying `file operation not specified, assuming %v`

pkg/installation/util_test.go
```diff
@@ -87,7 +87,33 @@ func Test_getDownloadTarget(t *testing.T) {
 			},
 			wantVersion: "deadbeef",
 			wantURI:     "https://uri.git",
-			wantFos:     nil,
+			wantFos:     []index.FileOperation{{From: "foo", To: "bar"}},
+			wantBin:     "kubectl-foo",
+			wantErr:     false,
+		}, {
+			name: "Find Matching Platform with default files",
```
ahmetb: `...with FileOperations unspecified` or `...defaulted`

ahmetb:
Overall looks good. I think we just need to distinguish clearly between unspecified (null) and empty ([]) while allowing one and failing on another.

docs/DEVELOPER_GUIDE.md
corneliusweig: Isn't the differentiation between empty and nil a language subtlety? From a user's perspective both cases mean "no rules given". And here I think we are complicating things if we make that distinction.

ahmetb: I'll move this to main thread.

ahmetb:
> Isn't the differentiation between empty and nil a programming language subtlety?

Sadly, no. This is independent of programming language.
In one case, you're specifying "I have no elements" in another, it's unspecified altogether.

More obvious examples of this can be seen in Kubernetes API design. For example, in Kubernetes NetworkPolicy API, specifying `ingress: []` as an empty array explicitly causes a policy to be created allowing _no_ traffic. However, not specifying this field doesn't have that impact.

Let's play safe in this case and (1) distinguish nil and default on that (2) validate error on empty array. Here's another reason to play safe: We can start with defaulting on just `nil` today and failijng on `[]`, if it works out we can expand the definition. However, if we default on both `nil` and `[]` (and realize it's wrong API design), we can't easily change the behavior without breaking the API.

I prefer to err on being more restrictive and being more lenient _only_ if there's a neet.

corneliusweig:
Ok, that forward compatibility argument is convincing. I'll address the nil vs empty issue, but I'll need a few days due to other obligations.

ahmetb:
Haha sorry I caused merge conflicts everywhere. Now that we have proper validation on `Platform`, and testutil.NewPlugin() utilities, this PR should be easier to code.

corneliusweig:
Rebase is finally done.

I added a new validation case which forbids empty file operations now. The difference between empty and unspecified is quite subtle though (this comes from our yaml library):

```yaml
      ...
      bin: foo-amd64
      files:  # present, but no values
```
is equivalent to
```yaml
      ...
      bin: foo-amd64
#    files:  <- not present
```

Whereas this triggers the error case:
```yaml
      ...
      bin: foo-amd64
      files: []
```

ahmetb:
I think it's ok to treat unspecified and `files: ` (with no values) the same.
On the other hand, `files: []` explicitly says I don't want any values in there.

docs/DEVELOPER_GUIDE.md
```diff
@@ -193,12 +186,36 @@ this:
...
 ```
 
-The resulting installation directory would up just with:
-
-```text
-.
-└── krew-foo-windows.exe
-```
+- To copy all files in the `bin/` directory of the extracted archive to
+  the root of your plugin, use the default:
```
ahmetb: hmm, to use the default, maybe they shouldn't need to specify `files:`?

pkg/index/validation/validate.go
```diff
@@ -110,15 +110,33 @@ func validatePlatform(p index.Platform) error {
...
+func validateFiles(fops []index.FileOperation) error {
+	if fops == nil {
+		return nil
+	}
+	if len(fops) == 0 {
+		return errors.New("files has to be unspecified or non-empty")
+	}
+	for _, op := range fops {
+		if op.From == "" {
+			return errors.New("from field has to be set")
```
ahmetb: can we say \"from\" and \"to\" (below) to distinguish field names from the rest of the error message. especially below it says `to ... to be set` so it's a bit confusing.

pkg/index/validation/validate.go
```diff
@@ -110,15 +110,33 @@ func validatePlatform(p index.Platform) error {
...
+		}
```
ahmetb: nit: can do else-if here to save 1 line

pkg/index/validation/validate_test.go
```diff
@@ -190,8 +190,8 @@ func TestValidatePlatform(t *testing.T) {
 			wantErr:  true,
 		},
 		{
-			name:     "no file operations",
-			platform: testutil.NewPlatform().WithFiles(nil).V(),
+			name:     "empty file operations",
```
ahmetb: can you also add WithFiles(nil) case (`wantErr:false`)?

pkg/installation/util.go
```diff
@@ -25,23 +25,29 @@ import (
...
 	if !ok {
 		return "", "", "", nil, p.Bin, errors.New("no matching platform found")
 	}
-	version = index.Spec.Version
+	version = plugin.Spec.Version
 	uri = p.URI
 	sha256sum = p.Sha256
+
+	files := p.Files
+	if len(files) == 0 {
```
ahmetb: I'm removing this method (`getDownloadTarget`) soon as part of #293.

I think we should consider a method like `func applyDefaults(*index.Platform)` to achieve defaulting on the FileOperations. Are you open to such a method, that's easily testable etc?

corneliusweig: In principle, I like the idea. However, I didn't find a good spot for this. The best place is close to validation, but there we pass in a copy of the plugin object which loses any mutations. And I didn't want to put this into `ReadPluginFile`.

So I ended up, putting it in `moveToInstallDir`. Not ideal, though..

pkg/index/validation/validate_test.go
corneliusweig: Those test cases are covered by `Test_validateFiles`. I only included this error case in this test, to check if files are validated at all. Do you think this is not enough?

pkg/index/validation/validate.go
corneliusweig: I enclosed field names in backticks. To be consistent, I did the same for other error messages.

docs/DEVELOPER_GUIDE.md
corneliusweig: Well, my thinking was to keep the two examples as similar as possible. But I agree that this looks odd. I changed the wording.

pkg/installation/util.go
ahmetb: I think that's why we should call `applyDefaults(&plugin)` right before we do moveToInstallDir. Then we can do  `moveToInstallDir(..., plugin.Files)`.

Don't move inside `moveToInstallDir` as the name indicates, it's not expected to do defaulting.

pkg/installation/move.go
```diff
@@ -165,6 +165,11 @@ func moveToInstallDir(srcDir, installDir string, fos []index.FileOperation) erro
 	}
 	defer os.RemoveAll(tmp)
 
+	if fos == nil {
+		fos = []index.FileOperation{{From: "*", To: "."}}
```
ahmetb: this makes the method less testable :|

semantically, if fos==nil you'd expect this method to do nothing, based on its current name + godoc. :)

pkg/installation/util.go
corneliusweig: Alright, I moved the logic one level up into `install`. Also, I added a unit test for `applyDefaults`.

pkg/installation/install_test.go
```diff
@@ -257,3 +259,41 @@ func Test_downloadAndExtract_fileOverride(t *testing.T) {
...
+			expected: testutil.NewPlatform().WithFiles([]index.FileOperation{}).V(),
+		},
+		{
+			name:     "with unspecified files",
+			platform: testutil.NewPlatform().WithFiles(nil).V(),
+			expected: testutil.NewPlatform().WithFiles([]index.FileOperation{{From: "*", To: "."}}).V(),
+		},
+		{
+			name:     "no defaults for other fields",
+			platform: testutil.NewPlatform().WithBin("").WithOS("").WithSelector(nil).WithSHA256("").WithURI("").V(),
```
ahmetb: oh boy this test is gonna be a problem in the future.
because we will keep adding defaults to testutil.NewPlatform(), and we will keep adding new WithXxx.
let's keep it until it breaks.

PR #262, opened 2019-07-16
hack: verify builds are installable

ahmetb:
This script tests if a build is installable (as documented in the readme)
from its tarball+yaml file.

I felt like adding this as extra confidence on all the refactoring we're doing,
and I know we have integration tests that cover this. Since installation is
almost like our API, I thought it would be good to test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ script:
 - hack/run-lint.sh
 - go test -v -short -coverprofile=coverage.txt -covermode=atomic ./...
 - hack/make-all.sh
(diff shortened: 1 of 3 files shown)


corneliusweig:
/approve
/lgtm
I hope that the number of shell scripts does not continue to grow at the same pace :wink:

ahmetb:
> I hope that the number of shell scripts does not continue to grow at the same pace 😉

HAHA I know. But as I was working on some new stuff I wasn't sure if I can install krew locally. Now I know the builds can install themselves. :P

PR #273, opened 2019-07-17
semver: add wrapper for SemVer functionality

ahmetb:
This package introduces a wrapper for `k8s.io/apimachinery/pkg/util/version`
package for its parts that we rely on for our handling of semver values.

go.{mod,sum} will likely conflict soon as there are other in-flight PRs.

Also refactored the validation code to verify 'version' value can be parsed.

/assign @corneliusweig

### DIFF
--- a/pkg/installation/semver/version.go
+++ b/pkg/installation/semver/version.go
@@ -1,0 +1,59 @@
+// Copyright 2019 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package semver is a wrapper for handling of semantic version
+// (https://semver.org) values.
+package semver
+
+import (
+	"strings"
+
+	"github.com/pkg/errors"
+	k8sver "k8s.io/apimachinery/pkg/util/version"
+)
+
+// Version is in-memory representation of a semantic version
+// (https://semver.org) value.
+type Version k8sver.Version
+
+// String returns string representation of a semantic version value with a
+// leading 'v' character.
+func (v Version) String() string {
+	vv := k8sver.Version(v)
+	s := (&vv).String()
+	if !strings.HasPrefix(s, "v") {
+		s = "v" + s
+	}
+	return s
+}
+
+// Parse parses a semantic version value with a leading 'v' character.
+func Parse(s string) (Version, error) {
+	var vv Version
+	if !strings.HasPrefix(s, "v") {
+		return vv, errors.Errorf("version string %q not starting with 'v'", s)
+	}
+	v, err := k8sver.ParseSemantic(s)
+	if err != nil {
+		return vv, err
+	}
+	return Version(*v), nil
+}
+
+// Less checks if a is strictly less than b (a<b).
+func Less(a, b Version) bool {
+	aa := k8sver.Version(a)
+	bb := k8sver.Version(b)
+	return aa.LessThan(&bb)
+}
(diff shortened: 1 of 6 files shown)


ahmetb:
Tests expected to fail because https://github.com/kubernetes-sigs/krew-index/pull/181 currently fails the `krew search` command.

pkg/installation/semver/version.go
```diff
@@ -0,0 +1,58 @@
...
+
+// Version describes a semantic version value
+type Version k8sver.Version
+
+// String returns string representation of a semantic version
+// (https://semver.org) value with leading 'v' character.
+func (v Version) String() string {
+	vv := k8sver.Version(v)
+	s := (&vv).String()
+	if !strings.HasPrefix(s, "v") {
```
corneliusweig: This check is obsolete, as it will always evaluate to true, so that could be a one-liner.

pkg/installation/semver/version.go
```diff
@@ -0,0 +1,58 @@
...
+		s = "v" + s
+	}
+	return s
+}
+
+// Parse parses a semantic version (https://semver.org) value with leading 'v'
+// character.
+func Parse(s string) (Version, error) {
+	var vv Version
+	if !strings.HasPrefix(s, "v") {
```
corneliusweig: We could be lenient here and just pass through the semver, as k8sver understands an optional `v` prefix.

pkg/installation/semver/version_test.go
```diff
@@ -0,0 +1,107 @@
...
+func TestParse(t *testing.T) {
+	cases := []struct {
+		name    string
+		in      string
+		want    string
+		wantErr bool
+	}{
+		{"empty", "", "", true},
+		{"zero is valid", "v0.0.0", "v0.0.0", false},
+		{"no v prefix", "v1.0.0", "v1.0.0", false},
```
corneliusweig: Test name is misleading:
```go
{"no v prefix", "v1.0.0", "v1.0.0", false},
```

pkg/installation/semver/version_test.go
```diff
@@ -0,0 +1,107 @@
...
+		{"major.minor.patch", "v1.2.3", "v1.2.3", false},
+		{"major.minor.patch-suffix", "v1.2.3-beta.2+foo.bar", "v1.2.3-beta.2+foo.bar", false},
+		{"leading zero in major", "v01.2.3", "", true},
+		{"leading zero in minor", "v1.02.3", "", true},
+		{"leading zero in patch", "v1.2.03", "", true},
+		{"negative value in major", "v-1.2.3", "", true},
+		{"negative value in minor", "v1.-2.3", "", true},
+		{"negative value in patch", "v1.2.-3", "", true},
+		{"empty pre-release identifier", "v1.0.1-", "", true},
+		{"empty meta identifier", "v1.0.1+", "", true},
```
corneliusweig: Maybe also worth testing if `[:alpha:]` chars lead to errors.

pkg/installation/semver/version.go
ahmetb: I think we should be less lenient first, then maybe more lenient.
Right now all plugins have v* prefix, so let's keep it?

pkg/installation/semver/version.go
ahmetb: I have this because this pkg is meant to be a wrapper.
If we don't like the underlying k8sver package, we can change it without changing our code. Not all semver parsers are enforcing v*, but I want to.

pkg/installation/semver/version_test.go
ahmetb: can you give a specific example?

pkg/installation/semver/version.go
corneliusweig: IMHO the k8sver handles errors just fine, so why add additional checks around that call? I don't think we will switch to another SemVer implementation which would require the v prefix (_that_ would justify the additional check).
But it's a minor thing, I'm fine either way. Your call.

pkg/installation/semver/version_test.go
corneliusweig: Something like
```go
{"major with alpha chars", "va.0.1", "", true},
{"minor with alpha chars", "v1.a.1", "", true},
{"patch with alpha chars", "v1.0.a", "", true},
```

ahmetb: that feels a bit too obvious but I'll add tests for `v0a.0.0`, `v0.0a.0` and `v0.0.0a`

Issue #270, opened 2019-07-17
minor: duplicate test object cleanup


ahmetb:
Looks like both pkg/receipt/receipt_test.go and pkg/info/info_test.go have this in-memory plugin + pluginName definition.

It would be good to dedupe this in a testdata/sampleplugin_test.go (exported identifier) and import here, I think.

/kind cleanup

corneliusweig:
Note: there are two todo items labeled with the issue number that should be addressed in this PR.

PR #276, opened 2019-07-17
Update plugin listings to use install receipts

ahmetb:
- Changed the way we determine the list of "installed plugins" by looking at
  install receipts directory.
- Added a receipt.Load() method for reading saved install receipts.

/assign @corneliusweig

### DIFF
--- a/pkg/installation/util.go
+++ b/pkg/installation/util.go
@@ -15,7 +15,6 @@
 package installation
 
 import (
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
@@ -23,8 +22,10 @@ import (
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
 
+	"sigs.k8s.io/krew/pkg/constants"
 	"sigs.k8s.io/krew/pkg/index"
 	"sigs.k8s.io/krew/pkg/pathutil"
+	"sigs.k8s.io/krew/pkg/receipt"
 )
 
 func findInstalledPluginVersion(installPath, binDir, pluginName string) (name string, installed bool, err error) {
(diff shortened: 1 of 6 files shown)


pkg/installation/util.go
```diff
@@ -79,27 +80,22 @@ func getDownloadTarget(index index.Plugin) (version, uri string, fos []index.Fil
 	return version, uri, p.Files, p.Bin, nil
 }
 
-// ListInstalledPlugins returns a list of all name:version for all plugins.
-func ListInstalledPlugins(installDir, binDir string) (map[string]string, error) {
-	installed := make(map[string]string)
-	plugins, err := ioutil.ReadDir(installDir)
+// ListInstalledPlugins returns a list of all install plugins in a
+// name:version format based on the install receipts at the specified dir.
+func ListInstalledPlugins(receiptsDir string) (map[string]string, error) {
```
corneliusweig: Nice, that just got a lot simpler :+1: 
Would be a good time to also add some tests for this.

pkg/receipt/receipt.go
```diff
@@ -34,3 +35,13 @@ func Store(plugin index.Plugin, dest string) error {
 	err = ioutil.WriteFile(dest, yamlBytes, 0644)
 	return errors.Wrapf(err, "write plugin receipt %q", dest)
 }
+
+// Load reads the plugin receipt at the specified destination.
+// If not found, it returns os.IsNotExist error.
+func Load(path string) (*index.Plugin, error) {
```
corneliusweig: This method returns a pointer instead of an object, which is inconsistent with how krew passes around plugin manifests internally. Although I favor the pointer approach, I think that consistency is worth more.
Being consistent also means, that this function can be a one-liner.

pkg/receipt/receipt_test.go
```diff
@@ -68,3 +70,20 @@ func TestStore(t *testing.T) {
 		t.Fatal(diff)
 	}
 }
+
+func TestLoad(t *testing.T) {
+	p, err := Load(filepath.Join("..", "..", "integration_test", "testdata", "foo.yaml"))
```
corneliusweig: This doesn't feel right. Depending on test data in another package looks like headache in the future.
What about writing the plugin manifest to disk and read with `Load`? For that it could be helpful to have a `testutil.ValidPluginManifest` global value which can be used by other tests as well.

ahmetb: > What about writing the plugin manifest to disk and read with `Load`?

this is a good idea.

> For that it could be helpful to have a `testutil.ValidPluginManifest` global value

this is mentioned in #270

pkg/receipt/receipt.go
ahmetb: I can't recall why I made it a pointer. I probably was thinking you making things *index.Plugin in another PR recently. There's no particular reason not to do this.

pkg/receipt/receipt_test.go
ahmetb: Hmm let's keep it as is, I'll add a TODO here. I have two concerns:

1. I don't want to do #270 in this PR.
2. Ideally I don't want Load()’s tests to fail because of a fault in Store(). But I think we can live with it.

pkg/installation/util.go
ahmetb: Since adding tests to this requires me to Store() some manifests by serializing the in-memory Plugin objects (#270) to a tempdir, I will need #270 to happen first.

To unblock here, I'm adding a TODO to the code explaining this situation.

corneliusweig:
Looks great!
/lgtm
/approve

Issue #277, opened 2019-07-18
Refactor all commands to determine they use plugin receipts


ahmetb:
I don't think we fully made a migration to fully rely on plugin receipts.

Last PR (#195) introduced "saving" plugin receipts during install/upgrade and cleaning them up during uninstall/remove. I'm not sure if other commands were covered.

This is a tracking item for making a pass on the code to make sure all commands (like info, list, upgrade, install, uninstall/remove) are using receipts. We should verify how these commands work under the covers by also reading the verbose output.

/priority backlog
/priority important-soon

ahmetb:
@corneliusweig I think this is now completed?

corneliusweig:
Yes, I just had another look and it looks all good. I even checked the log output.
There were a number of PRs regarding this migration: #195, #276, #278
/close

PR #278, opened 2019-07-18
Refactor upgrade machinery to use semver versions

### LINKED ISSUE #166: Proposal: Use semver to determine plugin versions and upgrades
As of writing the `version` field in plugin spec is purely for documentation.

We should consider enforcing and using this field (as a [semver](https://semver.org) value) so that we could rely on this value to determine if a plugin has a "newer" version or not.

Right now we rely on sha256 sum of the installed `spec.platforms` value (and compare it to the one on the local copy of the manifest) to determine if the version of the plugin we installed is the same or not (no notion of newer/older).

This would mean:

- Potentially **breaking** change to enforce all plugins to use `version` field. (Can be done in a non-breaking way if we keep the existing code path intact.)
- Add tests to `krew-index` repo's CI and check every plugin manifest update increments the `version` value.
- Refactor the plugin "upgrade" path:
  - compare the currently installed version (this can be encoded to the install directory name and/or to an "install receipt" #163) WITH the version field of the plugin in the local copy of the index,
  - determine if an upgrade can be made.

ahmetb:
This change starts using semver version of plugin manifests to:
- determine if upgrade is needed
- determine if a plugin is installed

by looking at receipts installation directory. It doesn't help with other commands just yet.

This fixes #166, and helps with #277 partially.

Other commands are probably not yet using receipts entirely.

### DIFF
--- a/pkg/download/verifier.go
+++ b/pkg/download/verifier.go
@@ -53,7 +53,7 @@ func (v sha256Verifier) Verify() error {
 	if bytes.Equal(v.wantedHash, v.Sum(nil)) {
 		return nil
 	}
-	return errors.Errorf("checksum does not match, wantReader: %x, got %x", v.wantedHash, v.Sum(nil))
+	return errors.Errorf("checksum does not match, want: %x, got %x", v.wantedHash, v.Sum(nil))
 }
 
 var _ Verifier = trueVerifier{}
--- a/pkg/installation/install.go
+++ b/pkg/installation/install.go
@@ -41,7 +41,7 @@ const (
 	krewPluginName = "krew"
 )
 
-func downloadAndMove(version, uri string, fos []index.FileOperation, downloadPath, installPath, forceDownloadFile string) (dst string, err error) {
+func downloadAndMove(version, sha256sum, uri string, fos []index.FileOperation, downloadPath, installPath, forceDownloadFile string) (dst string, err error) {
 	glog.V(3).Infof("Creating download dir %q", downloadPath)
 	if err = os.MkdirAll(downloadPath, 0755); err != nil {
 		return "", errors.Wrapf(err, "could not create download path %q", downloadPath)
@@ -53,7 +53,7 @@ func downloadAndMove(version, uri string, fos []index.FileOperation, downloadPat
 		fetcher = download.NewFileFetcher(forceDownloadFile)
 	}
 
-	verifier := download.NewSha256Verifier(version)
+	verifier := download.NewSha256Verifier(sha256sum)
 	if err := download.NewDownloader(verifier, fetcher).Get(uri, downloadPath); err != nil {
 		return "", errors.Wrap(err, "failed to download and verify file")
 	}
@@ -64,16 +64,15 @@ func downloadAndMove(version, uri string, fos []index.FileOperation, downloadPat
 // to not get the plugin dir in a bad state if it fails during the process.
 func Install(p environment.Paths, plugin index.Plugin, forceDownloadFile string) error {
 	glog.V(2).Infof("Looking for installed versions")
-	_, ok, err := findInstalledPluginVersion(p.InstallPath(), p.BinPath(), plugin.Name)
-	if err != nil {
-		return err
-	}
-	if ok {
+	_, err := receipt.Load(p.PluginInstallReceiptPath(plugin.Name))
+	if err == nil {
 		return ErrIsAlreadyInstalled
+	} else if !os.IsNotExist(err) {
+		return errors.Wrap(err, "failed to look up plugin receipt")
 	}
 
 	glog.V(1).Infof("Finding download target for plugin %s", plugin.Name)
-	version, uri, fos, bin, err := getDownloadTarget(plugin)
+	version, sha256, uri, fos, bin, err := getDownloadTarget(plugin)
 	if err != nil {
 		return err
 	}
@@ -81,7 +80,7 @@ func Install(p environment.Paths, plugin index.Plugin, forceDownloadFile string)
 	// The actual install should be the last action so that a failure during receipt
 	// saving does not result in an installed plugin without receipt.
 	glog.V(3).Infof("Install plugin %s", plugin.Name)
-	if err := install(plugin.Name, version, uri, bin, p, fos, forceDownloadFile); err != nil {
+	if err := install(plugin.Name, version, sha256, uri, bin, p, fos, forceDownloadFile); err != nil {
 		return errors.Wrap(err, "install failed")
 	}
 	glog.V(3).Infof("Storing install receipt for plugin %s", plugin.Name)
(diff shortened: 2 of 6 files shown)


pkg/download/verifier.go
```diff
@@ -53,7 +53,7 @@ func (v sha256Verifier) Verify() error {
 	if bytes.Equal(v.wantedHash, v.Sum(nil)) {
 		return nil
 	}
-	return errors.Errorf("checksum does not match, wantReader: %x, got %x", v.wantedHash, v.Sum(nil))
+	return errors.Errorf("checksum does not match, want: %x, got %x", v.wantedHash, v.Sum(nil))
```
corneliusweig: :question: :+1:

pkg/installation/install.go
```diff
@@ -41,7 +41,7 @@ const (
 	krewPluginName = "krew"
 )
 
-func downloadAndMove(version, uri string, fos []index.FileOperation, downloadPath, installPath, forceDownloadFile string) (dst string, err error) {
+func downloadAndMove(version, sha256sum, uri string, fos []index.FileOperation, downloadPath, installPath, forceDownloadFile string) (dst string, err error) {
```
corneliusweig: 7 arguments? I hope this is a very very private function. Can you add a todo so we get reminded that this needs to be refactored?

pkg/installation/install.go
```diff
@@ -64,33 +64,32 @@ func downloadAndMove(version, uri string, fos []index.FileOperation, downloadPat
 // to not get the plugin dir in a bad state if it fails during the process.
 func Install(p environment.Paths, plugin index.Plugin, forceDownloadFile string) error {
 	glog.V(2).Infof("Looking for installed versions")
-	_, ok, err := findInstalledPluginVersion(p.InstallPath(), p.BinPath(), plugin.Name)
-	if err != nil {
-		return err
-	}
-	if ok {
+	_, err := receipt.Load(p.PluginInstallReceiptPath(plugin.Name))
```
corneliusweig: As you are discarding the return value, could you also do a simpler `Lstat`?

pkg/installation/util.go
```diff
@@ -62,22 +35,23 @@ func pluginVersionFromPath(installPath, pluginPath string) (string, error) {
 	return elems[1], nil
 }
 
-func getPluginVersion(p index.Platform) (version, uri string) {
-	return strings.ToLower(p.Sha256), p.URI
-}
-
-func getDownloadTarget(index index.Plugin) (version, uri string, fos []index.FileOperation, bin string, err error) {
+func getDownloadTarget(index index.Plugin) (version, sha256sum, uri string, fos []index.FileOperation, bin string, err error) {
+	// TODO(ahmetb): We have many return values from this method, indicating
```
corneliusweig: :+1:

pkg/installation/install.go
```diff
@@ -113,18 +112,22 @@ func install(plugin, version, uri, bin string, p environment.Paths, fos []index.
...
 	}
 	glog.V(3).Infof("Finding installed version to delete")
-	version, installed, err := findInstalledPluginVersion(p.InstallPath(), p.BinPath(), name)
-	if err != nil {
-		return errors.Wrap(err, "can't uninstall plugin")
-	}
-	if !installed {
-		return ErrIsNotInstalled
+
+	if _, err := receipt.Load(p.PluginInstallReceiptPath(name)); err != nil {
```
corneliusweig: Again, a simpler `Lstat` would do.

pkg/installation/install.go
```diff
@@ -137,10 +140,10 @@ func Uninstall(p environment.Paths, name string) error {
 	if err := os.RemoveAll(pluginInstallPath); err != nil {
 		return errors.Wrapf(err, "could not remove plugin directory %q", pluginInstallPath)
 	}
-	PluginInstallReceiptPath := p.PluginInstallReceiptPath(name)
```
corneliusweig: Woops.. I should have caught this in the review.

pkg/installation/util.go
```diff
@@ -62,22 +35,23 @@ func pluginVersionFromPath(installPath, pluginPath string) (string, error) {
...
-		return "", "", nil, p.Bin, errors.New("no matching platform found")
+		return "", "", "", nil, p.Bin, errors.New("no matching platform found")
 	}
-	version, uri = getPluginVersion(p)
-	glog.V(4).Infof("Matching plugin version is %s", version)
-
-	return version, uri, p.Files, p.Bin, nil
+	version = index.Spec.Version
+	uri = p.URI
+	sha256sum = p.Sha256
```
corneliusweig: The original version had a `ToLower(p.Sha256)`. Are you sure it's not needed?

pkg/installation/install.go
```diff
@@ -113,18 +112,22 @@ func install(plugin, version, uri, bin string, p environment.Paths, fos []index.
 // Uninstall will uninstall a plugin.
 func Uninstall(p environment.Paths, name string) error {
 	if name == krewPluginName {
-		return errors.Errorf("removing krew is not allowed through krew. Please run:\n\t rm -r %s", p.BasePath())
+		glog.Errorf("Removing krew through krew is not supported.")
+		if !isWindows() { // assume POSIX-like
+			glog.Errorf("If you’d like to uninstall krew altogether, run:\n\trm -rf -- %q", p.BasePath())
```
corneliusweig: Very nice of you :)

pkg/installation/util.go
ahmetb: The fact that we're not sure points to a lack of tests issue.

I recall arguing that we should make casing of sha256sum mandatory and part of the validation. I think it's time to do that now.

pkg/installation/install.go
ahmetb: I think we need a `receipt.Exists`. Let's add that.

pkg/installation/install.go
ahmetb: The TODO is already at `getDownloadTarget` (see that method's diff). Once that's fixed we'd be just passing the returned struct here to simplify.

pkg/installation/install.go
ahmetb: I'm inclined to keep this as is. If receipt cannot be parsed during uninstallation, that points to a problem and we should probably stop.

In the future we might introduce a field like `status.originIndex` (or like `status.installPath`) to determine the installation path. So we'll actually need to use the receipt, soon.

pkg/installation/install.go
ahmetb: On a second thought, I'm thinking if an empty file named plugin.yaml exists instead of the actual receipt, despite being harmless today, it might complicate things.

This doesn't have a lot of cost. Can we keep?

pkg/installation/util.go
ahmetb: I've created #281 to address this.

pkg/installation/install.go
corneliusweig: Yeah, that makes sense. Then let's keep it as is.

PR #280, opened 2019-07-18
pkg/receipt: move under pkg/installation

### LINKED ISSUE #271: minor: move pkg/receipt
It seems like pkg/receipt's definition is very narrow (currently only one
method Store(), but I'll be adding Load() shortly).

It feels like we should do one of these:

- make it a subpackage of pkg/installation, named `installreceipt`
- move it under pkg/installation directly and rename exported methods to
  StoreInstallReceipt(), LoadInstallReceipt() etc.

/kind cleanup

ahmetb:
Moving to somewhere more logically relevant.
Fixes #271.

/hold
since other in-flight PRs (or this PR) like #278 will require rebase (depending
on which one gets merged first).


ahmetb:
ready to review

PR #281, opened 2019-07-20
pkg/index: validate sha256 values

ahmetb:
This makes sure the code only deals with valid lowercase sha256 values
(i.e. all lowercase) while doing comparisons etc.

### DIFF
--- a/pkg/index/validate.go
+++ b/pkg/index/validate.go
@@ -24,8 +24,13 @@ import (
 	"sigs.k8s.io/krew/pkg/installation/semver"
 )
 
+const (
+	sha256Pattern = `^[a-f0-9]{64}$`
+)
+
 var (
 	safePluginRegexp = regexp.MustCompile(`^[\w-]+$`)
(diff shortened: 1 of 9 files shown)


ahmetb:
This is expected to have merge-conflicts with #278 as it updates some test files deleted in that PR.

pkg/index/validate.go
```diff
@@ -93,6 +96,9 @@ func (p Platform) Validate() error {
 	if p.Sha256 == "" {
 		return errors.New("sha256 sum has to be set")
 	}
+	if !isValidSHA256(p.Sha256) {
+		return errors.Errorf("sha256 value %s is not a valid sha256sum string", p.Sha256)
```
corneliusweig: Could you mention the requirement, that the string is expected to be lower-case?

ahmetb: good point.

ahmetb: I decided to include the pattern in the error message.

PR #283, opened 2019-07-21
pkg/testutil: add in-memory Plugin builder

### LINKED ISSUE #270: minor: duplicate test object cleanup

ahmetb:
Introducing builder-pattern methods to create in-memory Plugin/Platform objects
that has good defaults, but can be overwritten for test purposes. Readability
of the test files (e.g. validation/*_test.go) now speak for themselves.

The testutil.NewPlugin() returns a _valid_ index.Plugin instance. Any tests
expected to fail can modify its values with the builder pattern. (ditto for
testutil.NewPlatform()).

- had to move validation code to pkg/index/validation due to a cyclic
  import situation that occurred. This should be fine.
- renamed some test cases for ease of readability
- I've thought about adding code validation for **_test.go to not to have any
  `index.Plugin{` or `index.Platform{` strings. But this might be an overkill.

Fixes #270. Beware of merge conflicts that will raise due to in-flight PRs.


pkg/testutil/plugin.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func (p *R) WithOSArch(os, arch string) *R {
+	p.v.Selector.MatchLabels = map[string]string{"os": os, "arch": arch}
+	return p
+}
+
+func (p *R) WithFiles(v []index.FileOperation) *R { p.v.Files = v; return p }
+func (p *R) WithBin(v string) *R                  { p.v.Bin = v; return p }
+func (p *R) WithURL(v string) *R                  { p.v.URI = v; return p }
```
corneliusweig:
```suggestion
func (p *R) WithURI(v string) *R                  { p.v.URI = v; return p }
```
for consistency's sake

pkg/info/info_test.go
```diff
@@ -117,6 +92,7 @@ func TestLoadManifestFromReceiptOrIndex(t *testing.T) {
 			} else {
 				if err != nil {
 					t.Error(err)
+					return
```
corneliusweig: maybe use `t.Fatal(err)` instead?

pkg/testutil/plugin.go
```diff
@@ -0,0 +1,95 @@
...
+	defaultMeta             = metav1.TypeMeta{
+		APIVersion: constants.CurrentAPIVersion,
+		Kind:       constants.PluginKind,
+	}
+
+	defaultPlatformURI    = "http://example.com/"
+	defaultPlatformSHA256 = "deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef"
+	defaultPlatformOSArch = map[string]string{
+		"os":   "linux",
+		"arch": "amd64"}
```
corneliusweig: super-nit: that closing brace looks a bit awkward.

ahmetb: this syntax is fine :D

pkg/info/info_test.go
ahmetb: since this is a for-loop with test cases, it's not a great idea. if we Fatal, other cases won't run.

ahmetb:
> There are some left-over Plugin usages in `util_test.go` and `platform_test.go` which could be cleaned up. Did you leave those out intentionally?

Yeah it seems like I forgot.
Perhaps we can even explore adding a check for preventing index.Plugin{} or index.Platform{} initializations and have all tests use this approach.

ahmetb:
> platform_test.go

So looks like platform.go can't use this utility package, because it uses a cyclic import when used in test case `Test_matchPlatformToSystemEnvs()`. 

I'm planning to move this method to `pkg/installation` as pkg/index.PluginSpec isn't the best place to have this method. I'll do this post this PR.

PR #285, opened 2019-07-21
test: remove -v from tests

ahmetb:
-v is only useful if you want to see all the test cases that are being run.
I don't think this is useful locally or in the build system. I've been having
trouble scrolling a lot and trying to find the FAIL: line.

Any failed test will still print its t.Log or stdout/stderr printing along with
the test method name, so this should be fine.

/assign @corneliusweig
/kind cleanup


corneliusweig:
Yeah, makes sense.
/lgtm
/approve

PR #286, opened 2019-07-22
Move platform detection to pkg/installation

### LINKED ISSUE #284: move matchPlatformToSystemEnvs to pkg/installation
Currently matchPlatformToSystemEnvs() is on the PluginSpec type itself.
We should ideally reduce the methods with non-trivial logic around Plugin,
PluginSpec and Platform types.

This also helps us use the new testutil.NewPlatform() etc while testing as
currently it can't be used due to cyclic imports (which is a sign of code smell
here).

/kind cleanup

ahmetb:
/kind cleanup

- moved Platform matching methods to pkg/installation, where:
  - they make more sense
  - they can be tested without worrying about cyclic imports due to pkg/testutil.
- rewrote some tests in platform_test.go to be shorter.

- added a check to verify-code-patterns that uses a heuristic to detect inline
  initializations of index.(Plugin|Platform) structs.

Fixes #284.


pkg/index/validation/validate_test.go
```diff
@@ -150,6 +150,12 @@ func TestValidatePlugin(t *testing.T) {
 			plugin:     testutil.NewPlugin().WithName("foo").WithVersion("v01.02.3-a").V(),
 			wantErr:    true,
 		},
+		{
```
ahmetb: this is off-topic for this PR but added her as this case wasn't covered.

hack/verify-code-patterns.sh
```diff
@@ -43,3 +43,17 @@ if [[ -n "$out" ]]; then
...
   exit 1
 fi
+
+# Do not initialize index.{Plugin,Platform} structs in test code.
+#
+# TODO(ahmetb): ideally this plugin should cleanly scan all (index\.)(Plugin|Platform) occurrences except slice literals
+# (which are prefixed with []). However I could not get negative lookbehind pattern to work with grep to match to ones
+# that do not have a [] prefix.
+out="$(grep --include '*_test.go' --exclude-dir 'vendor/' -EIrn '\s+(index\.)(Plugin|Platform){' || true)"
+if [[ -n "$out" ]]; then
```
ahmetb: also a bit off-topic to this PR (and hacky). but seems to be doing the job so far. ;)

pkg/installation/platform_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"sigs.k8s.io/krew/pkg/index"
+	"sigs.k8s.io/krew/pkg/testutil"
+)
+
+func Test_osArch(t *testing.T) {
+	inOS, inArch := runtime.GOOS, runtime.GOARCH
+	outOS, outArch := osArch()
+	if inOS != outOS {
+		t.Fatalf("returned OS=%q; expected=%q", outOS, inOS)
```
corneliusweig: `Errorf` would do fine here and below.

pkg/installation/platform_test.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	if inArch != outArch {
+		t.Fatalf("returned Arch=%q; expected=%q", outArch, inArch)
+	}
+}
```
corneliusweig: another awkward brace. I know, it's fine ;)

pkg/installation/platform_test.go
```diff
@@ -0,0 +1,79 @@
...
+	os.Setenv("KREW_OS", customOS)
+	os.Setenv("KREW_ARCH", customArch)
+	defer func() {
+		os.Unsetenv("KREW_ARCH")
+		os.Unsetenv("KREW_OS")
+	}()
+
+	outOS, outArch := osArch()
+	if customOS != outOS {
+		t.Fatalf("returned OS=%q; expected=%q", outOS, customOS)
```
corneliusweig:
```suggestion
		t.Errorf("returned OS=%q; expected=%q", outOS, customOS)
```

pkg/installation/platform_test.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	if customArch != outArch {
+		t.Fatalf("returned Arch=%q; expected=%q", outArch, customArch)
+	}
+}
+
+func Test_matchPlatform(t *testing.T) {
+	const targetOS, targetArch = "foo", "amd64"
+	matchingPlatform := testutil.NewPlatform().WithOSArch("foo", "amd64").V()
```
corneliusweig: Could you add another non-matching platform here, where the OS matches, but the architecture differs? Like
```go
nonMatchingPlatform := testutil.NewPlatform().WithOSArch("foo", "arm").V()
```

hack/verify-code-patterns.sh
```diff
@@ -43,3 +43,17 @@ if [[ -n "$out" ]]; then
   echo >&2 "$out"
   exit 1
 fi
+
+# Do not initialize index.{Plugin,Platform} structs in test code.
+#
+# TODO(ahmetb): ideally this plugin should cleanly scan all (index\.)(Plugin|Platform) occurrences except slice literals
+# (which are prefixed with []). However I could not get negative lookbehind pattern to work with grep to match to ones
+# that do not have a [] prefix.
+out="$(grep --include '*_test.go' --exclude-dir 'vendor/' -EIrn '\s+(index\.)(Plugin|Platform){' || true)"
```
corneliusweig:
```suggestion
out="$(grep --include '*_test.go' --exclude-dir 'vendor/' -EIrn '[^]](index\.)(Plugin|Platform){' || true)"
```
This works with GNU grep. Can you check?

hack/verify-code-patterns.sh
corneliusweig: I think it's a good idea to have this test. If we find it is too restricting, we can still re-evaluate again.

corneliusweig:
@ahmetb Off-topic: as long as you don't have to rebase, could you put your review amendments in a separate commit? That makes the second review round easier.

hack/verify-code-patterns.sh
ahmetb: Can you explain `[^]]`? To me it reads like `[^]` is a negated set but it's empty, then there's a `]` literal outside that's supposed to match.

corneliusweig: I didn't know either. But if there is a negated character set, the first char following the caret is always interpreted verbatim. So this is indeed a negated char set of just `]`. Iow it matches anything but a `]`, which is what we need here, right?

pkg/installation/platform_test.go
ahmetb: I don't get it :)

hack/verify-code-patterns.sh
ahmetb: try on https://www.regexr.com

ahmetb: https://regexr.com/4ht16

corneliusweig: I didn't find out what regex standard regexr.com uses. My GNU grep 3.1 does what I described, I just double-checked. Have you checked, if it also works with BSD grep?

corneliusweig: I finally found a proper reference: `man 7 regex` says

> To  include a literal ']' in the list, make it the first character (following a possible '^').  To include a literal '-', make it the first or last character, or the
       second endpoint of a range.

ahmetb: For gnu grep, `[^]]` and `[^\]]` are different things:

```
$ grep --include '*_test.go' --exclude-dir 'vendor/' -EIrn '[^\]](index\.)(Plugin|Platform){'
cmd/validate-krew-manifest/main_test.go:177:	err := isOverlappingPlatformSelectors([]index.Platform{p1, p2})
cmd/validate-krew-manifest/main_test.go:192:	err := isOverlappingPlatformSelectors([]index.Platform{p1, p2})

$ grep --include '*_test.go' --exclude-dir 'vendor/' -EIrn '[^]](index\.)(Plugin|Platform){'
cmd/validate-krew-manifest/main_test.go:169:	p1 := index.Platform{
cmd/validate-krew-manifest/main_test.go:174:	p2 := index.Platform{
cmd/validate-krew-manifest/main_test.go:184:	p1 := index.Platform{
cmd/validate-krew-manifest/main_test.go:189:	p2 := index.Platform{
```

BSD grep is **very** dumb. It accepts arguments but then accepts input from STDIN (its cli is not fully compatible with gnu grep).

```
$ /usr/bin/grep --include '*_test.go' --exclude-dir 'vendor/' -EIrn '[^]](index\.)(Plugin|Platform){' | cat
grep: warning: recursive search of stdin
```

maybe we should try to fix but I don't care much.

ahmetb: I've incorporated your recommendation.

PR #288, opened 2019-07-22
pkg/validation: add selector validation

### LINKED ISSUE #282: pkg/index/validate: nil Platform selectors
While refactoring tests, I've noticed we have index.Platform{Selector:nil}
usages that haven't been caught.

The behavior for nil selectors are currently not defined. I'm assuming since
this field is a pointer (*metav1.LabelSelector), it matches nothing.
However, allowing this field to be empty further invites nil dereference
issues, though I haven't seen anyone hit this, mostly because we review every
plugin manifest today.

We should add validation checking Selector != nil.

/kind cleanup

### LINKED ISSUE #287: pkg/validate: Validate keys in platform selectors
Looks like we are not really validating keys that go into the platform.selector
field right now.

/kind backlog
/priority important-soon

ahmetb:
- unexport validatePlatform as it's not used outside pkg/validation.
- introduce validateSelector that validates selectors for:
  - unsupported keys
  - nil selector
  - empty selector (both for matchLabels and matchExpressions)
  as these situations are not supported, such as no plugin is valid for all
  platforms supported by krew.

Fixes #282, fixes #287.

### DIFF
--- a/pkg/index/validation/validate.go
+++ b/pkg/index/validation/validate.go
@@ -124,6 +124,9 @@ func validateSelector(sel *metav1.LabelSelector) error {
 	if sel == nil {
 		return errors.New("nil selector is not supported")
 	}
+	if sel.MatchLabels == nil && len(sel.MatchExpressions) == 0 {
+		return errors.New("empty selector is not supported")
+	}
 
 	// check for unsupported keys
 	keys := []string{}
--- a/pkg/index/validation/validate.go
+++ b/pkg/index/validation/validate.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"sigs.k8s.io/krew/pkg/constants"
 	"sigs.k8s.io/krew/pkg/index"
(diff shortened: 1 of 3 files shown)


pkg/index/validation/validate_test.go
```diff
@@ -204,14 +193,80 @@ func TestValidatePlatform(t *testing.T) {
...
+		name    string
+		sel     *metav1.LabelSelector
+		wantErr bool
+	}{
+		{
+			name:    "nil selector",
+			sel:     nil,
+			wantErr: true,
+		},
+		{
```
corneliusweig: This test does not pass, and should be added:
```go
                 {
			name:    "wildcard selector",
			sel:     &metav1.LabelSelector{},
			wantErr: true,
		},
```

ahmetb: That's a great catch. Added a commit.

pkg/index/validation/validate.go
```diff
@@ -124,6 +124,9 @@ func validateSelector(sel *metav1.LabelSelector) error {
 	if sel == nil {
 		return errors.New("nil selector is not supported")
 	}
+	if sel.MatchLabels == nil && len(sel.MatchExpressions) == 0 {
```
corneliusweig: To be on the safe side, this should be
```go
if len(sel.MatchLabels) == 0 && len(sel.MatchExpressions) == 0 {
```

Otherwise, `sel.MatchLabels` could be non-nil, but empty. And that would match everything, right?

ahmetb: we have a check for that case below?

corneliusweig: Oh, I see.

PR #296, opened 2019-07-23
pkg/installation: cleanup refactoring

### LINKED ISSUE #293: Proposed pkg/installation refactor
/kind cleanup
/priority P1

Problems:

- some methods are getting out of control, exposing code smell e.g.
    - `func downloadAndMove(version, sha256sum, uri string, fos []index.FileOperation, downloadPath, installPath, forceDownloadFile string) (dst string, err error) {`
    - `func install(plugin, version, sha256sum, uri, bin string, p environment.Paths, fos []index.FileOperation, forceDownloadFile string) error {`
    - `func getDownloadTarget(index index.Plugin) (version, sha256sum, uri string, fos []index.FileOperation, bin string, err error) {`

    ^ The repetition in method signatures clearly show a design mistake. As
    we add more download options (e.g. `forceDownloadFile` parameter above,
    also poorly named) these do not scale.

- variables exposing a large functionality set (like `environment.Paths` are
  propagated deep inside the call chain. Ideally, we should find what's used,
  by precomputing these values, and avoid passing them down the call stack).

    - similarly, variables that are used only as path components (like `version`)
      are propagated deep down the call stack.

- `func downloadAndMove` does seemingly two independent operations (i.e.
  downloading+verifying, and moving files. This method is very hard to test.

- `func getDownloadTarget` does nothing meaningful that's worthy of testing
  (and is poorly named)

- `func downloadAndMove` is returning the path the files are moved to (via
  `func moveToInstallDir`, which takes `version` (mentioned above). Instead,
  `moveToInstallDir` should be given the path so it doesn't have to compute
  and return this. Also the returned value is not used.

  - the path value returned by `downloadAndMove` is used in `install()` in a
    rather weird way: we compute its absolute path, and use `pathutil.IsSubPath`
    to determine if it's within the install path.

    I can't understand why this check exists, I think we added it at some point
    to make sure the `installpath+bin` is not a symlink that's referencing
    outside the filesystem. (e.g. the `bin` file inside the tarball was a symlink
    that's like ../../../usr/bin/malicious).

    Any help understanding this is appreciated.

Proposed solution:

- Create a `type InstallPlan struct` that contains the information to perform
  installation (e.g. paths precomputed, download/checksum unfo, symlink target)
  and make internal methods use it (but avoid making this a "property bag").

    - Perhaps introduce an `InstallOpts` type to add options like archive
      override.

- Unroll `getDownloadTarget` into Install()/Upgrade() methods, as all it does
  is to match to a Platform spec.

- Split `downloadAndMove` to do less things + testable.

I'm not sure where to start, but I intend to send small PRs about these over
the next weeks.

ahmetb:
- `getDownloadTarget()`: removed, it didn't do anything interesting.
- `downloadAndMove()`:
    - renamed to `downloadAndExtract()` which better reflects the functionality
    - reduced methods from 7 to 4
    - refactored the "move" operation out of this method (decouples "move"
      from "download+extract") and we now call "move" from install().
    - added unit tests (both for downloading a file, and --archive override)

- introduced `installOperation` which contains the minimal information to
  perform installation operation.

- introduced `InstallOpts` type to scale out (in the future)

- `install()`:
    - accepts fewer arguments (2) than before (8)
    - no longer knows about index.Plugin, or environment.Paths

- cmd/install: removed a redundant check detecting specifying plugin names
  along with --manifest option.

Fixes #293
/kind cleanup
/assign @corneliusweig

### DIFF
--- a/pkg/installation/install.go
+++ b/pkg/installation/install.go
@@ -30,39 +30,34 @@ import (
 	"sigs.k8s.io/krew/pkg/pathutil"
 )
 
-// Plugin Lifecycle Errors
+// InstallOpts specifies options for plugin installation operation.
+type InstallOpts struct {
+	ArchiveFileOverride string
+}
+
+type installOperation struct {
+	pluginName string
+	platform   index.Platform
+
+	downloadStagingDir string
+	installDir         string
+	binDir             string
+}
+
+const (
+	krewPluginName = "krew"
+)
+
+// Plugin lifecycle errors
 var (
 	ErrIsAlreadyInstalled = errors.New("can't install, the newest version is already installed")
 	ErrIsNotInstalled     = errors.New("plugin is not installed")
 	ErrIsAlreadyUpgraded  = errors.New("can't upgrade, the newest version is already installed")
 )
 
-const (
-	krewPluginName = "krew"
-)
-
-func downloadAndMove(version, sha256sum, uri string, fos []index.FileOperation, downloadPath, installPath, forceDownloadFile string) (dst string, err error) {
-	glog.V(3).Infof("Creating download dir %q", downloadPath)
-	if err = os.MkdirAll(downloadPath, 0755); err != nil {
-		return "", errors.Wrapf(err, "could not create download path %q", downloadPath)
-	}
-	defer os.RemoveAll(downloadPath)
-
-	var fetcher download.Fetcher = download.HTTPFetcher{}
-	if forceDownloadFile != "" {
-		fetcher = download.NewFileFetcher(forceDownloadFile)
-	}
-
-	verifier := download.NewSha256Verifier(sha256sum)
-	if err := download.NewDownloader(verifier, fetcher).Get(uri, downloadPath); err != nil {
-		return "", errors.Wrap(err, "failed to download and verify file")
-	}
-	return moveToInstallDir(downloadPath, installPath, version, fos)
-}
-
 // Install will download and install a plugin. The operation tries
 // to not get the plugin dir in a bad state if it fails during the process.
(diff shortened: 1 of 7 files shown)


pkg/installation/install.go
```diff
@@ -71,42 +66,79 @@ func Install(p environment.Paths, plugin index.Plugin, forceDownloadFile string)
...
+	if err := downloadAndExtract(op.downloadStagingDir, op.platform.URI, op.platform.Sha256, opts.ArchiveFileOverride); err != nil {
+		return errors.Wrap(err, "failed to download and extract")
 	}
 
-	subPathAbs, err := filepath.Abs(dst)
+	if err := moveToInstallDir(op.downloadStagingDir, op.installDir, op.platform.Files); err != nil {
+		return errors.Wrap(err, "failed while moving files to the installation directory")
+	}
+
+	subPathAbs, err := filepath.Abs(op.installDir)
```
corneliusweig: Could you extract the whole sub-path checking logic to a separate function?

ahmetb: Can we do that in another PR? 😄

ahmetb:
> Finally `install` becomes somewhat testable. This could be one of the next steps. For that we just need to override `downloadAndExtract` or mock the http server as you already did.

Yeah, I think `install` is becoming more approachable for testing.
Although I think we should also add tests for `downloadAndExtract`, and some other `moveXxx` methods. I'm less worried about `install`

pkg/installation/install.go
corneliusweig: sure :)

PR #297, opened 2019-07-23
integration_test: add more test cases

### LINKED ISSUE #294: integration_test: cover behavior for re-running commands and more negative cases
I think we're lacking some tests around "what happens when I run this command
again". For example, most prominently:

- `krew install` should fail second time if first time succeeds.
- `krew remove` should fail if first time succeeds (ensures we cleaned up the
  receipt correctly, otherwise krew might think the plugin is still there).
- `krew upgrade` should succeed second time if first time upgrades a plugin

/label good first issue
/priority P2
/kind cleanup

### LINKED ISSUE #295: integration_test: increase error handling coverage
Some cases in cmd/krew/* are not covered by integration tests. Examples:

- `install`: `--manifest` and positional arguments cannot be specified
- `install`: cannot specify plugin names both via STDIN and positional arguments
- `install`: test if plugins can be installed by specifying names via STDIN
  (i.e. 'restore' use case)
- `list`: test if plugin names can be saved to a file properly when STDOUT is
  not a tty (i.e. 'backup' use case)

/kind cleanup
/priority P2

ahmetb:
Added more tests or expanded existing tests to:

- cover args/stdin handling edge cases for `install` cmd
- cover status code of re-runs of `install`, `update`, and `remove`
- cover status when index not initialized in `remove`, `upgrade` and `install`
- several other minor adjustments

Fixes #294
Fixes #295

### DIFF
--- a/integration_test/install_test.go
+++ b/integration_test/install_test.go
@@ -49,8 +49,9 @@ func TestKrewInstallReRun(t *testing.T) {
 	test, cleanup := NewTest(t)
 	defer cleanup()
 
-	test.WithIndex().Krew("install", validPlugin).RunOrFail()
-	test.WithIndex().Krew("install", validPlugin).RunOrFail()
+	test = test.WithIndex()
+	test.Krew("install", validPlugin).RunOrFail()
+	test.Krew("install", validPlugin).RunOrFail()
 	test.AssertExecutableInPATH("kubectl-" + validPlugin)
 }
 
--- a/integration_test/install_test.go
+++ b/integration_test/install_test.go
@@ -16,6 +16,7 @@ package integrationtest
 
 import (
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"sigs.k8s.io/krew/pkg/constants"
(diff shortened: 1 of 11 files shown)


integration_test/install_test.go
```diff
@@ -27,12 +28,68 @@ const (
...
+}
+
+func TestKrewInstall_MultiplePositionalArgs(t *testing.T) {
+	skipShort(t)
 
 	test, cleanup := NewTest(t)
 	defer cleanup()
 
-	test.WithIndex().Krew("install", validPlugin).RunOrFailOutput()
+	test.WithIndex().Krew("install", validPlugin, validPlugin2).RunOrFailOutput()
```
corneliusweig:
```suggestion
	test.Krew("install", validPlugin, validPlugin2).RunOrFailOutput()
```
If you want to keep the symmetry, this is fine, but it's not necessary.

integration_test/install_test.go
```diff
@@ -27,12 +28,68 @@ const (
...
+	test.AssertExecutableInPATH("kubectl-" + validPlugin)
+}
+
+func TestKrewInstallReRun(t *testing.T) {
+	skipShort(t)
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	test.WithIndex().Krew("install", validPlugin).RunOrFail()
+	test.WithIndex().Krew("install", validPlugin).RunOrFail()
```
corneliusweig:
```suggestion
	test.Krew("install", validPlugin).RunOrFail()
```

ahmetb: Um I think both are wrong.

this should be just

test=test.WithIndex()
test.krew("install", validplugin)

ahmetb:
> Was that intentional? Myself, I cannot think of a way how to get proper output for ttys, so it looks not testable to me.

I think `krew list > file.txt` is supposed to be printing only names (not a table). Let's check.

ahmetb:
yep:

```
k krew list | cat
get-all
konfig
krew
mtail
view-secret
who-can
```

I'll add a test for this as well.

ahmetb:
`krew list` is actually tested fine, I'm now updating its tests to be more precise.

ahmetb:
Fixed now.

corneliusweig:
You are really lifting krew's code quality to the next level. :clap: for your efforts!
/lgtm
/approve

PR #300, opened 2019-07-31
Validate plugins can be removed without local manifest

ahmetb:
This validates the scenario that we can remove/rename plugins in the index
without breaking local user installations (since this is now possible with
the install receipts).

/assign @corneliusweig

### DIFF
--- a/integration_test/uninstall_test.go
+++ b/integration_test/uninstall_test.go
@@ -14,7 +14,13 @@
 
 package integrationtest
 
-import "testing"
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"sigs.k8s.io/krew/pkg/constants"
+)
 
 func TestKrewUninstall(t *testing.T) {
 	skipShort(t)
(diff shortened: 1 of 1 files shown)


integration_test/uninstall_test.go
```diff
@@ -49,3 +55,22 @@ func TestKrewRemove_AliasSupported(t *testing.T) {
...
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	test = test.WithIndex()
+	localManifest := filepath.Join(test.Root(), "index", "plugins", validPlugin+constants.ManifestExtension)
+	if _, err := os.Stat(localManifest); err != nil {
+		t.Fatalf("could not read local manifest file at %s: %v", localManifest, err)
+	}
+	test.Krew("install", validPlugin).RunOrFail()
+	if err := os.RemoveAll(localManifest); err != nil {
```
corneliusweig:
```suggestion
	if err := os.Remove(localManifest); err != nil {
```
should do as well.

integration_test/uninstall_test.go
```diff
@@ -49,3 +55,22 @@ func TestKrewRemove_AliasSupported(t *testing.T) {
...
+		t.Fatalf("failed to remove local manifest file: %v", err)
+
```
corneliusweig: That empty line looks a bit awkward

integration_test/uninstall_test.go
```diff
@@ -49,3 +55,22 @@ func TestKrewRemove_AliasSupported(t *testing.T) {
...
+	}
+	test.Krew("remove", validPlugin)
```
corneliusweig:
```suggestion
	test.Krew("remove", validPlugin).RunOrFail()
```

integration_test/uninstall_test.go
```diff
@@ -49,3 +55,22 @@ func TestKrewRemove_AliasSupported(t *testing.T) {
...
 }
+
+func TestKrewRemove_ManifestRemovedFromIndex(t *testing.T) {
+	skipShort(t)
+
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	test = test.WithIndex()
+	localManifest := filepath.Join(test.Root(), "index", "plugins", validPlugin+constants.ManifestExtension)
```
corneliusweig: I'm a bit unhappy that we depend on the implementation detail of where the manifest is stored. It would be cleaner to use `environment.Paths` here. Maybe it's time to refactor the `environment` package, so that it can be used without setting up env variables.

ahmetb: Honestly I agree, but we'd immediately know when we change the paths because the test would start breaking.

ahmetb: That said I think we can expose `NewPaths()` here that doesn't rely on environment variables, for testing or other purposes:

image

corneliusweig: Ok, let's do that in another PR and close this one. I think there are more locations which can be made simpler.

PR #302, opened 2019-07-31
Forbid long lines in manifests during validation

### LINKED ISSUE #298: validation: Enforce line lengths in manifests?
Currently, the manifest validation does not consider line length, so that manifests can contain very long lines. This can lead to badly formatted output on small terminals. Should we therefore enforce a maximum line length? For terminals, a 80 character width is standard.

Further, are there more formatting validations which could be enforced during validation?

corneliusweig:
This PR adds checks for small line lengths in manifest text blocks. This is desired for proper formatting on small screens.

I opted against hard checks, because there are good reasons to sometimes neglect the length limitation. For example, long URLs.

Close #298

### DIFF
--- a/cmd/validate-krew-manifest/main.go
+++ b/cmd/validate-krew-manifest/main.go
@@ -77,13 +77,13 @@ func validateManifestFile(path string) error {
 	}
 	glog.Infof("structural validation OK")
 
-	if maxLen := 80; maxLen < getMaxLinelength(p.Spec.Description) {
+	if maxLen := 80; !validateLineLength(p.Spec.Description, maxLen) {
 		glog.Warningf("line length in `description` exceeds %d characters", maxLen)
 	}
-	if maxLen := 70; maxLen < getMaxLinelength(p.Spec.ShortDescription) {
+	if maxLen := 70; !validateLineLength(p.Spec.ShortDescription, maxLen) {
 		glog.Warningf("line length in `shortDescription` exceeds %d characters", maxLen)
 	}
-	if maxLen := 80; maxLen < getMaxLinelength(p.Spec.Caveats) {
+	if maxLen := 80; !validateLineLength(p.Spec.Caveats, maxLen) {
 		glog.Warningf("line length in `caveats` exceeds %d characters", maxLen)
 	}
 
@@ -208,12 +208,16 @@ func allPlatforms() [][2]string {
 	}
 }
 
-func getMaxLinelength(s string) int {
-	max := 0
+// validateLineLength checks if the maximum line length is below maxLen.
+// Lines may exceed this limit, if they are just one word, i.e. without spaces.
+func validateLineLength(s string, maxLen int) bool {
 	for _, line := range strings.Split(s, "\n") {
-		if l := len(line); max < l {
-			max = l
+		if len(line) <= maxLen {
+			continue
+		}
+		if strings.Contains(strings.TrimSpace(line), " ") {
+			return false
 		}
 	}
-	return max
+	return true
 }
--- a/cmd/validate-krew-manifest/main.go
+++ b/cmd/validate-krew-manifest/main.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -215,9 +216,18 @@ func validateLineLength(s string, maxLen int) bool {
 		if len(line) <= maxLen {
 			continue
 		}
-		if strings.Contains(strings.TrimSpace(line), " ") {
+		if !containsURI(line) {
 			return false
 		}
 	}
 	return true
+}
+
+func containsURI(line string) bool {
+	for _, word := range strings.Fields(line) {
+		if _, err := url.ParseRequestURI(word); err == nil {
+			return true
+		}
+	}
+	return false
 }
--- a/cmd/validate-krew-manifest/main.go
+++ b/cmd/validate-krew-manifest/main.go
@@ -78,13 +78,13 @@ func validateManifestFile(path string) error {
 	glog.Infof("structural validation OK")
 
 	if maxLen := 80; !validateLineLength(p.Spec.Description, maxLen) {
-		glog.Warningf("line length in `description` exceeds %d characters", maxLen)
+		return fmt.Errorf("line length in `description` exceeds %d characters", maxLen)
 	}
 	if maxLen := 70; !validateLineLength(p.Spec.ShortDescription, maxLen) {
-		glog.Warningf("line length in `shortDescription` exceeds %d characters", maxLen)
+		return fmt.Errorf("line length in `shortDescription` exceeds %d characters", maxLen)
 	}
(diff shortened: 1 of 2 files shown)


ahmetb:
I have a feeling like people won't see the warnings, mostly because they don't arrive to the PR via a chatbot message. :|

So if we don't fail the build, they'll go unchecked for sure.

corneliusweig:
Yes, I kind of agree. I also don't look at the CI output as long as it is green. But we can't fail the validation, can we? Or do you think we could simply exclude lines with urls?

Otherwise, it probably doesn't make much sense to include this.

corneliusweig:
I'd like to re-iterate the idea of checking the line length in manifests.

So the logic is now that long lines are forbidden, unless the line is a single word (i.e. does not contain a space). Additionally, the line length constraint is now a hard constraint.

corneliusweig:
> Sounds good, how's the status on the current plugin manifests. Do they all pass?

Actually it looks more like one half of our manifests violate some length constraint. In most cases this was due to `shortDescription` as far as I remember.

cmd/validate-krew-manifest/main.go
```diff
@@ -77,6 +77,16 @@ func validateManifestFile(path string) error {
 	}
 	glog.Infof("structural validation OK")
 
+	if maxLen := 80; !validateLineLength(p.Spec.Description, maxLen) {
+		return fmt.Errorf("line length in `description` exceeds %d characters", maxLen)
+	}
+	if maxLen := 70; !validateLineLength(p.Spec.ShortDescription, maxLen) {
+		return fmt.Errorf("line length in `shortDescription` exceeds %d characters", maxLen)
+	}
+	if maxLen := 80; !validateLineLength(p.Spec.Caveats, maxLen) {
```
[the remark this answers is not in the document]
corneliusweig: Yes, I was also considering to look for `http(s)?...` and ignore those lines for length validation. This may be the better option, but it's also a little more complex.

corneliusweig:
What do you think about allowing long lines whenever they contain a URL?

ahmetb:
That sounds reasonable.

ferhatelmas:
> What do you think about allowing long lines whenever they contain a URL?

This is also something done by [prettier](https://github.com/prettier/prettier).
If it can't break the word, it leaves it on the line.

cmd/validate-krew-manifest/main.go
```diff
@@ -77,6 +78,16 @@ func validateManifestFile(path string) error {
 	}
 	glog.Infof("structural validation OK")
 
+	if maxLen := 80; !validateLineLength(p.Spec.Description, maxLen) {
+		return fmt.Errorf("line length in `description` exceeds %d characters", maxLen)
```
ferhatelmas: why don't we use `pkg/errors`?

cmd/validate-krew-manifest/main.go
```diff
@@ -77,6 +78,16 @@ func validateManifestFile(path string) error {
...
+	}
+	if maxLen := 70; !validateLineLength(p.Spec.ShortDescription, maxLen) {
```
ferhatelmas: IMHO, `isValidLineLength` signals what it does better.

cmd/validate-krew-manifest/main.go
```diff
@@ -197,3 +208,26 @@ func allPlatforms() [][2]string {
 		{"darwin", "amd64"},
 	}
 }
+
+// validateLineLength checks if the maximum line length is below maxLen.
+// Lines may exceed this limit, if they are just one word, i.e. without spaces.
+func validateLineLength(s string, maxLen int) bool {
+	for _, line := range strings.Split(s, "\n") {
```
ferhatelmas: how about carriage return `\r`?

cmd/validate-krew-manifest/main.go
```diff
@@ -197,3 +208,26 @@ func allPlatforms() [][2]string {
...
+		if len(line) <= maxLen {
```
ferhatelmas: how about runes?

cmd/validate-krew-manifest/main_test.go
```diff
@@ -183,3 +183,58 @@ func Test_isOverlappingPlatformSelectors_overlap(t *testing.T) {
...
+			name: "long line with a URL",
+			input: "	https://krew.dev/is_really_awesome   ",
+			expected: true,
+		},
+		{
+			name: "multiple lines with long line containing a URL",
+			input: `
+line below limit
+line below limit
+This is a long line with a URL https://krew.dev/is_really_awesome
```
ferhatelmas: `long url ..... some short words`

this test would be beneficial since last fields actually should be on their own line.

cmd/validate-krew-manifest/main.go
corneliusweig: It should not happen, but better safe than sorry :)

cmd/validate-krew-manifest/main.go
corneliusweig: I agree, the function name is o_O

cmd/validate-krew-manifest/main.go
corneliusweig: I think as long as the error message is specific enough, it is more user-friendly to have a single line instead of a lengthy stack trace.

cmd/validate-krew-manifest/main.go
corneliusweig: good point

cmd/validate-krew-manifest/main_test.go
corneliusweig: If I understand correctly, you mean require that the URL is the last word in the line exceeding the allowed line length limit? I think we should not overdo things. Whatever logic we choose needs to be simple and comprehensible.

ahmetb:
/retitle Disable long lines on manifests during validation

cmd/validate-krew-manifest/main_test.go
corneliusweig: @ferhatelmas Would you be ok with the current logic? If we find it insufficient we can still reiterate.

cmd/validate-krew-manifest/main.go
ferhatelmas: SGTM

cmd/validate-krew-manifest/main_test.go
ferhatelmas: I am good with current implementation. No need to be blocked.

My suggestion for an improvement.

For example, `long-url and some text`, even if it is a long line and could be split like `long-url` and `and some text`, it will pass validation due to url.

ferhatelmas: Should we create an issue to keep track of it or is it a minor?

corneliusweig: So I think we can gather some experience with the current logic and revisit later. But if you are convinced that we should revisit this at some point, there's no harm in opening an issue right now.

corneliusweig:
@ahmetb WDYT? Should we give this a try?

corneliusweig:
I think the fact that this has been open for so long shows that there is simply no consensus around forbidding long lines in manifests.
If you think we should revisit this, please leave a comment.

PR #307, opened 2019-08-18
Add index revision to plugin overview page

corneliusweig:
When resolving the index revision, be careful to always report a commit sha and not branch names.

### DIFF
--- a/cmd/generate-plugin-overview/generate-plugin-overview.go
+++ b/cmd/generate-plugin-overview/generate-plugin-overview.go
@@ -32,13 +32,16 @@ const (
 	separator  = " | "
 	pageHeader = `## Available kubectl plugins
 
-The following kubectl plugins are currently available
-on [Krew plugin index](https://sigs.k8s.io/krew-index).
-
 To install these kubectl plugins:
 
 1. [Install Krew](https://github.com/kubernetes-sigs/krew#installation)
 2. Run ` + "`kubectl krew install PLUGIN_NAME`" + ` to install a plugin via Krew.
+
+The following kubectl plugins are currently available on
+[Krew plugin index](https://sigs.k8s.io/krew-index). Note that this table may be
+outdated. For the most up-to-date list of plugins, visit the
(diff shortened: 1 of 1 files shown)


cmd/generate-plugin-overview/generate-plugin-overview.go
```diff
@@ -32,13 +33,14 @@ const (
 	separator  = " | "
 	pageHeader = `## Available kubectl plugins
 
-The following kubectl plugins are currently available
-on [Krew plugin index](https://sigs.k8s.io/krew-index).
+The following kubectl plugins are available on
+[Krew plugin index](https://sigs.k8s.io/krew-index) as of revision %s.
```
ahmetb: I don’t think most users care about commit numbers. 
Do you think this will serve a good purpose? Instead we can just tell them “run this command for the most up-to-date list”.

corneliusweig:
Hmm 🤔, I agree that the exact commit number is not so important. But I also don't think that any user would run our page generator.

Basically, what needs to come across is that the page may be outdated. We can also add such a note without mentioning the commit number. But I thought in the current way a user can determine what has happened since last generating that overview page.

ahmetb:
> But I thought in the current way a user can determine what has happened since last generating that overview page.

For that to happen I think we actually need to link to plugins/ directory at that revision, and at the HEAD. I don't think anyone is realistically going to diff the two revisions.

So I recommend maybe we can just say:

> for the most update list of plugins, visit [krew-index](#) repository or run `kubectl krew search`

wdyt?

corneliusweig:
> So I recommend maybe we can just say:
> 
> > for the most up-to-date list of plugins, visit [krew-index](#) repository or run `kubectl krew search`
> 
> wdyt?

Ok with me. I'll update the PRs

ahmetb:
Reminder to push updates here as well.

corneliusweig:
> Reminder to push updates here as well.

Sorry, that slipped through. I thought I had done that. Can you take a look?

ahmetb:
/lgtm
thanks!

Issue #292, opened 2019-07-22
Configure linter for unused methods


ahmetb:
Every now and then, we do major refactoring that leaves unused methods behind.

It would be great to have a linter that can notify us about unused methods,
variables, constants. I suspect this is probably easier to configure with "unexported" methods,
variables etc. However, if there's a way to do this with "exported" methods,
variables etc, we should do it because now anything in this repository (sigs.k8s.io/krew/...)
is expected to be used in this repository. So if something is not referenced,
ideally it should be cleaned up.

/label good-first-issue
/priority P3
/kind cleanup

corneliusweig:
Hmm.. I just tried configuring the `unused` linter for `golangci-lint` with
```patch
diff --git a/.golangci.yml b/.golangci.yml
index ff0bbe9..3a53d8c 100644
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -5,6 +5,9 @@ linters-settings:
     # it's a comma-separated list of prefixes
     local-prefixes: sigs.k8s.io/krew
 
+  unused:
+    check-exported: false
+
 # options for analysis running
 run:
   # which dirs to skip: they won't be analyzed;
```

That setting reports a lot of false positives, which makes it impossible to use in the CI. Btw, the `unused` linter is active by default. However, it does not consider functions as unused, if they are used in tests. So if all tests are deleted, the unused function from #290 is correctly reported. So it seems that our current meta-linter cannot handle this as desired.

ahmetb:
> So if all tests are deleted,

haha that's a bummer. :D They should totally have a flag to ignore usage in tests.

ferhatelmas:
For a quick win, how about enabling `deadcode` and `varcheck` with `tests` flag?

ferhatelmas:
Not very useful at the moment but there might be beneficial cases.

ferhatelmas:
https://github.com/kubernetes-sigs/krew/issues/291 would be useful because packages under internal can only be used in the same repo. Otherwise, they are exported, linter won't complain to prevent false positive.

PR #318, opened 2019-08-26
lint: enable deadcode and varcheck

ferhatelmas:
related to #292

didn't enable `tests: true` since they are included by [default](https://github.com/golangci/golangci-lint/blob/master/.golangci.example.yml#L16)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -31,5 +31,7 @@ linters:
     - staticcheck
     - structcheck
     - prealloc
+    - deadcode
+    - varcheck
   disable:
     - errcheck


corneliusweig:
https://golangci.com/product#linters says

> varcheck: Finds unused global variables and constants.
deadcode: Finds unused code.

More info here: https://github.com/remyoudompheng/go-misc/tree/master/deadcode
and https://gitlab.com/opennota/check

corneliusweig:
So I just checked, and also `deadcode` won't report functions which are no longer used by production code, but called by some test code :sob:

corneliusweig:
True, true. So let's get this merged then
/lgtm
/approve

PR #319, opened 2019-08-26
lint: bump golangci-lint to 1.17.1

ferhatelmas:
while checking https://github.com/kubernetes-sigs/krew/pull/318, realized there is a newer version of the metalinter so here it is.


corneliusweig:
Ah.. we switched to a newer version of go, which prevented that earlier #211

PR #320, opened 2019-08-26
lint: make test config explicit

ferhatelmas:
addresses the request from [this comment](https://github.com/kubernetes-sigs/krew/pull/318#issuecomment-524986118) by explicitly including test files into lint

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -16,6 +16,9 @@ run:
     - hack
     - docs
 
+  # include test files
+  tests: true
+
 linters:
   enable:
     - gocritic


corneliusweig:
This setting is `true` by default (see https://github.com/golangci/golangci-lint/blob/d2b1eea2c6171a1a1141a448a745335ce2e928a1/.golangci.example.yml#L16). Is there a particular reason why you want to make that explicit?

ferhatelmas:
Yes, it's true by default but as in the [linked comment](https://github.com/kubernetes-sigs/krew/pull/318#issuecomment-524986118), @ahmetb suggested to make it explicit unless hurts.

corneliusweig:
Hmm.. if either of you feel strongly about making this explicit, I won't say no. But it feels a bit redundant.

PR #321, opened 2019-08-27
Correct import order

ferhatelmas:
https://github.com/kubernetes-sigs/krew/blob/master/.golangci.yml#L6 wasn't applied correctly. 

This PR fixes existing issues but the root cause is still there. Since golangci-lint won't be fixed soon. I suggest integrating [impi](https://github.com/pavius/impi) in the short run (as a follow up PR or update on this)

### DIFF
--- a/cmd/krew/cmd/install.go
+++ b/cmd/krew/cmd/install.go
@@ -19,14 +19,14 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/golang/glog"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
 	"sigs.k8s.io/krew/pkg/index"
 	"sigs.k8s.io/krew/pkg/index/indexscanner"
 	"sigs.k8s.io/krew/pkg/index/validation"
 	"sigs.k8s.io/krew/pkg/installation"
-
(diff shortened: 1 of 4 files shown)


corneliusweig:
Hey thanks for bringing this up! This is indeed an annoying loophole in the current rules. Let's merge this now and evaluate impi in a separate PR.

Do you happen to know the golangci-lint issue which tracks this bug?
/lgtm
/approve

corneliusweig:
Yes, goland is not great at this. But ultimately, this should be the job of goimports to set it right. My annoying workaround is currently, to put all imports into one single block and then run goimports.

ferhatelmas:
Actually, it's not a golangci-lint issue but goimports.

Canonical issue: https://github.com/golang/go/issues/20818
A failed attempt to fix it: https://github.com/mvdan/gofumpt/issues/8

Also there are various complaints in golangci-lint but there are mostly irrelevant, not being the root cause. That's why I didn't link them, here.

impi is simple since it doesn't format but only validates. When we keep imports consistent and get alerted on wrong order, fixing should be easy then sounds good enough to me.

Green light for impi integration?

corneliusweig:
I'd say yes, let's try this out!

Issue #324, opened 2019-08-30
Release krew v0.3.0


ahmetb:
It seems like there's no reason to hold off onto the v0.3 

- v0.2.1 happened 309 days ago (Oct 24, 2018) and Krew has been working pretty reliably.
- We made a lot of changes since v0.2.1, most of them are cleanup of technical debt. However, these need more testing before more changes are accumulated.
- All P0 and P1 bugs are now closed.

We only have a major backlog item we're deferring to v0.4 release:
- Support for custom index repositories (#23), which will require a long thought process and a detailed design change followed by fundamental core refactoring (which is now easier to do.)

Major notes from this release:

- Post-upgrade, krew will prompt users for running a mandatory migration command before any krew subcommand can be used.
- Krew no longer breaks when an installed plugin deleted/renamed from the upstream index.

/cc @corneliusweig 
^ can you please try [making a release](https://github.com/kubernetes-sigs/krew/blob/master/docs/RELEASING_KREW.md) to exercise this path and see if you have permissions?

corneliusweig:
It seems I am not allowed to directly push to the krew repository or create tags. The git config used for pushing is:
```
[remote "origin"]
   url = git@github.com:kubernetes-sigs/krew.git
   fetch = +refs/heads/*:refs/remotes/origin/*
```

Btw, it would be nice to have the above major notes also in the tag description for `v0.3.0`. If you go ahead with this task could you add them manually?

---

And what do you think about listing the merged PRs. For example, minikube has a script (https://github.com/kubernetes/minikube/blob/master/hack/release_notes/listpullreqs.go) which will produce:

```md
Collecting pull request that were merged since the last release: v0.2.1 (2018-10-24 19:45:20 +0000 UTC)
* Add security notice to be shown after install and upgrade [#316](https://github.com/kubernetes-sigs/krew/pull/316)
* Correct import order [#321](https://github.com/kubernetes-sigs/krew/pull/321)
* lint: make test config explicit [#320](https://github.com/kubernetes-sigs/krew/pull/320)
* lint: enable deadcode and varcheck [#318](https://github.com/kubernetes-sigs/krew/pull/318)
* lint: bump golangci-lint to 1.17.1 [#319](https://github.com/kubernetes-sigs/krew/pull/319)
* Clean up stale krew installations on windows [#314](https://github.com/kubernetes-sigs/krew/pull/314)
[...]

PR #326, opened 2019-09-01
Include list of merged pull requests in release notes

corneliusweig:
As mentioned in https://github.com/kubernetes-sigs/krew/issues/324#issuecomment-526905406

### DIFF
--- a/hack/make-release-notes.sh
+++ b/hack/make-release-notes.sh
@@ -18,6 +18,10 @@
 # You can run this script while tagging the release as:
 #     git tag -a v0.1 -m "$(TAG=v0.1 hack/make-release-notes.sh)"
 
+set -euo pipefail
+
+SCRIPTDIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
 TAG="${TAG:?TAG environment variable must be set for this script}"
 if ! [[ "$TAG" =~ v.* ]]; then
   echo >&2 "TAG must be in format v.*"
@@ -33,6 +37,22 @@ download_assets=(
   krew.exe.sha256
   krew.yaml
 )
+
+install_release_notes() {
+  relnotes_dir="$(mktemp -d)"
+  trap 'rm -rf -- ${relnotes_dir}' EXIT
+
+  cd "$relnotes_dir"
+  go mod init foo
+  GOBIN="${SCRIPTDIR}" go install github.com/corneliusweig/release-notes
+  cd -
+}
+
+# install release-notes if not present in the hack folder
+if ! [[ -x "${SCRIPTDIR}/release-notes" ]]; then
+  echo >&2 'Installing release-notes'
+  install_release_notes
+fi
 
 echo "Installation"
 echo "------------"
@@ -54,3 +74,11 @@ git log "$(git describe --tags --abbrev=0)..HEAD" --format=%an |
   sed -E 's,^(\s+[0-9]+\s),- ,g'
 echo
 echo "(krew ${TAG} is tagged on $(date -u).)"
+echo
+echo '<details>'
+echo '<summary>Merged pull requests</summary>'
+echo # this empty line is important for correct markdown rendering
+# you can pass your github token with --token here if you run out of requests
+"${SCRIPTDIR}/release-notes" kubernetes-sigs krew
+echo '</details>'
+echo
--- a/hack/.gitignore
+++ b/hack/.gitignore
@@ -1,0 +1,1 @@
+/release-notes


ferhatelmas:
why isn't it extracted to its own repo and maintained there?

It's simple but very likely to diverge this way.

ahmetb:
Extra repo mgmt is currently more costly, especially for standalone tools like this.

hack/make-release-notes.sh
```diff
@@ -38,6 +42,10 @@ echo "Installation"
 echo "------------"
 echo "To install this release, refer to the instructions at ${readme}."
 echo
+echo "Merged pull requests"
```
ahmetb: I think this is definitely lower priority than contributors list + release asset links.

Maybe move this to the end?

And maybe if possible we should make it collapsible, like:

<details>
<summary>Pull Requests</summary>
<pre>
Html list here
</pre>
</details>

with

```
<details>
<summary>...</summary>
<pre>
...
</pre>
</details>
```

ferhatelmas:
> Extra repo mgmt is currently more costly, especially for standalone tools like this.

@ahmetb thanks, that's fair but how new changes in tools are propagated between repos right now?

ahmetb:
Looks like this tool is coming from minikube repo. In that case, there potentially could be a common repo, but again, making the case for it could be harder than copying + freezing the code.

In this case, as long as the code works and delivers the same result, we don't need to get new updates to it necessarily. We might even deem the tool unnecessary before it starts breaking.

hack/make-release-notes.sh
corneliusweig: Yes, PRs have low priority, so I moved it to the end. I think it even works with markdown in the collapsed region, if a blank line is inserted. I haven't checked how the generated output renders on github though...

corneliusweig:
@ferhatelmas I _think_ this comes from minikube, however I'm not even 100% sure here, because the same code is found in several repos.
As to upstream changes: we should be safe here, because github has a stable API, so the script will not suddenly stop working. Besides, we don't want another output formatting, so the script simply does what it does.

hack/release_notes/go.mod
```diff
@@ -0,0 +1,10 @@
+module sigs.k8s.io/krew/hack/release_notes
```
ahmetb: Idea: why not make this a sub-module on minikube that we can `go get`?

corneliusweig: Now that's a good idea! I'll open an issue and see if they are open for this.

hack/tools/go.mod
```diff
@@ -0,0 +1,5 @@
+module tools
+
+go 1.12
+
+require github.com/corneliusweig/release-notes v0.0.0-20191014214505-0be5c7c66752
```
ahmetb: maybe tag a release?

hack/tools/tools.go
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import _ "github.com/corneliusweig/release-notes"
```
ahmetb: this is not great. :( introduces a go package here.

let's just do something like this in the script, we already do it elsewhere:

```
temp="$(mkdir -p)"
trap -- 'rm -rf "${temp}"'
cd "$temp"
go mod init foo
go get github.com/corneliusweig/release-notes
# mv to '$(go env GOPATH)/bin' or something like that
```

hack/tools/go.mod
ahmetb: let's say I have this

v0.3.1-alpha.1 -->  v0.3.1 --> v0.3.2 --> v0.3.3-alpha.1 --> v0.3.3

when I generate notes for v0.3.3, does it compute since v0.3.2 –or since v0.3.3-alpha.1?

corneliusweig: That would only include in the range v0.3.3-alpha.1 to present. But it totally makes sense to change that :)

hack/tools/tools.go
corneliusweig: That's also fine with me.

hack/tools/go.mod
ahmetb: Yeah we’d only want changes between revisions went to the user.

corneliusweig: I've introduced an option for the tool, so that it now calculates the difference since v0.3.2 and skips the alpha version. Also, I tagged a release :)

hack/tools/tools.go
corneliusweig: should be fixed now.

corneliusweig:
@ahmetb I finally addressed your comments:

- alpha releases are now ignored by the external tool when determining the latest previous release
- the installation method no longer needs a sub-module.

hack/make-release-notes.sh
```diff
@@ -34,6 +38,22 @@ download_assets=(
   krew.yaml
 )
 
+install_release_notes() {
+  relnotes_dir="$(mktemp -d)"
+  trap 'rm -rf -- ${relnotes_dir}' EXIT
+
+  cd "$relnotes_dir"
+  go mod init foo
+  GOBIN="${SCRIPTDIR}" go install github.com/corneliusweig/release-notes
```
ahmetb: I recall the whole point of this mktemp trick was to pin it to a particular version :)

ahmetb:
I think this is good, two concerns:

- we don't pin to a version of release-notes tool
- we don't use the same semver library as that tool, basically we just need to be able to tell if https://godoc.org/k8s.io/apimachinery/pkg/util/version#Version.PreRelease string empty or not to be able to tell a stable version or not.

These can be addressed later if you want.

/lgtm
/approve

PR #327, opened 2019-09-02
lint: enable errcheck and fix some issues

### LINKED ISSUE #325: Consider enabling errcheck linter
`errcheck` is in general a useful linter but it's disabled at the moment. Unless there is a specific reason to disable it, we might consider enabling it.

/kind cleanup
/kind proposal

ferhatelmas:
* add `.errcheck-excluded.txt` to disable some function calls as an example
* fixes #325

### DIFF
--- a/.errcheck-excluded.txt
+++ b/.errcheck-excluded.txt
@@ -1,0 +1,2 @@
+os.MkdirAll
+io.Copy
--- a/.errcheck-excluded.txt
+++ b/.errcheck-excluded.txt
@@ -1,2 +1,0 @@
-os.MkdirAll
-io.Copy
--- a/cmd/krew/cmd/root.go
+++ b/cmd/krew/cmd/root.go
@@ -68,9 +68,9 @@ func init() {
 			pflag.Lookup(f.Name).Hidden = true
 		}
 	})
-	// Set glog default to stderr
 	if err := flag.Set("logtostderr", "true"); err != nil {
-		glog.Fatal(err)
+		fmt.Printf("can't set log to stderr %+v", err)
+		os.Exit(1)
 	}
 
 	paths = environment.MustGetKrewPaths()
--- a/cmd/krew/cmd/root.go
+++ b/cmd/krew/cmd/root.go
@@ -62,13 +62,16 @@ func Execute() {
 
 func init() {
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
-	flag.CommandLine.Parse([]string{}) // convince pkg/flag we parsed the flags
+	_ = flag.CommandLine.Parse([]string{}) // convince pkg/flag we parsed the flags
 	flag.CommandLine.VisitAll(func(f *flag.Flag) {
 		if f.Name != "v" { // hide all glog flags except for -v
 			pflag.Lookup(f.Name).Hidden = true
 		}
 	})
-	flag.Set("logtostderr", "true") // Set glog default to stderr
+	// Set glog default to stderr
+	if err := flag.Set("logtostderr", "true"); err != nil {
+		glog.Fatal(err)
+	}
 
 	paths = environment.MustGetKrewPaths()
 	if err := ensureDirs(paths.BasePath(),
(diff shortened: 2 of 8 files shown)


ferhatelmas:
I had added `fixes` keyword but it seems not allowed so changed it to `related to`.

cmd/validate-krew-manifest/main.go
```diff
@@ -40,7 +40,10 @@ var flManifest string
 
 func init() {
 	flag.StringVar(&flManifest, "manifest", "", "path to plugin manifest file")
-	flag.Set("logtostderr", "true") // Set glog default to stderr
+	// Set glog default to stderr
+	if err := flag.Set("logtostderr", "true"); err != nil {
+		glog.Fatal(err)
```
ahmetb: I think it would be good to use fmt.Println + os.Exit here, as glog is currently not initialized probably.

cmd/krew/cmd/root.go
```diff
@@ -62,13 +62,16 @@ func Execute() {
...
-	flag.CommandLine.Parse([]string{}) // convince pkg/flag we parsed the flags
+	_ = flag.CommandLine.Parse([]string{}) // convince pkg/flag we parsed the flags
 	flag.CommandLine.VisitAll(func(f *flag.Flag) {
 		if f.Name != "v" { // hide all glog flags except for -v
 			pflag.Lookup(f.Name).Hidden = true
 		}
 	})
-	flag.Set("logtostderr", "true") // Set glog default to stderr
+	// Set glog default to stderr
+	if err := flag.Set("logtostderr", "true"); err != nil {
```
ahmetb: glog might not be initialized yet.
maybe use fmt.Println+os.Exit(1)

corneliusweig:
Looks good to me. Thanks for getting this done!

Have you tried changing the PR description from "related to #325" to "fixes #325"? I think it's only forbidden in commit messages but fine in PR descriptions.

.errcheck-excluded.txt
```diff
@@ -0,0 +1,2 @@
+os.MkdirAll
```
corneliusweig: Would it also be possible to explicitly ignore errors in the exact locations where we want to ignore errors. Like so:
```go
_ = os.MkdirAll(...)
```

ferhatelmas: definitely. Then we shouldn't have this file.

ferhatelmas:
> I think it's only forbidden in commit messages but fine in PR descriptions.

@corneliusweig updated and seems so.

.errcheck-excluded.txt
ferhatelmas: yes, it's possible. For being explicit, I dropped this file.

ahmetb: since it's just 2 lines or possibly even 1, let's just drop this file if possible and call errcheck with --flags. it's kinda annoying to have an extra file in repo root just for this.

ferhatelmas: yes, dropped. Will push soon.

ferhatelmas:
> Looking at this a little more, I do not understand why some errors are not reported. For example, in pkg/receiptsmigration/migration_test.go:66 there is a Setenv call which returns an unhandled error that is not reported. On the other hand, the unhandled MkdirAll error in line 72 is reported. So what is the logic behind reporting some but not all errors?

It seems a bug from golangci-lint. Even if errcheck reports it, golangci-lint doesn't report it. Reported https://github.com/golangci/golangci-lint/issues/657

cmd/validate-krew-manifest/main.go
ferhatelmas: updated as suggested.

cmd/krew/cmd/root.go
ferhatelmas: updated as suggested.

.errcheck-excluded.txt
ferhatelmas: pushed.

corneliusweig:
/lgtm
waiting for @ahmetb to approve.

PR #331, opened 2019-09-03
lint: integrate shfmt and reformat

### LINKED ISSUE #329: Consider enabling a shell formatter
It seems more and more scripts are being added for different tasks.

Similar to having consistency in go, a formatter for shell could ensure it in scripts such as https://github.com/mvdan/sh

/kind cleanup
/kind proposal

ferhatelmas:
* integrate `mvdan/shfmt` to have consistent style in shell scripts
* it will be running part of the lint
* inconsistent style is updated by running it

Fixes #329


ferhatelmas:
It seems failure is [manifest validation](https://travis-ci.org/kubernetes-sigs/krew/builds/580440194#L396) and irrelevant to this PR.

ahmetb:
The failure seems irrelevant to this PR but still relevant.
Looks like sometimes we generate krew version with v0.0.0-0*. 

That last zero is causing the problem. Do you mind Changing it to: v0.0.0-detached.$commit perhaps? It should be in the version stamping section of the build script.

ferhatelmas:
> Do you mind Changing it to: v0.0.0-detached.$commit perhaps? 
@ahmetb in this PR ?

ahmetb:
Yeah sure

ferhatelmas:
OK, updated. I hope this is what you mean.

ahmetb:
yeah looks good.

hack/format-scripts.sh
```diff
@@ -0,0 +1,30 @@
...
+gopath="$(go env GOPATH)"
+
+# install shfmt that ensures consistent formal in scripts
+if ! [[ -x "${gopath}/bin/shfmt" ]]; then
+  echo >&2 'Installing shfmt'
+  curl --silent --fail --location https://github.com/mvdan/sh/releases/download/v2.6.4/shfmt_v2.6.4_linux_amd64 >"${gopath}/bin/shfmt"
+  chmod u+x "${gopath}/bin/shfmt"
+fi
+
+"$gopath/bin/shfmt" -w -i=2 hack
```
ahmetb: use SCRIPT_DIR and compute where exactly hack/ is (see other scripts).
if I call this after `cd hack` it won't work.

ahmetb: I was thinking this PR would also have a `hack/validate-scripts-formatting.sh` that checks if this command prints any diff and if so it would print the diff to have the committer fix the errors by telling user to run `shfmt -w -i=2 hack/`. 

Should we do that in another PR?
If we don't fail the build, people won't realize they should be running this script.

ferhatelmas: I agree so I added into lint 
https://github.com/kubernetes-sigs/krew/pull/331/files#diff-c8cc35391f3940f8aa5f58aed0516ca1R57

However, I can rename this file (that just formats all files) as your suggestion (`validate-script-formatting.sh` - note I dropped `s`, `script` instead of `scripts`)

hack/format-scripts.sh
```diff
@@ -0,0 +1,30 @@
...
+set -euo pipefail
+
+[[ -n "${DEBUG:-}" ]] && set -x
+
+gopath="$(go env GOPATH)"
+
+# install shfmt that ensures consistent formal in scripts
+if ! [[ -x "${gopath}/bin/shfmt" ]]; then
+  echo >&2 'Installing shfmt'
+  curl --silent --fail --location https://github.com/mvdan/sh/releases/download/v2.6.4/shfmt_v2.6.4_linux_amd64 >"${gopath}/bin/shfmt"
```
corneliusweig: can you `go get` instead? Not everybody is working on a linux machine.

hack/run-lint.sh
```diff
@@ -20,31 +20,38 @@ set -euo pipefail
...
+  echo >&2 'Installing impi'
+  install_impi
 fi
 
 "$gopath/bin/impi" --local sigs.k8s.io/krew --scheme stdThirdPartyLocal ./...
+
+# install shfmt that ensures consistent format in shell scripts
+if ! [[ -x "${gopath}/bin/shfmt" ]]; then
+  echo >&2 'Installing shfmt'
+  curl --silent --fail --location https://github.com/mvdan/sh/releases/download/v2.6.4/shfmt_v2.6.4_linux_amd64 >"${gopath}/bin/shfmt"
```
corneliusweig: `go get` as above would be better IMHO

hack/format-scripts.sh
ferhatelmas: done

but how about go part in lint, it would fail too right now.

hack/run-lint.sh
ferhatelmas: done

hack/format-scripts.sh
ferhatelmas: done

hack/run-lint.sh
ferhatelmas: do you think if we should pull a specific version?

hack/format-scripts.sh
```diff
@@ -20,11 +20,20 @@ set -euo pipefail
 
 gopath="$(go env GOPATH)"
 
+install_shfmt() {
```
corneliusweig: Idea for the future: we could extract this little helper. I think that same snippet appears in about 4 different locations by now.

hack/run-lint.sh
corneliusweig: Let's try with installing from master. If it ever makes problems, let's pin the version.

ferhatelmas: good then.

hack/format-scripts.sh
ferhatelmas: This one specifically is in 2 places but there are a few more. Yes, I agree.

ferhatelmas: To track, created https://github.com/kubernetes-sigs/krew/issues/332

hack/format-scripts.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
```
corneliusweig: Picking up @ahmetb's comment from https://github.com/kubernetes-sigs/krew/issues/332#issuecomment-528600895, we should probably remove this reformat script. For example, there is also no script for `goimports -w` or similar. Instead, I recommend documenting the shell linter in `docs/CONTRIBUTOR_GUIDE.md`.

ahmetb: Agreed.

ahmetb: Also The error message in the validation script should tell the committer what command to run to fix, and where the errors were found (or a diff).

ferhatelmas: Done, I guess. Please, give another look. Thanks!

docs/CONTRIBUTOR_GUIDE.md
```diff
@@ -34,6 +34,19 @@ and run:
...
+
+```bash
+hach/run-lint.sh
+```
+
+If there is a divergence, diff will be provided. Otherwise, no output.
+Diff can be fixed by running:
+
+```bash
+shfmt -w -i=2 **/*.sh
```
ahmetb: I have a suspicion that `**` does nothing on bash 3.x (requires `shopt -s globstar`), it's fine on bash 4.x. Also, would be good to limit this to only `hack/`?

ahmetb: Yeah let's do this `hack/` because the validating script (`run-lint.sh`) is also just running on hack/.

hack/run-lint.sh
```diff
@@ -20,31 +20,53 @@ set -euo pipefail
...
+  install_shfmt
+fi
+
+SCRIPTDIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+shfmt_help() {
+  echo "To fix run: shfmt -w -i=2 ${SCRIPTDIR}"
+}
+
+trap 'shfmt_help' ERR
+"$gopath/bin/shfmt" -d -i=2 "${SCRIPTDIR}"
```
ahmetb: @ferhatelmas  let's do the following instead because (1) trap here is unnecessary (2) diff can be pretty large:

```sh
shfmt_out="$("$gopath/bin/shfmt" -l -i=2 "${SCRIPTDIR}")
if [[ -n "${shfmtout}" ]]; fi
  echo >&2 "The following shell scripts need to be formatted, run: 'shfmt -w -i=2 ${SCRIPTDIR}'
  echo >&2 "${shfmt_out}"
  exit 1
fi
```

`shfmt -l` just lists the files, and exits with code=0, so it's easier.

ferhatelmas: LGTM, will update.

But trap was needed if diff was printed because I wanted to show run command at the end after diff (exiting).

ferhatelmas: Updated. 

However, after second thought, since we require lint, diff in the previous implementation should have been very small incrementally.

ahmetb: Valid point, although the diffs can be harder to navigate especially when a new script is added.

PR #335, opened 2019-09-09
lint: simplify scripts while formatting

ferhatelmas:
simplify shell scripts while formatting similar to `gofmt -s`.

### DIFF
--- a/docs/CONTRIBUTOR_GUIDE.md
+++ b/docs/CONTRIBUTOR_GUIDE.md
@@ -45,7 +45,7 @@ Otherwise, problematic scripts will be listed and a command will be
 provided to format them:
 
 ```bash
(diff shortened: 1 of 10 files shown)


ahmetb:
Let's not do this. It intervenes with the style quite a bit.

ahmetb:
All it seems to do is to just unquote variables when they're used inside `[` or `[[`. There's no harm in using double quotes for strings in `[` and `[[` (numbers are a different story).

ferhatelmas:
> Let's not do this. It intervenes with the style quite a bit.

I thought creating an issue to discuss at first but then did since it was tiny.

Closing then?

ahmetb:
Yeah, the value add is very small.
/close

Issue #338, opened 2019-09-11
Stop releasing krew as .zip


ahmetb:
History here: Initially, krew only had .zip support for archives, but we added
.tar.gz in v0.2. This is why we're distributing krew itself (as a plugin) in
both extensions.

Currently we have a single file containing all release binaries in both
archieve formats (we might consider splitting in the future), however right now
there's no need for us to have both (.tar.gz and .zip).

Especially for analytics (download counts tracking via GitHub releases API),
having both files is just making it harder.

For v0.3.0, I'm hoping to get rid of .zip format. The only **downside** I can
think of: If a windows user needs the krew.exe, they need to install an extra
tool to extract it out of .tar.gz file.

Thoughts?

/kind proposal
/kind cleanup
/assign

ahmetb:
Actually it seems like since krew needs to be installed via krew for the first time, windows users will need either a `krew.exe` or `krew.zip`. Otherwise, they have to download something extra like 7-ZIP to be able to get the krew.exe out.

What if we distributed these:
- krew.tar.gz{,.sha256}
- krew.exe{,.sha256}

corneliusweig:
Sounds good to me.

ahmetb:
Fixed with #339.

PR #339, opened 2019-09-11
Remove krew.zip from release, add krew.exe

ahmetb:
This patch removes krew.zip (previously used in krew.yaml as well as
windows installation instructions) and adds:

- krew.exe (to be used in first-time installation on windows)
- krew.exe.sha256 (checksum file)

ref: #338 
/hold

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,10 +34,10 @@ deploy:
   on:
     tags: true
   file:
+  - out/krew.yaml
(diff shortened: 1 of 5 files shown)


corneliusweig:
/lgtm
/approve
Maybe it's worth thinking about releasing the `exe` file zipped. Brings a 12MB binary down to 4.something.

PR #349, opened 2019-09-29
Upgrade golangci-lint to v1.19.1

ferhatelmas:
Nothing serious, internal cleanup but some dep upgrades.


corneliusweig:
Other projects seem to have trouble with excessive memory usage in this version. But I think krew is small enough so that it doesn't matter :)
Thanks for contributing this!
/lgtm
/approve

Issue #193, opened 2019-05-30
Allow installation from remote manifests without downloading the bundle


corneliusweig:
I find it annoying that I need to download the file bundle _and_ the manifest when testing a `krew-index` PR. It could be much simpler, if `krew` would accept remote URIs for the manifest location and if krew would download the file bundle referenced in there. Is there a particular reason why `krew` must not do this?

Thinking further, `krew` should keep track of the manifest URIs if not coming from the krew index. On `krew update`, it should then also download new versions from those URIs and handle those just as any other plugin. This would then allow plugin authors to maintain their own krew manifest within the plugin repo with all the benefits of krew (easy install, updates).

WDYT?

ahmetb:
You can easily do

    --manifest <(curl url) --archive <(curl url)

We discussed this in early days and decided not to add more magic and rely on unix conventions. Maybe it’s worth documenting though.

corneliusweig:
Well I would argue that the cmd-line hack is less practical. After all, you need to copy/paste two different urls, not just one:
```
--manifest <(curl manifest-url) --archive <(curl archive-url)
```

BTW, what do you think about keeping track of the manifest URLs so that they can be updated during `krew update`?

ahmetb:
To test manifests in index repo, I have the refined-github extension, so I can click Copy to copy manifest file contents into clipboard, then I have this alias:

    krew_install_clipboard="kubectl krew install --manifest <(pbpaste)"

so I just run `krew_install_clipboard`. That's just my personal workflow.


> BTW, what do you think about the idea of keeping track of the manifest URLs so that they can be updated during `krew update`?

We shouldn't go down this path. One-off installations are meant only for dev/test. We provide the --manifest/--archive just to ease local development experience. Once we start supporting multiple indexes, I don't think there will be a real need for non-developers to use --manifest/--archive.

corneliusweig:
But that's exactly the point: if the location of plugins is recorded, they are no longer one-off. In a way, this is the multiple indexes approach in the extreme, where every plugin manifest is maintained in its own repository.
In contrast to multiple indexes, this would allow plugin authors to benefit from easy krew installation (+updates) without the need to add the plugin to an index.

ahmetb:
The only problem I see with that is it ends up creating two paths for obtaining manifests. One being a local git repo that’s refreshed directly/indirectly via `krew update`; another one that comes from a permalink/URL directly.

[...]

PR #350, opened 2019-10-03
Adds --manifest-url flag to allow plugin installation from remote manifests

rewanth1997:
This feature addition lets the user/developer to install a krew plugin directly from the source URL instead of manually downloading the plugin file.

## Problem statement
Install kubectl plugin from config file

### Current approach
```
wget https://raw.githubusercontent.com/rewanth1997/kubectl-fields/master/deploy/krew/plugin.yaml
kubectl krew install --manifest=./plugin.yaml
```

### Proposed approach
```
kubectl krew install --manifest-url="https://github.com/rewanth1997/kubectl-fields/raw/master/deploy/krew/plugin.yaml"
```

This new feature addition will download the plugin from the URL to a temporary location and then installs the plugin from that temporary fie using the regular `manifest` method.

Please review the feature request. I'm happy to make the necessary changes if any.

Regards,
Rewanth


ahmetb:
This proposal was previously discussed & denied, I believe.

rewanth1997:
Oh, I couldn't find any such discussion on Github. 

This PR downloads the yaml file from the URL and stores it in a temporary location. After download, it performs the same operation as the `--manifest` flag. After installation, this feature cleans up the file from the system.

I couldn't think of any potential drawbacks. In fact, I see this as an advantage for developers/users. May I know the reason behind denying this idea?

ahmetb:
I also tried to find but couldn't. Now I'm thinking if it was a private discussion with @corneliusweig.

The main reasoning was a about a few points:

1. You can just do `--manifest <(curl url)` and compose tools/pipes/sub-shells as one should do with UNIX command-line tools.

2. `--manifest-url` is yet another thing to maintain, and makes the user/test surface larger. We thought about detecting url in `--manifest` but that is also "magic", which is not a good thing to do in tools.

Since this has came up multiple times by now, we might be open to it.

/unassign @soltysh 
/assign
/assign @corneliusweig

corneliusweig:
Hi @rewanth1997, the issue @ahmetb mentioned is #193.

---

I think the main reason against the `--manifest-url` option is that this is only intended for plugin devs and not for krew users. Adding this `--manifest-url` option would also encourage end-users to install dangling plugins via this option. However, we don't want this, because `krew` should also take care of updates and this doesn't work when installing from URL.

However, even plugin devs do stumble over this `--manifest <(curl URL)` thing. How about this: if a user tries to install with `--manifest URL`, we could detect if the argument is in fact a URL and print an error suggesting to try `--manifest <(curl URL)` instead?

rewanth1997:
I wasn't aware of the `update` issue.

I just check the `update` module in krew, https://github.com/kubernetes-sigs/krew/blob/master/pkg/gitutil/git.go#L59. With this implementation, the plugins installed via `kubectl install --manifest <(curl ...)` won't be updated.

The idea of throwing a custom error to use curl seems good but building a generic regex expression for URL is very tough. Limiting this just to `http/https` protocol check isn't a good idea either.

corneliusweig:
@rewanth1997 Implementing the URL parser is not necessary. Go already has that built in. See https://github.com/kubernetes-sigs/krew/pull/302/files
```go
if _, err := url.ParseRequestURI(word); err == nil {
   return true
}
```

rewanth1997:
I don't think this works properly,

`url.ParseRequestURI` considers anything starting with "/" as URL. It might be an absolute path to the plugin file in the host machine.

"/home/rewanth/kubectl-fields/deploy/krew/plugin.yaml" gets detected as URL.

```
package main

import (
    "fmt"
    "net/url"
)

func main() {
    fmt.Println(isUrl("https://github.com/kubernetes-sigs/krew/pull/350"))
    fmt.Println(isUrl("/home/rewanth/kubectl-fields/deploy/krew/plugin.yaml"))
    fmt.Println(isUrl("plugin.yaml"))
}

func isUrl(toTest string) bool {
    _, err := url.ParseRequestURI(toTest)
    if err != nil {
        return false
    } else {
        return true
    }
}
```

Output
```
true
true
false
```

rewanth1997:
```
rewanth@ubuntu:/tmp/krew/out/bin$ ./krew-linux_amd64 install --manifest=https://raw.githubusercontent.com/rewanth1997/kubectl-fields/master/deploy/krew/plugin.yaml
F1013 22:59:14.313896  107113 root.go:58] Cannot install plugin directly from URL
To install from URL, use --manifest <(curl https://raw.githubusercontent.com/rewanth1997/kubectl-fields/master/deploy/krew/plugin.yaml)
```

@corneliusweig I added an additional check to check for `Schema` and `Host` to validate the URL. Please review the PR.

corneliusweig:
Good find! I'll consider adding the same logic in the other PR.

Can you also add some tests for this behavior? Otherwise this looks good. Thanks!

rewanth1997:
I validated the module independently with different protocols like SMB, FTP, HTTP, etc. I believe these checks are good enough for integration. Please let me know if something else needs to be done.

List of tests performed: [https://play.golang.org/p/xUN0SHp1c9t](https://play.golang.org/p/xUN0SHp1c9t)

ahmetb:
@corneliusweig have we settled on --manifest vs --manifest-url?
It seems like latter is more verbose (as opposed to _magic_) and can be easily implemented, albeit with extra testing requirements.

corneliusweig:
@ahmetb The scope of this PR was changed. It now only prints an error message with a hint to use `--manifest <(curl filename)`, if it detects that `filename` is a remote URL.

corneliusweig:
/retitle Print hint when user tries to install via manifest from remote URL

corneliusweig:
@rewanth1997 You could add an e2e test for install with manifest which does one of the checks automatically (e.g. `https://github.com/kubernetes-sigs/krew/pull/350`). A good place would be in `integration_test/install_test.go`.

rewanth1997:
Hi @corneliusweig @ahmetb ,

I have a concern about the current idea. You mentioned in the above [comment-539180993](https://github.com/kubernetes-sigs/krew/pull/350#issuecomment-539180993) that `--manifest-url` has to be avoided because it installs dangling plugins and `krew update` is not applied on them. In that case, even the plugins installed by `--manifest` argument cannot be updated by `krew update`.

If `--manifest` is allowed to install dangling plugins, I think `--manifest-url` should also be allowed.

**Proposed solution:**
We will revert back to my original commit of adding `--manifest-url` and add a warning like *"Plugins installed via --manifest/--manifest-url are not updated by krew"* when the users install plugin with `--manifest/--manifest-url`

What are your views on this new proposal?

ahmetb:
Yeah I think --manifest-url is not that bad.
We just need to write tests for it (for every case we've been testing with --manifest).

No need to add a warning. The flag description clearly says `(Development-only)`.

rewanth1997:
I already implemented the code for `--manifest-url`. I can add the test cases too for this module.

Once we get @corneliusweig views on adding `--manifest-url`, I will revert back to first commit and will add the test cases as well.

corneliusweig:
@rewanth1997 Alright then, let's have the `--manifest-url` flag. Sorry if I caused additional work with this back-and-forth. Please also change back the PR title to something meaningful.

rewanth1997:
Not a problem. I will make the required changes.

rewanth1997:
/retitle Adds --manifest-url option to allow installation from remote manifests

@corneliusweig Please review the latest code which includes integration tests as well

rewanth1997:
@corneliusweig Your kconfig repo (https://github.com/corneliusweig/konfig/) isn't having a valid plugin structured yaml file which can be passed to `--manifest-url` option for running the integration tests.

Hence, I added my plugin file (kubectl-fields) to run the integration tests.

cmd/krew/cmd/install.go
```diff
@@ -62,6 +99,20 @@ Remarks:
 			var pluginNames = make([]string, len(args))
 			copy(pluginNames, args)
 
+			// Downloads manifest file from given URL
+			if *manifestURL != "" {
+				fileName, err := downloadFile(*manifestURL)
+				if err != nil {
+					fmt.Fprintln(os.Stderr, "Error downloading file")
```
corneliusweig: Please do
```go
return errors.Wrapf(err, "Error downloading manifest from %q", *manifestURL)
```

Just logging an error won't be enough ;)

---

Please also put some thought into the question whether the cleanup function (remove the temp-file) should be called when an error is returned here. Otherwise we might leak temporary files.

cmd/krew/cmd/install.go
```diff
@@ -30,10 +32,45 @@ import (
 	"sigs.k8s.io/krew/pkg/installation"
 )
 
+// Downloads data from url to a filepath location
+func downloadFile(url string) (string, error) {
```
corneliusweig: Can you move that function into the `internal` package?

cmd/krew/cmd/install.go
```diff
@@ -62,6 +99,20 @@ Remarks:
 			var pluginNames = make([]string, len(args))
 			copy(pluginNames, args)
 
+			// Downloads manifest file from given URL
+			if *manifestURL != "" {
```
corneliusweig: I wonder if you could find a way to not make the `install` function even longer. It already is quite long and is very hard to test. Adding this here only makes it worse. So if you find a way to move this code into some helper where it can be tested independently, that would be great!

pkg/constants/constants.go
```diff
@@ -19,6 +19,7 @@ const (
 	PluginKind        = "Plugin"
 	ManifestExtension = ".yaml"
 	KrewPluginName    = "krew" // plugin name of krew itself
+	ManifestURL       = "https://github.com/rewanth1997/kubectl-fields/raw/master/deploy/krew/plugin.yaml"
```
corneliusweig: These constants are for production code only. I think `testutil_test.go` would be a better place for this.

integration_test/install_test.go
```diff
@@ -105,6 +105,17 @@ func TestKrewInstall_Manifest(t *testing.T) {
 	test.AssertExecutableInPATH("kubectl-" + validPlugin)
 }
 
+func TestKrewInstall_ManifestURL(t *testing.T) {
```
corneliusweig: Maybe also add a test-case where `--manifest-url` _and_ `--archive` are given? What should happen in that case?

ahmetb: I think we can make these flags mutually exclusive and explicitly error out. People most probably won't even use it.

That said, --archive just always uses the provided file instead of downloading the .zip file from the internet. So it's actually still valid here, but that's not why people are going to use --manifest-url for. 

So yeah, I recommend 1) explicitly fail in this case 2) add integration tests to test that.

cmd/krew/cmd/install.go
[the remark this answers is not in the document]
rewanth1997: Done

cmd/krew/cmd/install.go
```diff
@@ -30,10 +32,45 @@ import (
...
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("bad status: %s", resp.Status)
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
```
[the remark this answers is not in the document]
rewanth1997: Done

cmd/krew/cmd/install.go
rewanth1997: Done. I made a call to cleanup function to clear the temp files.

cmd/krew/cmd/install.go
rewanth1997: Done

cmd/krew/cmd/install.go
rewanth1997: Unfortunately no. We cannot shift this piece of code to any helper function because we use `defer` to cleanup the temporary files after installation here (https://github.com/rewanth1997/krew/blob/manifest-url/cmd/krew/cmd/install.go#L74)

If we move the piece of code to another function, `defer` will work locally to the particular helper function only and will loose its primary purpose.

pkg/constants/constants.go
rewanth1997: Done

integration_test/install_test.go
rewanth1997: Done

rewanth1997:
Hi @corneliusweig ,

Apologies for the delay. I made all the recommended changes, please review the latest code.

```
Do you think you could set up a local test server with import "net/http/httptest" to serve the manifest for tests?
```
I didn't understand this question. Can you elaborate more?

rewanth1997:
Sorry for the delay @corneliusweig @ahmetb 
I was busy working on a client's project and didn't get time to fix this PR. Since I have free time now, I will be more active and resolve the issues quickly.

cmd/krew/cmd/install.go
rewanth1997: I tried to make the code as minimal as possible in this section. Its only 5 lines of code.
```
downloadFile()
cleanup()
check_err
assign temporary file name to the manifest variable
```

ahmetb: Sorry I'm jumping on this but why do we need to write the temporary file onto the disk? 

Maybe if we changed the installation parameter with an `io.ReadCloser` we can solve this more cleanly?

corneliusweig:
@rewanth1997 What I mean is that you don't fetch the manifest from some real repo, but instead you create an http server that serves the manifest and the archive. This has two advantages:

- When working on krew, we have full control over the manifest and archive. If the test breaks, we can fix it.
- There will be a much smaller delay when "downloading", because the files are served locally. So the tests will run faster.

I think the `"net/http/httptest"` package has the required utils to do this.

ahmetb:
Some examples with httptest pkg:
- https://itnext.io/how-to-stub-requests-to-remote-hosts-with-go-6c2c1db32bf2
- https://medium.com/what-i-talk-about-when-i-talk-about-technology/go-code-examples-httptest-newserver-f965fb349884
- https://medium.com/@xoen/go-testing-technique-testing-json-http-requests-76d9ce0e11f

cmd/krew/cmd/install.go
rewanth1997: My initial idea was to the downloaded data to a temporary file on the disk and then to pass the temporary file path as a value to `--manifest` variable.

Thanks for pointing the `ReadCloser` interface. This will definitely make the code look cleaner. I can directly save the downloaded file in a variable/buffer and pass it to respective variables rather than storing on disk and removing it later. I will make the changes asap.

rewanth1997:
@ahmetb @corneliusweig  I stored the downloaded file in the buffer. Please share your views on the latest implementation.

I will add test cases getting served via local server asap.

rewanth1997:
@ahmetb @corneliusweig All the recommended changes have been done. Please review the code that serves files from the local test server.

cmd/krew/cmd/install.go
```diff
@@ -94,6 +102,21 @@ Remarks:
...
+					return errors.Wrapf(err, "Error downloading manifest from %q", *manifestURL)
+				}
+				plugin, err := indexscanner.DecodePluginFile(resp)
+				if err != nil {
+					return errors.Wrap(err, "failed to load custom manifest file")
+				}
+				if err := validation.ValidatePlugin(plugin.Name, plugin); err != nil {
+					return errors.Wrap(err, "plugin manifest validation error")
+				}
+				install = append(install, plugin)
```
ahmetb: this is too repetitive with the next block, it's not good for code cleanliness.

rewanth1997: True. 

he only alternative is to define `plugin` and `err` as global variables and then pass them to the repetitive block of code. I thought it would be a bad idea to declare variables with global scope and I preferred the repetitive method.

I can implement the alternative method as explained above if that makes sense.

ahmetb: hmm, why don't we make `decodepluginfile` work on io.ReadCloser?

Both resp.Body and os.OpenFile return io.ReadCloser.

rewanth1997:
```
var plugin index.Plugin
var gerr error

if *manifestURL != "" {
	resp, err := internal.DownloadFile(*manifestURL)
	if err != nil {
		return errors.Wrapf(err, "Error downloading manifest from %q", *manifestURL)
	}
	plugin, gerr = indexscanner.DecodePluginFile(resp)
} else if *manifest != "" {
	plugin, gerr = indexscanner.ReadPluginFile(*manifest)
}

if gerr != nil {
	return errors.Wrap(gerr, "failed to load custom manifest file")
}
if err := validation.ValidatePlugin(plugin.Name, plugin); err != nil {
	return errors.Wrap(err, "plugin manifest validation error")
}
install = append(install, plugin)
```

How does this look? This snippet will replace this piece of code, https://github.com/kubernetes-sigs/krew/blob/6df65f26ed4b6420ad8ed0678ddb4cfe0a0ad99e/cmd/krew/cmd/install.go#L105-L129

integration_test/testdata/konfig_localhost.yaml
```diff
@@ -0,0 +1,28 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
```
ahmetb: why do we need this file?
As far as I can tell you don't need to fetch the plugin from a hardcoded port number.
All you care about while implementing `--manifest-url` is that manifest is read from URL and then you can pass `--archive` override.

you put `uri: http://ignored`

integration_test/testutil_test.go
```diff
@@ -37,8 +37,10 @@ import (
 const (
 	persistentIndexCache = "krew-persistent-index-cache"
 	krewBinaryEnv        = "KREW_BINARY"
-	validPlugin          = "konfig" // a plugin in central index with small size
-	validPlugin2         = "mtail"  // a plugin in central index with small size
+	validPlugin          = "konfig"          // a plugin in central index with small size
+	validPlugin2         = "mtail"           // a plugin in central index with small size
+	LocalhostURL         = "127.0.0.1:60564" // make sure this port is open. incase of editing port number, make sure to update the URL in testdata/konfig_localhost.yaml
```
ahmetb: you shouldn't need this. just simplify by reading tarball from `--archive` in integration test?

integration_test/install_test.go
```diff
@@ -105,6 +111,95 @@ func TestKrewInstall_Manifest(t *testing.T) {
...
+		t.Errorf("Expected install to fail but was successful")
+	}
+}
+
+func TestKrewInstall_ManifestURLAndManifest(t *testing.T) {
+	skipShort(t)
+
+	server, err := startServer()
+	if err != nil {
+		t.Errorf("Trouble starting local server")
```
ahmetb: you aren't logging what the error was. just t.Fatalf(error)

Also errorf doesn't stop the test. so the test will continue to execute.

integration_test/install_test.go
```diff
@@ -105,6 +111,95 @@ func TestKrewInstall_Manifest(t *testing.T) {
...
+	defer server.Close()
+
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	err = test.Krew("install",
+		"--manifest-url", LocalhostManifestURL,
+		"--manifest", filepath.Join("testdata", fooPlugin+constants.ManifestExtension)).Run()
+	if err == nil {
+		t.Errorf("Expected install to fail but was successful")
```
ahmetb: lowercase errors

integration_test/install_test.go
```diff
@@ -105,6 +111,95 @@ func TestKrewInstall_Manifest(t *testing.T) {
 	test.AssertExecutableInPATH("kubectl-" + validPlugin)
 }
 
+func startServer() (*httptest.Server, error) {
+	server := httptest.NewUnstartedServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.RequestURI {
+		case "/konfig_localhost.yaml":
```
ahmetb: why don't you just use http.FileServer ?

cmd/krew/cmd/install.go
ahmetb: good enough but instead of creating `gerr`, you should probably wrap/return errors immediately where they occur.

Ideally I'd love to see us using `indexscanner.Decode` just accept `io.ReadCloser`. Maybe we can do it later, not in this PR.

ahmetb:
Currently this PR is too big. Ideally you should add 1 if statement + 1 unit test to cover that the parsing happens. Nothing else should be necessary as far as I can tell, overall please try to reduce complexity. We shouldn't need ~200 lines of code for a simple flag.

integration_test/install_test.go
rewanth1997: I thought having a custom handler function will be helpful if we need to serve additional files in the future from different directories.

integration_test/testdata/konfig_localhost.yaml
rewanth1997: I added a testcase in which `--archive` is not specified along with `--manifest-url`. How can we ignore the URL in that case? We would need a valid URL to download the plugin.

integration_test/install_test.go
rewanth1997: Done

integration_test/install_test.go
ahmetb: But we currently don’t, and this implementations is 20x longer. :)

integration_test/install_test.go
rewanth1997: Done

integration_test/install_test.go
rewanth1997: I will fix it

cmd/krew/cmd/install.go
rewanth1997: Done

integration_test/install_test.go
rewanth1997: Done

integration_test/testdata/konfig_localhost.yaml
rewanth1997: Done

integration_test/testutil_test.go
rewanth1997: As we discussed previously `--archive` cannot be used with `--manifest-url`. Hence, I added this.

rewanth1997:
I agree that this PR is a bit lengthy.

The recommended changes are made and the size of this PR is notably reduced. I added extra unit tests with a comination of `--manifest-url` and other flags like `--archive` and `--manifest`.

This PR has 1 if statement, a file that contains code to download files from URL, 3 unit test cases and the setup of a local test server. I think ~150 lines of code is a reasonable length to code all the above operations.

@ahmetb @corneliusweig Please review the latest code.

integration_test/install_test.go
```diff
@@ -105,6 +110,75 @@ func TestKrewInstall_Manifest(t *testing.T) {
...
+	server.Listener = listener
+	server.Start()
+	return server, nil
+}
+
+func TestKrewInstall_ManifestURL(t *testing.T) {
+	skipShort(t)
+
+	server, err := startLocalServer()
+	if err != nil {
```
ahmetb: Do this:

Pass t to startLocalServer, Mark that method as t.Helper() and if err starting server, t.Fatal in that method; not here.

integration_test/testutil_test.go
```diff
@@ -37,8 +37,10 @@ import (
 const (
 	persistentIndexCache = "krew-persistent-index-cache"
 	krewBinaryEnv        = "KREW_BINARY"
-	validPlugin          = "konfig" // a plugin in central index with small size
-	validPlugin2         = "mtail"  // a plugin in central index with small size
+	validPlugin          = "konfig"          // a plugin in central index with small size
+	validPlugin2         = "mtail"           // a plugin in central index with small size
+	localServerURL       = "127.0.0.1:45076" // change port number incase of bind error
```
ahmetb: Please do not hardcode. 

httputil can start a server at a random port. 
Just return its base url and close function from test util func.

cmd/krew/cmd/install.go
```diff
@@ -94,11 +102,27 @@ Remarks:
...
+
+				if *manifestURL != "" {
+					resp, err := internal.DownloadFile(*manifestURL)
+					if err != nil {
+						return errors.Wrapf(err, "Error downloading manifest from %q", *manifestURL)
+					}
+					plugin, err = indexscanner.DecodePluginFile(resp)
+					if err != nil {
+						return errors.Wrap(err, "failed to load custom manifest file")
+					}
```
corneliusweig: this should go into `internal.DownloadFile`

cmd/krew/cmd/internal/downloadFile.go
```diff
@@ -0,0 +1,30 @@
...
+package internal
+
+import (
+	"io"
+	"net/http"
+)
+
+// DownloadFile function takes URL as input, stores the contents in temporary file and returns filename, cleanup function to delete temporary files and error
+func DownloadFile(url string) (io.ReadCloser, error) {
+	resp, err := http.Get(url)
```
corneliusweig: `http.Get` returns a response object whose `Body` is expected to be closed by the user. See the documentation for `http.Do` for details:

> // If the returned error is nil, the Response will contain a non-nil
// Body which the user is expected to close. If the Body is not both
// read to EOF and closed, the Client's underlying RoundTripper
// (typically Transport) may not be able to re-use a persistent TCP
// connection to the server for a subsequent "keep-alive" request.

I think it would make more sense to not just fetch any file with this method, but instead move the whole plugin parsing (`indexscanner.DecodePluginFile`) into this function (see the previous comment). This also improves the testability.

Consequently, the method should be called `ReadRemotePluginManifest` or similar.

cmd/krew/cmd/install.go
```diff
@@ -62,6 +62,14 @@ Remarks:
 			var pluginNames = make([]string, len(args))
 			copy(pluginNames, args)
 
+			if *archiveFileOverride != "" && *manifestURL != "" {
```
corneliusweig: What is the rationale behind forbidding this? Almost nobody will ever try this, I think. But for tests, it could be handy to combine `--manifest-url` and `--archive`.

cmd/krew/cmd/install.go
```diff
@@ -62,6 +62,14 @@ Remarks:
...
+				return errors.New("--archive cannot be specified with --manifest-url")
+			}
+
+			if *manifest != "" && *manifestURL != "" {
```
corneliusweig: This on the other hand makes perfect sense to me.

cmd/krew/cmd/install.go
```diff
@@ -94,11 +102,27 @@ Remarks:
...
+					}
+					plugin, err = indexscanner.DecodePluginFile(resp)
+					if err != nil {
+						return errors.Wrap(err, "failed to load custom manifest file")
+					}
 				}
+
 				if err := validation.ValidatePlugin(plugin.Name, plugin); err != nil {
 					return errors.Wrap(err, "plugin manifest validation error")
 				}
```
corneliusweig: This block should be extracted into its own function, so that the abstraction level in this function is consistent with lines 98,99.

cmd/krew/cmd/install.go
rewanth1997: As recommended by @ahmetb here, https://github.com/kubernetes-sigs/krew/pull/350#discussion_r337132131 I added this check. Please advise on how to proceed further.

rewanth1997:
@ahmetb @corneliusweig 

All the tests are successful on my local system. Just to confirm, I forked my current working repository (https://github.com/rewanth1997/krew/tree/manifest-url) into my test account (https://github.com/testinguser883/krew/tree/manifest-url) and the build is successful there also (https://travis-ci.org/testinguser883/krew/builds/622366453). I'm not sure why travis is failing here.

I made all the suggested changes expect `--archive` and `--manifest-url` conflict.

I added the abstraction layer, removed hard-coded port number, closed response object after use and fixed a few linter bugs as well. Please review the latest code.

cmd/krew/cmd/install.go
rewanth1997: Done

cmd/krew/cmd/internal/downloadFile.go
rewanth1997: Done

cmd/krew/cmd/install.go
rewanth1997: Done

integration_test/testutil_test.go
rewanth1997: Done

integration_test/install_test.go
rewanth1997: Done

cmd/krew/cmd/install.go
corneliusweig: I see, sorry for the confusion. Given that Ahmet had the same idea here https://github.com/kubernetes-sigs/krew/pull/350#discussion_r353975232, I think that we should go for the other option and only treat `--manifest` and `--manifest-url` mutually exclusive, but not `--manifest-url` and `--archive`.

corneliusweig:
@rewanth1997 I think you need to rebase your current branch on top of `origin/master`. Here is what you need to do:
```
git remote add krew-origin git@github.com:kubernetes-sigs/krew.git
git fetch krew-origin
git rebase krew-origin/master
```

I guess you will have to resolve a few merge conflicts.

cmd/krew/cmd/internal/read_plugin.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	if manifestURL != "" {
+		plugin, err = readRemotePluginManifest(manifestURL)
+		if err != nil {
+			return index.Plugin{}, err
+		}
+	}
+
+	return plugin, nil
```
corneliusweig: Why not simply
```go
if manifest != "" {
	return indexscanner.ReadPluginFile(manifest)
}
return readRemotePluginManifest(manifestURL)
```

If both are given, `manifest` should take precedence anyway.

cmd/krew/cmd/internal/read_plugin.go
```diff
@@ -0,0 +1,60 @@
...
+	plugin, err := indexscanner.DecodePluginFile(resp.Body)
+	if err != nil {
+		return index.Plugin{}, err
+	}
+
+	return plugin, nil
+}
+
+// GetPlugin is an abstraction layer for manifest and manifest-url handlers and returns plugin object
+func GetPlugin(manifest string, manifestURL string) (index.Plugin, error) {
```
corneliusweig: One of the major points of extracting logic into this function is to add unit tests. Can you add some? It should be rather easy, because you already did the major setup work for the integration tests.
In the unit tests you can also cover some error cases, for example when nothing is found for the given URL.

cmd/krew/cmd/install.go
```diff
@@ -94,11 +102,14 @@ Remarks:
...
-				plugin, err := indexscanner.ReadPluginFile(*manifest)
+			var plugin index.Plugin
+			if *manifest != "" || *manifestURL != "" {
+				var err error
+				plugin, err = internal.GetPlugin(*manifest, *manifestURL)
 				if err != nil {
 					return errors.Wrap(err, "failed to load custom manifest file")
 				}
+
 				if err := validation.ValidatePlugin(plugin.Name, plugin); err != nil {
```
corneliusweig: I would have moved that into `GetPlugin` as well, but I guess it's ok.

rewanth1997: I thought writing these separately will be able to better differentiate between `loading manifest` and `validation of manifest` errors

cmd/krew/cmd/internal/read_plugin.go
rewanth1997: Yeah, true. Both won't be accepted because I added a pre-check to make sure `manifest` and `manifest-url` are mutually exclusive.

rewanth1997: Done

cmd/krew/cmd/internal/read_plugin.go
```diff
@@ -0,0 +1,60 @@
...
+package internal
+
+import (
+	"net/http"
+
+	"sigs.k8s.io/krew/pkg/index"
+	"sigs.k8s.io/krew/pkg/index/indexscanner"
+)
+
+// readRemotePluginManifest function takes URL as input and returns plugin as output
```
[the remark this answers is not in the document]
rewanth1997: Done

cmd/krew/cmd/internal/read_plugin.go
```diff
@@ -0,0 +1,60 @@
...
+func readRemotePluginManifest(url string) (index.Plugin, error) {
+	resp, err := http.Get(url)
+	if err != nil {
+		return index.Plugin{}, err
+	}
+	defer resp.Body.Close()
+
+	plugin, err := indexscanner.DecodePluginFile(resp.Body)
+	if err != nil {
+		return index.Plugin{}, err
```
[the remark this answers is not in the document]
rewanth1997: Done

cmd/krew/cmd/internal/read_plugin.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"sigs.k8s.io/krew/pkg/index"
+	"sigs.k8s.io/krew/pkg/index/indexscanner"
+)
+
+// readRemotePluginManifest function takes URL as input and returns plugin as output
+func readRemotePluginManifest(url string) (index.Plugin, error) {
+	resp, err := http.Get(url)
+	if err != nil {
+		return index.Plugin{}, err
```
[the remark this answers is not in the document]
rewanth1997: Done

cmd/krew/cmd/internal/read_plugin.go
rewanth1997: Added multiple cases with empty file, invalid file, not existing url, etc.

rewanth1997:
@corneliusweig All recommended changes updated.

cmd/krew/cmd/install.go
rewanth1997: Done

rewanth1997:
@corneliusweig Even the mutually exclusive condition is fixed in the latest commit. All tests are passed locally. Please let me know if any further changes are required.

ahmetb:
The tests are failing with a [very unusual error](https://user-images.githubusercontent.com/159209/70401632-6d4a1c00-19e5-11ea-958c-591e791f752c.png).
I'll make another review or push commits directly to your branch, thanks for staying on top of this @rewanth1997.

integration_test/install_test.go
```diff
@@ -105,6 +109,115 @@ func TestKrewInstall_Manifest(t *testing.T) {
...
+
+	server := httptest.NewUnstartedServer(http.FileServer(http.Dir("testdata")))
+
+	listener, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatalf("trouble starting local server")
+	}
+	server.Listener = listener
+	server.Start()
+	return server, server.URL
```
ahmetb: You can probably just do this:

```
httptest.NewServer(..)
return server.URL, server.Close
```

it does everything you want, and then you can embed it to the test because it's 2 lines.

integration_test/install_test.go
```diff
@@ -105,6 +109,115 @@ func TestKrewInstall_Manifest(t *testing.T) {
...
+
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	err := test.Krew("install",
+		"--manifest-url", baseURL+emtpyManifestURI).Run()
+	if err == nil {
+		t.Errorf("expected install to fail but was successful")
+	}
+}
```
ahmetb: what does this really test?
our validation method? aren't those already covered with unit tests?

integration_test/install_test.go
```diff
@@ -105,6 +109,115 @@ func TestKrewInstall_Manifest(t *testing.T) {
...
+
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	err := test.Krew("install",
+		"--manifest-url", baseURL+invalidManifestURI).Run()
+	if err == nil {
+		t.Errorf("expected install to fail but was successful")
+	}
+}
```
ahmetb: what does this really test?
our validation method? aren't those already covered with unit tests?

cmd/krew/cmd/internal/read_plugin.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// GetPlugin is an abstraction layer for manifest and manifest-url handlers and returns plugin object
+func GetPlugin(manifest string, manifestURL string) (index.Plugin, error) {
+	if manifest != "" {
+		return indexscanner.ReadPluginFile(manifest)
+	}
+
+	return readRemotePluginManifest(manifestURL)
+}
```
ahmetb: This isn't really a good way to do things.
We could do this at the call site, now you just embedded the decisionmaking here, and it doesn't have unit tests.

cmd/krew/cmd/internal/read_plugin.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	defer resp.Body.Close()
+
+	plugin, err := indexscanner.DecodePluginFile(resp.Body)
+	if err != nil {
+		return index.Plugin{}, errors.Wrapf(err, "Error decoding manifest from %q", url)
+	}
+
+	return plugin, nil
+}
```
ahmetb: this doesn't have unit tests :)
would be good to unit-test this, rather than trying to test /empty.yaml /invalid.yaml etc.

cmd/krew/cmd/internal/read_plugin.go
```diff
@@ -0,0 +1,49 @@
+// Copyright 2019 The Kubernetes Authors.
```
ahmetb: this cmd/ package does nearly nothing noteworthy that requires creating a new package for. especially because cmd/ directory directly has "tool" names (krew, generate-plugin-list, ...) this doesn't fit here.

ahmetb:
I think this PR has showed us fundamentally how we should approach accepting patches to krew codebase.

To me, minimalism is key, and ideally we should not require ~200 lines of _new_ code to add one flag like this.

I'll be preparing a PR that does the same thing. I think some issues in the PR stems from the fact that our indexscanner API isn't the best it could be. Let's see how much new code it'll end up in.

cmd/krew/cmd/install.go
```diff
@@ -81,8 +85,8 @@ Remarks:
 				return errors.New("must specify either specify either plugin names (via positional arguments or STDIN), or --manifest; not both")
```
ahmetb: you missed this part. this ought to also factor in --manifest-url.

ahmetb:
/hold
In about an hour I prepared a patch that does this: #429 
It only results in ~120 lines of net added code while addressing the underlying technical debt and adding some missing tests.

rewanth1997:
> The tests are failing with a [very unusual error](https://user-images.githubusercontent.com/159209/70401632-6d4a1c00-19e5-11ea-958c-591e791f752c.png).

Yes, even I'm not sure why? This branch is building successfully without any errors
https://travis-ci.org/testinguser883/krew/builds/622405722?utm_source=github_status&utm_medium=notification

It would be great if you can point out why the build is failing here (maybe it's because this branch is not synced with the current master repo?)

ahmetb:
Yeah, most likely you’ll need to rebase. 
On the other hand, we can continue on my PR as we’re getting to ~100 comments on this and we still have some way to go. I think my patch addresses this more cleanly, but thank you for doing the work on this.

rewanth1997:
Yeah, sure. We will work on your PR. I will try to give inputs to the best of my knowledge.

@corneliusweig @ahmetb Thanks for considering my idea, being so patient and reviewing my PR.

corneliusweig:
Hey @rewanth1997, sorry that your PR doesn't make it into krew's codebase. Don't be upset though. The final draft was already much more mature than the initial one, so you probably could learn a lot by doing it. For your future work, you should focus on testing your code. You can learn a lot from looking at @ahmetb's version :)

/close

PR #354, opened 2019-10-23
Add override flag to list to display full output

JJJJJones:
For certain workflows having list display the plugins and their version info is desirable. I added a flag to bypass the `if !isTerminal(os.Stdout) {...}` check that prevents the full output for redirected output or external program calls and an integration test of the flag. The behavior of `kubectl krew list` is unchanged.

Extras
- implemented sort test for lists
- added testing support for MacOS (requires GNU grep and GNU sed)

### DIFF
--- a/cmd/krew/cmd/list.go
+++ b/cmd/krew/cmd/list.go
@@ -69,7 +69,7 @@ Remarks:
 		PreRunE: checkIndex,
 	}
 
-	listCmd.Flags().BoolVarP(overrideFlag, "override", "o", false, "override standard limited terminal output behavior")
+	overrideFlag = listCmd.Flags().BoolP("override", "o", true, "override standard limited terminal output behavior")
 
 	rootCmd.AddCommand(listCmd)
 }
--- a/cmd/krew/cmd/list.go
+++ b/cmd/krew/cmd/list.go
@@ -29,7 +29,7 @@ import (
 )
 
 func init() {
-	var overrideFlag bool
+	var overrideFlag *bool
 
 	// listCmd represents the list command
 	listCmd := &cobra.Command{
@@ -48,7 +48,7 @@ Remarks:
 			}
 
 			// return sorted list of plugin names when piped to other commands or file
-			if !isTerminal(os.Stdout) && !overrideFlag {
+			if !isTerminal(os.Stdout) && !*overrideFlag {
 				var names []string
 				for name := range plugins {
 					names = append(names, name)
@@ -68,7 +68,8 @@ Remarks:
 		},
 		PreRunE: checkIndex,
 	}
-	listCmd.Flags().BoolVarP(&overrideFlag, "override", "o", false, "override standard limited terminal output behavior")
+
+	listCmd.Flags().BoolVarP(overrideFlag, "override", "o", false, "override standard limited terminal output behavior")
 
 	rootCmd.AddCommand(listCmd)
 }
--- a/cmd/krew/cmd/list.go
+++ b/cmd/krew/cmd/list.go
@@ -29,6 +29,8 @@ import (
 )
 
 func init() {
+	var overrideFlag bool
+
 	// listCmd represents the list command
 	listCmd := &cobra.Command{
 		Use:   "list",
@@ -46,7 +48,7 @@ Remarks:
 			}
 
 			// return sorted list of plugin names when piped to other commands or file
-			if !isTerminal(os.Stdout) {
+			if !isTerminal(os.Stdout) && !overrideFlag {
 				var names []string
 				for name := range plugins {
 					names = append(names, name)
@@ -66,6 +68,7 @@ Remarks:
 		},
 		PreRunE: checkIndex,
 	}
+	listCmd.Flags().BoolVarP(&overrideFlag, "override", "o", false, "override standard limited terminal output behavior")
 
 	rootCmd.AddCommand(listCmd)
 }
--- a/cmd/krew/cmd/list.go
+++ b/cmd/krew/cmd/list.go
@@ -29,7 +29,7 @@ import (
 )
 
 func init() {
-	var wideFlag *bool
+	var namesOnlyFlag, wideFlag *bool
 
 	// listCmd represents the list command
 	listCmd := &cobra.Command{
@@ -48,7 +48,7 @@ Remarks:
 			}
 
 			// return sorted list of plugin names when piped to other commands or file
-			if !isTerminal(os.Stdout) && !*wideFlag {
+			if !isTerminal(os.Stdout) && (!*wideFlag || *namesOnlyFlag) {
 				var names []string
 				for name := range plugins {
 					names = append(names, name)
@@ -69,7 +69,8 @@ Remarks:
 		PreRunE: checkIndex,
 	}
 
-	wideFlag = listCmd.Flags().BoolP("wide", "w", true, "override standard limited terminal output behavior")
+	namesOnlyFlag = listCmd.Flags().Bool("names-only", true, "print only the names of plugins")
+	wideFlag = listCmd.Flags().BoolP("wide", "w", true, "print all plugins and their version")
 
 	rootCmd.AddCommand(listCmd)
 }
--- a/cmd/krew/cmd/list.go
+++ b/cmd/krew/cmd/list.go
@@ -29,7 +29,7 @@ import (
 )
 
 func init() {
-	var overrideFlag *bool
+	var wideFlag *bool
 
 	// listCmd represents the list command
 	listCmd := &cobra.Command{
@@ -48,7 +48,7 @@ Remarks:
 			}
 
 			// return sorted list of plugin names when piped to other commands or file
-			if !isTerminal(os.Stdout) && !*overrideFlag {
+			if !isTerminal(os.Stdout) && !*wideFlag {
 				var names []string
 				for name := range plugins {
 					names = append(names, name)
@@ -69,7 +69,7 @@ Remarks:
 		PreRunE: checkIndex,
 	}
 
-	overrideFlag = listCmd.Flags().BoolP("override", "o", true, "override standard limited terminal output behavior")
+	wideFlag = listCmd.Flags().BoolP("wide", "w", true, "override standard limited terminal output behavior")
 
 	rootCmd.AddCommand(listCmd)
 }
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,6 @@ require (
 	github.com/sahilm/fuzzy v0.0.5
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/pflag v1.0.3
-	golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2
 	golang.org/x/net v0.0.0-20190628185345-da137c7871d7 // indirect
 	golang.org/x/sys v0.0.0-20190712062909-fae7ac547cb7 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
--- a/go.mod
+++ b/go.mod
@@ -15,9 +15,11 @@ require (
 	github.com/sahilm/fuzzy v0.0.5
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/pflag v1.0.3
+	golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2
 	golang.org/x/net v0.0.0-20190628185345-da137c7871d7 // indirect
 	golang.org/x/sys v0.0.0-20190712062909-fae7ac547cb7 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
+	gopkg.in/yaml.v2 v2.2.2
 	k8s.io/apimachinery v0.0.0-20190717022731-0bb8574e0887
 	k8s.io/client-go v7.0.0+incompatible
 	k8s.io/klog v0.3.3 // indirect
--- a/hack/verify-code-patterns.sh
+++ b/hack/verify-code-patterns.sh
@@ -16,8 +16,20 @@
 
 set -euo pipefail
 
+# Appropriately set grep
+if [[ $(uname) == "Darwin" ]]; then
+  out="$(command -v ggrep)"
+  if [[ -n $out ]]; then
+    grep="ggrep"
+  else
+    echo >&2 "GNU grep is required for MacOS users to successfully run local tests; it can be found on homebrew as 'grep'"
+  fi
+else
+  grep="grep"
+fi
+
 # Disallow usage of ioutil.TempDir in tests in favor of testutil.
-out="$(grep --include '*_test.go' --exclude-dir 'vendor/' -EIrn 'ioutil.\TempDir' || true)"
+out="$(${grep} --include '*_test.go' --exclude-dir 'vendor/' -EIrn 'ioutil.\TempDir' || true)"
 if [[ -n "$out" ]]; then
   echo >&2 "You used ioutil.TempDir in tests, use 'testutil.NewTempDir()' instead:"
   echo >&2 "$out"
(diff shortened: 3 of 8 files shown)


cmd/krew/cmd/list.go
```diff
@@ -67,6 +69,8 @@ Remarks:
 		PreRunE: checkIndex,
 	}
 
+	listCmd.Flags().BoolVarP(overrideFlag, "override", "o", false, "override standard limited terminal output behavior")
```
corneliusweig: `override` does not say what it overrides and is thus not a good name. I think `--wide` is pretty common in k8s and should convey the workings quite well.

hack/verify-code-patterns.sh
```diff
@@ -16,16 +16,28 @@
 
 set -euo pipefail
 
+# Appropriately set grep
```
corneliusweig: I think the changes here to our test scripts should be moved to another PR, where we can discuss this independently.

integration_test/testutil_test.go
```diff
@@ -276,6 +295,26 @@ func (it *ITest) initializeIndex() {
 	if err := cmd.Run(); err != nil {
 		it.t.Fatalf("cannot restore index from cache: %s", err)
 	}
+
+	pluginList := indexDir + "/plugins/"
+
+	getVersion := func(dir, plugin string) string {
```
corneliusweig: Determining the version is quite klunky. If you want to assert the version, why not just use plugin `foo` from `sigs.k8s.io/krew/integration_test/testdata`. This has a known, fixed version string.

integration_test/testutil_test.go
```diff
@@ -37,10 +38,28 @@ import (
 const (
 	persistentIndexCache = "krew-persistent-index-cache"
 	krewBinaryEnv        = "KREW_BINARY"
-	validPlugin          = "konfig" // a plugin in central index with small size
-	validPlugin2         = "mtail"  // a plugin in central index with small size
+	validPlugin          = "konfig"          // a plugin in central index with small size
```
corneliusweig: This list of (valid) external plugins is getting out of hand. Please don't add that many :pray:

integration_test/list_test.go
```diff
@@ -40,5 +40,41 @@ func TestKrewList(t *testing.T) {
...
+
+	test.Krew("install", validPlugin5, validPlugin4, validPlugin3, validPlugin2, validPlugin).RunOrFail()
+	expected = []byte(validPlugin + "\n" + validPlugin2 + "\n" + validPlugin5 + "\n" + validPlugin4 + "\n" + validPlugin3 + "\n")
+
+	sortedList := test.Krew("list").RunOrFailOutput()
+	if diff := cmp.Diff(sortedList, expected); diff != "" {
+		t.Fatalf("'list' output doesn't match:\n%s", diff)
+	}
+
+	genSpaces := func(s string) string {
```
corneliusweig: This `genSpaces` hack is not necessary IMHO. Why not just use a reasonable regexp?

integration_test/list_test.go
```diff
@@ -40,5 +40,41 @@ func TestKrewList(t *testing.T) {
...
+
+	test.Krew("uninstall", validPlugin).RunOrFail()
+	expected = []byte{'\n'}
+
+	uninstallList := test.Krew("list").RunOrFailOutput()
+	if diff := cmp.Diff(uninstallList, expected); diff != "" {
+		t.Fatalf("'list' output doesn't match:\n%s", diff)
+	}
+
+	test.Krew("install", validPlugin5, validPlugin4, validPlugin3, validPlugin2, validPlugin).RunOrFail()
```
corneliusweig: I think checking the sorting is not a good test-case for e2e tests:

* it requires installing several plugins, which is slow
* it can be tested independently with a unit test, if the production logic is refactored and the sorting is pulled out of the `cmd` package.

So I'd recommend to remove this.

integration_test/list_test.go
```diff
@@ -40,5 +40,41 @@ func TestKrewList(t *testing.T) {
 	if diff := cmp.Diff(eventualList, expected); diff != "" {
 		t.Fatalf("'list' output doesn't match:\n%s", diff)
 	}
-	// TODO(ahmetb): install multiple plugins and see if the output is sorted
+
```
corneliusweig: The other integration test files cleanly separate test cases into separate test functions. Please also follow that pattern here.

go.mod
```diff
@@ -15,9 +15,11 @@ require (
 	github.com/sahilm/fuzzy v0.0.5
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/pflag v1.0.3
+	golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2
```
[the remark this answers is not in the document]
JJJJJones: That seems to have slipped by from earlier, i'll run `go mod tidy` and commit.

cmd/krew/cmd/list.go
```diff
@@ -46,7 +48,7 @@ Remarks:
 			}
 
 			// return sorted list of plugin names when piped to other commands or file
-			if !isTerminal(os.Stdout) {
+			if !isTerminal(os.Stdout) && !*overrideFlag {
```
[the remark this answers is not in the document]
JJJJJones: I'll write that option up and commit it

cmd/krew/cmd/list.go
JJJJJones: I'll rename it right now

integration_test/list_test.go
JJJJJones: > it requires installing several plugins, which is slow

The plugins selected are all shell scripts to keep things quick. I'm not sure how performance sensitive the testing actually is, `initializeIndex()` re-expands the indexDir multiple times in testing. Maybe all of `intiailzeIndex()` should fall under `initIndexOnce.Do()`?

integration_test/list_test.go
JJJJJones: At the time I expected a reasonable regex to end up being more complicated; i'll swap it out.

integration_test/list_test.go
JJJJJones: Changed and i'll move it to another PR

hack/verify-code-patterns.sh
JJJJJones: Undoing the changes, moving them to a different PR

JJJJJones:
> Hey, thanks for contributing!

Your most welcome hopefully this (minor) addition helps others.

> Let's first discuss if this flag is really needed. Can you comment why krew info <name> is not suitable for your purpose?

Our organization has a package manager for the tools our devs use to work on the Kubernetes cluster and i'm looking to bundle Krew support into the next release. With Krew being the official plugin manager I think circumventing Krew to manage plugins ourselves would be a poor choice and not worth the development time. Krew's problem is that the `krew info <plugin>` breaks existing abstractions of version checking logic but `krew list` slots in perfectly.

`krew list` succinctly conveys information but not all calls of it will be from the terminal, so to limit the options for those utilizing a more automated or external process can be undesirable.

As for preventing harm: I couldn't think of a separate user workflow that is broken by the addition of a `--wide` flag while the upkeep of the addition is very minimal for maintainers.

integration_test/testutil_test.go
JJJJJones: Moving to a new PR

integration_test/list_test.go
JJJJJones: TBC at the new PR

integration_test/testutil_test.go
JJJJJones: TBC new PR

cmd/krew/cmd/list.go
JJJJJones: Renamed to "--wide"/"-w" and added "--names-only"

corneliusweig:
I'm thinking that switching the output format based on whether it is a terminal or not has its disadvantages (e.g. testing or parsing the version). So I like the idea of adding one of `--wide` or `--names`/`--names-only` and remove the `isTerminal` magic.
@ahmetb What's your take on this?

ahmetb:
Sorry, I haven't been tracking this PR actively.
Why are we trying to bypass the isTTY check? Can someone give a quick rundown of what's happening?

corneliusweig:
@JJJJJones wants to extract version info from the `krew list` command. But when he tries to pipe its output, `krew` switches to name-only mode. So the suggestion is to add a flag to prohibit that -- or disable the automagic terminal detection to change the output format.

ahmetb:
Sorry, still haven't looked at the code but based on what you say @corneliusweig, we're solving this at the wrong place.

Please do not complicate isTTY path. This is adding an "exception" to a "rule" and makes it hard to test/maintain. (In general we should be against these things).

Does this user actually need every plugin's version number or just one?

I recommend offering this information on `krew info <NAME>` command output as a field, and potentially offer structured output (json, yaml, or possibly even jsonpath).

Then they would do:


```
for plugin in `$(kubectl krew list)`; do
    kubectl krew info "$plugin" --format=json | jq -r '.currentVersion'
done
```

`kubectl krew info` already shows version number, we can augment this as "available version" and "current version" if necessary. Then the code would look like:

corneliusweig:
I agree that on override to the isTTY behavior is not a good idea. But my question is: should we keep the isTTY behavior? Or should there rather be a flag to switch between name-only and wide mode? After all, it is quite surprising that `krew list` output changes when redirected to a file/pipe.

In the long term, offering json/yaml output to `krew info` is the cleaner approach. @JJJJJones would the solution suggested by @.ahmetb work for you?

JJJJJones:
Hello @ahmetb.

> Does this user actually need every plugin's version number or just one?

We do need every plugin's version.

> Sorry, still haven't looked at the code
> Please do not complicate isTTY path. This is adding an "exception" to a "rule" and makes it hard to test/maintain. (In general we should be against these things).

Please do look at the code, its an extremely brief addition with no impact on the default isTTY behavior of `krew list`; my intent from the start has been to keep this extremely easy to maintain. I've also written up the integration tests though that i'm moving to a different PR.

> This is adding an "exception" to a "rule"
> But my question is: should we keep the isTTY behavior?

What was the reason for this rule in the first place?

> In the long term, offering json/yaml output to krew info is the cleaner approach.

I think that's a great addition to Krew but for my needs i'm primarily focused on bypassing isTTY() or, as might be suggested, removing isTTY(). `krew list` provides exactly the info I need and the arbitrary change in its behavior of isTTY() is the problem.

> @JJJJJones would the solution suggested by @.ahmetb work for you?

It can but isn't the quickest, most optimal solution for our needs. `krew info <plugin>` breaks existing abstractions in the version checking logic along with added overhead to support. `krew list` (plugins+versions) slots into our tool perfectly.

ahmetb:
Rule: “krew list” is used to backup list of versions so we don’t want to change it’s output. 

Why don’t you want something like: krew list —format={json,yaml} instead?

You are not supposed to parse table outputs from kubectl. I can guarantee you we’ll break the table format within less than a year.

JJJJJones:
> Rule: “krew list” is used to backup list of versions so we don’t want to change it’s output (when redirected to a file).

`krew list` doesn't output versions when redirected to a file though.

> You are not supposed to parse table outputs from kubectl. I can guarantee you we’ll break the table format within less than a year.

Krew could also just keep the table format as an option and just add JSON/YAML.

> Why don’t you want something like: krew list —format={json,yaml} instead?

I do, its just not what we need right now for out tool.

ahmetb:
> `krew list` doesn't output versions when redirected to a file though.

Precisely. And this is by design. We just want plugin names to be backed up so users can do:

```
krew list > backup.txt
krew install < backup.txt
```

Since we don't support specifying version during installing, it does not make sense to print version to the output.

> I do, its just not what we need right now for our tool.

Since this patch (behavior change, add new flag) is not reversible, once that flag is there we have to support it for a long time. That's why we can't make a short-term solution like this.

Let's open a new issue that proposes adding a structured output flag to `krew list`. We can easily implement it and it would not get a pushback. 

I'm gonna close this because I currently don't see a reason where we should ignore isTTY detection for producing a parseable output. (Also a good reminder for you: Please open an issue to propose a change first rather than submitting the code, and this goes for other OSS repos as well.) 
/close

JJJJJones:
> Precisely. And this is by design. We just want plugin names to be backed up so users can do:

Then why have `kubectl krew list` print out version info to begin with? I understand wanting the default behavior for redirected output be to just print the plugin names, but why arbitrarily deny a capability afforded to end-users?

> Since we don't support specifying version during installing, it does not make sense to print version to the output.

Yes which this PR didn't change.

`krew list > backup.txt`
`krew install < backup.txt`

This would still behave identically.

> Since this patch (behavior change, add new flag) is not reversible, once that flag is there we have to support it for a long time.

Its 4 lines of code (I forgot to revert the go modules) and I already drafted the integration test for it, so supporting such a minor addition has been covered.

> That's why we can't make a short-term solution like this.

I planned on making this our long-term solution since most other tools don't deliver structured version output that needing to support such for Krew (and Krew alone) seemed entirely undesirable from any and all angles. I can't speak with outright certainty that the solution i've proposed here will *always* be the right one so I left my language vague.

> Also a good reminder for you: Please open an issue to propose a change first

My apologies, i'll open one for the problems with the shell scripts for MacOS i'm experiencing.

> Let's open a new issue that proposes adding a structured output flag to krew list. We can easily implement it and it would not get a pushback.

My hope was to avoid this because it would break abstractions and add overhead to our tool. The aversion to such a simple feature with such negligible maintenance (most of which I have handled) gives me pause. Structured output is indeed very nice, but for some use-cases YAML/JSON is unnecessary.

ahmetb:
I think we're not using the same language here. :)

> Then why have `kubectl krew list` print out version info to begin with?

Because `krew` commands are meant to be "human readable"; not machine parseable.

`krew install` and `kubectl list` specifically consume/produce output that's different when piped. That's used for backups. And that's been like that since the beginning. It's part of our contract.


> This would still behave identically.

You are correct, however:

> Its 4 lines of code (I forgot to revert the go modules) and I already drafted the integration test for it, so supporting such a minor addition has been covered.

If everyone wanted to add command-line options (which, as I said in my previous comment, **we have to support for a long time**), we would end up with an unmaintainable set of flags.

And it's not 4 lines of code. You just introduced a corner case that needs to be covered with an integration test?

> but for some use-cases YAML/JSON is unnecessary and excessive.

You are absolutely right. That's why `kubectl` exposes [various options](https://kubernetes.io/docs/reference/kubectl/overview/#output-options) like:

- `--output="jsonpath={...formatting...}"`
- `--output="custom-columns={..column list..}"` (which is what you want)
- `--output=yaml`, `--output=json`, `--output=name`

Maybe for you json/yaml is excessive but many others will find the structured nature of json/yaml very reassuring and useful. It's a standard in Kubernetes ecosystem and makes sense to apply here. (In fact, we'd be inheriting _most of the code_ to make this happen from kubernetes genericclioptions package.)

That's why this is the right thing to do.

JJJJJones:
> And it's not 4 lines of code. You just introduced a corner case that needs to be covered with an integration test?

That corner case already existed as there was no testing of non-redirected output from `krew list`.

> Maybe for you json/yaml is excessive but many others will find the structured nature of json/yaml very reassuring and useful

Yes. I wasn't saying JSON/YAML are excessive in every case or that JSON/YAML output shouldn't be an option, just that in our case its excessive.

> It's a standard in Kubernetes ecosystem and makes sense to apply here. (In fact, we'd be inheriting most of the code to make this happen from kubernetes genericclioptions package.)

Great, if the flags added by output formatting for `krew list` aren't excessive and we can have a custom-columns output and its coming from genericclioptions ([this?](https://github.com/kubernetes/cli-runtime/tree/master/pkg/genericclioptions)) then we can all get what we wanted. I'll make an issue.

JJJJJones:
Issue posted: https://github.com/kubernetes-sigs/krew/issues/356

PR #360, opened 2019-10-29
add support for symlinks in tarballs

### LINKED ISSUE #359: Support symbolic links in archive files
i have a tar.gz file that contains symlinks. when krew (0.3.1) tries to install my archive, i see this error:

> failed to install plugin "kubeui": install failed: failed to download and extract: failed to download and verify file: failed to extract file: unable to handle file type 50 for "Kui-darwin-x64/Kui.app/Contents/Frameworks/Mantle.framework/Mantle" in tar

that file is a symlink, and the first symlink in the archive. i don't see anything else special about it that might lead to errors.

starpit:
This PR extends the downloader to support symlinks in tarball archives.

Fixes #359


corneliusweig:
/hold
Wait until discussion in #359 is settled.

pkg/download/downloader.go
```diff
@@ -125,6 +125,10 @@ func extractTARGZ(targetDir string, at io.ReaderAt, size int64) error {
 			if err := os.MkdirAll(path, os.FileMode(hdr.Mode)); err != nil {
 				return errors.Wrap(err, "failed to create directory from tar")
 			}
+		case tar.TypeSymlink:
```
ahmetb: fwiw .zip files will need this too.

pkg/download/downloader.go
```diff
@@ -125,6 +125,18 @@ func extractTARGZ(targetDir string, at io.ReaderAt, size int64) error {
...
+			// the following two guard enforce a very conservative symlink policy:
+			// 1) no ".." references; and 2) no absolute path references
+			if strings.Contains(hdr.Linkname, "..") {
```
ahmetb: this is... dangerous, and not necessarily authoritative, as one might work around it using escape characters etc. :) 

We should calculate it using a method we have that's tested, like IsSubPath or something.

ahmetb: Also `..` references shouldn't be prevented. It makes sense to me have symlinks that refer to something on an upper directory that's still in the plugin's extract directory. (In fact, I assumed that's your primary use case.)

We need to probably calculate the final path of the link and use path cleaning + then check if it falls within the plugin's extraction directory.

We might consider linking the symlinks as absolute, but I don't see any upsides to that and makes it harder to move ~/.krew, if needed for some reason.

starpit: are we ok with tarballs that escape to the staging directory? i.e. one level up? given that krew unpacks everything in a staging directory this is probably ok? if we aren't ok with this, then the escape detection logic gets more complex. we'd have to keep track of every tree in the forest being extracted

starpit:
i pushed a commit that improves the escape detection logic. this latest version allows escape to the staging directory, but nothing outside of that. 

also, on the comment re: string comparison, it now compares the result of two canonicalized filepaths, rather than a random string and a filepath string.

starpit:
added zip support with the the fourth commit

pkg/download/downloader.go
ahmetb: these two sound like conflicting statements:

> are we ok with tarballs that escape to the staging directory?

> given that krew unpacks everything in a staging directory this is probably ok?

We need to make sure the computed target (path join+clean) of a symlink doesn't escape the `targetDir` parameter of the function.

starpit: thanks, yes. the latest code does this. i forgot to update the comment. fixing.

pkg/download/downloader.go
```diff
@@ -92,6 +105,30 @@ func extractZIP(targetDir string, read io.ReaderAt, size int64) error {
 	return nil
 }
 
+func symlink(targetDir, oldname, newname string) error {
```
ahmetb: I believe we already have symlink methods in the rest of the codebase.
Would be good to extract them into a pkg perhaps.

Also this `symlink` name doesn't reflect what the method does.

Also for escape checks, can you consider using https://godoc.org/sigs.k8s.io/krew/pkg/pathutil#IsSubPath which I mentioned earlier?

pkg/download/downloader_test.go
```diff
@@ -70,38 +107,82 @@ func Test_extractZIP(t *testing.T) {
 		defer zipReader.Close()
 		stat, _ := zipReader.Stat()
 		if err := extractZIP(tmpDir.Root(), zipReader, stat.Size()); err != nil {
-			t.Fatalf("extractZIP(%s) error = %v", tt.in, err)
+			if !tt.expectErr {
```
ahmetb: Are you checking where `tt.expectErr` is set to `true` but `err==nil`?
(This is repeated in several other places.)

It seems like there's a check for this below but I think they can be merged into a single if/else block I guess?

pkg/download/downloader.go
starpit: thanks for the pointer ot IsSubPath. it is close. unfortunately i think this case is a bit different. consider this case:

foo/bar/../../../baz -> newname

we need to canonicalize the oldname. go does not have a relative-canonicalize function in its standard library. there is filepath.Abs, which does canonicalize, but it also makes the path absolute.

ahmetb: filepath.Clean

ahmetb: Which is also used in IsSubPath :)

starpit: thanks

pkg/download/downloader_test.go
starpit: there are four conditions to be checked:

1) did the extract fail in a way that wasn't expected?
2) if it didn't fail, was a failure expected?
3a) otherwise the zip part worked; now check whether the expected files appeared; but, 3b) skip this part if it failed as expected

i think there are four paths in the code, matching these four cases.

starpit:
updated to use pathutil; also cleaned up the downloader test to avoid repeating the symlink test spec for zip versus tarball variants

pkg/download/downloader.go
```diff
@@ -72,6 +74,19 @@ func extractZIP(targetDir string, read io.ReaderAt, size int64) error {
 			return errors.Wrap(err, "could not open inflating zip file")
 		}
 
+		if f.FileInfo().Mode()&os.ModeSymlink == os.ModeSymlink {
+			buf := new(bytes.Buffer)
```
ahmetb: can use ioutil.readall here

pkg/download/downloader.go
```diff
@@ -72,6 +74,19 @@ func extractZIP(targetDir string, read io.ReaderAt, size int64) error {
...
+			_, err := io.Copy(buf, src)
+			if err != nil {
+				return errors.Wrap(err, "problem reading symlink target")
+			}
+			oldname := strings.TrimSpace(buf.String())
```
ahmetb: why trimming space? maybe the filename has a trailing space that requires preserving?

pkg/download/downloader_test.go
```diff
@@ -37,24 +37,91 @@ func testdataPath() string {
...
+
+func SymlinkTest4(ext string) ArchiveTest {
+	return ArchiveTest{
+		in:      "test-with-symlinks-escaping-parent3" + ext,
+		files:   nil,
+		wantErr: true,
+	}
+}
+
+func SymlinkTest5(ext string) ArchiveTest {
```
ahmetb: I think we have waaay to many tests here.

For testdata, we just need 1 symlink .tar.gz and 1 .zip in total to verify that works.

For everything else (such as disallowed links), we can test them by unit-testing the method that decides it.

pkg/download/downloader_test.go
```diff
@@ -37,24 +37,91 @@ func testdataPath() string {
...
-		files []string
-	}{
+	tests := []ArchiveTest{
 		{
 			in: "test-with-directory.zip",
 			files: []string{
 				"/test/",
 				"/test/foo",
 			},
+			wantErr: false,
```
ahmetb: as you can tell, these tests were designed to succeed. for "error" cases, you can test them in unit-tests.

ahmetb:
@starpit Left more comments, I feel like this is going to be quite a lot of back and forth to get right. I can take over the PR anytime and iterate on it, if you need me to.

pkg/download/downloader_test.go
starpit: sure. personally, i'm a fan of end-to-end tests when they are cheap.

starpit:
- ioutil and trim fixes
- removed two end-to-end tests in favor of a small suite of unit tests against `isSymlinkAllowed()`

pkg/download/downloader_test.go
```diff
@@ -55,6 +95,28 @@ func Test_extractZIP(t *testing.T) {
...
+		t.Fatalf("disallowed symlink escape was allowed")
+	}
+	if err := isSymlinkAllowed("foo/bar", "gorp/../..", "foo/bar/baz/bam"); err != nil {
+		t.Fatalf("allowed symlink escape was disallowed")
+	}
+	if err := isSymlinkAllowed("foo/bar", "gorp/../../..", "foo/bar/baz/bam"); err == nil {
+		t.Fatalf("disallowed symlink escape was allowed")
+	}
+	if err := isSymlinkAllowed("foo/bar", "/absolute", "foo/bar/baz/bam"); err == nil {
+		t.Fatalf("disallowed symlink absolute path escape was allowed")
```
ahmetb: you can keep these but these are better suited for a `TestIsSymlinkAllowed` method.

pkg/download/downloader.go
```diff
@@ -92,6 +107,38 @@ func extractZIP(targetDir string, read io.ReaderAt, size int64) error {
...
+	}
+
+	return nil
+}
+
+// make a symlink from oldname to newname within the given enclosing targetDir
+// enforces two symlink policies
+// 1) no symlinks to absolute paths (oldname must not be an absolute path)
+// 2) no symlinks that escape targetDir
+func symlinkIfAllowed(targetDir, oldname, newpath string) error {
```
ahmetb: better names for this is would be: makeSafeSymlink or makeRelativeOnlySymlink

pkg/download/downloader.go
```diff
@@ -92,6 +107,38 @@ func extractZIP(targetDir string, read io.ReaderAt, size int64) error {
...
+
+	if _, isNonEscaping := pathutil.IsSubPath(targetDir, oldpath); !isNonEscaping {
+		glog.V(4).Infof("invalid escaping symlink %v -> %v\n", oldname, newpath)
+		return errors.New("invalid symlink referencing a parent path in tar")
+	}
+
+	return nil
+}
+
+// make a symlink from oldname to newname within the given enclosing targetDir
```
ahmetb: nit: GoDoc convention requires first word of the comment is the name of the method.

pkg/download/downloader.go
```diff
@@ -72,6 +75,18 @@ func extractZIP(targetDir string, read io.ReaderAt, size int64) error {
 			return errors.Wrap(err, "could not open inflating zip file")
 		}
 
+		if f.FileInfo().Mode()&os.ModeSymlink == os.ModeSymlink {
+			buf, err := ioutil.ReadAll(src)
+			if err != nil {
+				return errors.Wrap(err, "problem reading symlink target")
+			}
+			oldname := string(buf)
```
ahmetb: var name "oldname" doesn't make sense there, something like dst or target is better

pkg/download/downloader.go
```diff
@@ -92,6 +107,38 @@ func extractZIP(targetDir string, read io.ReaderAt, size int64) error {
 	return nil
 }
 
+func isSymlinkAllowed(targetDir, oldname, newpath string) error {
```
ahmetb: inconsistent capitalization and wording: targetDir, oldname, newpath.
proposed: targetDir, srcPath, dstPath

pkg/download/downloader.go
```diff
@@ -92,6 +107,38 @@ func extractZIP(targetDir string, read io.ReaderAt, size int64) error {
...
+// make a symlink from oldname to newname within the given enclosing targetDir
+// enforces two symlink policies
+// 1) no symlinks to absolute paths (oldname must not be an absolute path)
+// 2) no symlinks that escape targetDir
+func symlinkIfAllowed(targetDir, oldname, newpath string) error {
+	if err := isSymlinkAllowed(targetDir, oldname, newpath); err != nil {
+		return err
+	}
+	if err := os.Symlink(oldname, newpath); err != nil {
+		return errors.Wrap(err, "failed to create symlink from tar")
```
ahmetb: nit: you can actually directly return this `errors.Wrap` result as when err is nil, it will return nil.

pkg/download/downloader_test.go
```diff
@@ -37,11 +37,51 @@ func testdataPath() string {
 	return filepath.Join(pwd, "testdata")
 }
 
+type ArchiveTest struct {
+	in    string
+	files []string
+}
+
+func SymlinkTest1(ext string) ArchiveTest {
```
ahmetb: names like 1 2 3  are rarely helpful. Full names are better even when they're long.

pkg/download/downloader.go
starpit: oldname comes from the documentation for os.Symlink: https://golang.org/pkg/os/#Symlink

pkg/download/downloader_test.go
ahmetb: reading again, these don't seem to have anything to do with the rest of this test method. let's move out to unit-test.

pkg/download/downloader.go
starpit: these inconsistencies come from the surrounding context (os.Symlink). i feel at this point that you should just make the little changes you feel are appropriate?

ahmetb:
With this patch are you currently able to get an Electron-based plugin archive to be installed by Krew?

pkg/download/downloader.go
tallclair: Needs documentation. What is the expected behavior of this function?

pkg/download/downloader.go
```diff
@@ -92,6 +107,38 @@ func extractZIP(targetDir string, read io.ReaderAt, size int64) error {
...
+	// no symlinks to absolute paths
+	if filepath.IsAbs(oldname) {
+		glog.V(4).Infof("invalid absolute symlink %v -> %v\n", oldname, newpath)
+		return errors.New("invalid symlink referencing an absolute path in tar")
+	}
+
+	oldpath := filepath.Join(filepath.Dir(newpath), filepath.Clean(oldname))
+	fmt.Printf("LINK %s %s\n", oldpath, newpath)
+
+	if _, isNonEscaping := pathutil.IsSubPath(targetDir, oldpath); !isNonEscaping {
```
tallclair: This is going to be vulnerable to symlink chaining. E.g.

```
target = /a/b
tree:
/a/b/c
  c1 -> ../
  c2 -> c1/..
  c3 -> c2/..
```

c1 oldpath resolves to /a/b -> allowed
c2 oldpath resolves to /a/b/c -> allowed
c3 oldpath resolves to /a/b/c -> allowed

but c3 actually points to /a/b/c/../../.. = /

starpit: thanks for the comments @tallclair! i think i understand risk, e.g.

```sh
echo "no sir" > /tmp/nope
mkdir -p /tmp/stage/inner && cd /tmp/stage/inner
ln -s .. l1 # <-- l1 points to /tmp/stage
ln -s l1/.. l2 # <-- l2 points to /tmp
cat l2/nope # <-- oops! because l2/nope is /tmp/nope
```

however, line 120 validates `oldpath`, *not* `oldname

in my two example links, oldname would be `..` and `l1/..`, respectively -- both seemingly safe

whereas oldpath would be `/tmp/stage` for the first symlink, which correctly passes the `IsSubPath` validator (i.e. `isNonEscaping==true`); whereas, for the second link, oldpath would be `/tmp` which would result `isNonEscaping==false`

i must be missing something, in my analysis, because i'm sure much deeper and more careful thought has been put into this problem than in my tea-infused morning thoughts.

starpit: thanks for finding this. in go, `filepath.Clean` also behaves oddly, e.g. with

```sh
ln -s l1 ..
ln -s l1/.. l2
```

if we process the second symlink before the first, then `filepath.Clean("l1/..")` oddly returns ".". And so the currently logic in this PR validates l2, even if we want to disallow it as escaping. 

on the other hand, if we process these in creation order, then i think we can be safe (where we add a call to filepath.Abs, i.e. 
```go
oldpath, err := filepath.Abs(... what we currently assign to oldpath)
```

ahmetb: I think @tallclair has meant this scenario:

```
c1 -> ../..
c2 -> ./c1/foo
```

in this case, c2 will be escaping the extract dir.

(It's expected filepath.Clean("foo/..") will return ".")

We need to make sure everything we pass to [IsSubPath](https://godoc.org/sigs.k8s.io/krew/pkg/pathutil#IsSubPath) is evaluated first if there are symlinks on the way (recursively).

Therefore in your code, we will need to `readlink` (which I am not sure if works recursively) on oldname as well as newpath.

soltysh:
I'll be personally in favour of what Tim proposes, check the linked PR of his and apply that same pattern here.

corneliusweig:
@starpit You mentioned symlink problems of kui+krew in today's SIG-CLI meeting. Can you give a small update as to the current state (you also mentioned that OSX is a particular challenge)? In particular, it would be interesting for us if this PR is still important for your work on kui.

Issue #358, opened 2019-10-29
even with --archive option, krew still insists on validating sha256


starpit:
krew nicely lets me override the manifest with a `--archive` command line option. however, if my manifest specifies a sha256, the krew installation will fail, because the manifest specifies a sha256 that will surely not match the archive given on the command line. maybe i'm missing something?

as a closely related issue, the error reporting in this case could perhaps use some improvement.

if i invoke krew with
```sh
kubectl krew install --manifest=kubeui.yaml --archive=./dist/electron/Kui-darwin-x64.tar.gz 
```

i get this error, which is a sha256 validation error, but the `uri` string is misleading, as it does not match the `--archive=` option. also it took me a while to parse this out. i first read "failed to download and extract", and only with more careful parsing (and ignoring the misleading uri string) did i finally parse this out as a sha256 mismatch error

> failed to install plugin "kubeui": install failed: failed to download and extract: failed to download and verify file: failed to get the uri "https://github.com/kui-shell/plugin-kubeui/releases/download/v1.0.0/kubectl-kubeui_darwin_amd64.zip": checksum does not match, want: f6f2509fc58607030659b90424fabf96a25981652a162fccbd195158bdf3dca0, got 162dd81efcd806a94346535c87724cf06125d9c64677d2c61294304bddeedfb0

ahmetb:
I recall there was a time we skipped checksum verification for local --archive files. Maybe there's a bug that prevents that.

You're right, it would be be better if the URL mentioned the local file and not the URL in the manifest (despite not actually downloading).

ahmetb:
/cc @ferhatelmas
/cc @corneliusweig 
can one of you please investigate?

corneliusweig:
Thanks for reporting this. Indeed, the error message is a bit misleading here, so that should be improved.

On the other hand, I do believe that it is the better behavior to enable shaSum validation for `--manifest`+`--archive`. This is the only way for plugin authors to check if their manifest is valid without first uploading the archive to a hosting platform. Thus, the shaSum check should also fail hard when installing locally.

ahmetb:
> This is the only way for plugin authors to check if their manifest is valid without first uploading the archive to a hosting platform.

Yeah, I started to recall that this is why we enabled checksum check with `--manifest` again. Do you know what uses the noopVerifier today?

starpit:
thanks for the consideration. i guess i found it a bit tedious to have to modify the file every time i rebuilt my test archive.

[...]

PR #361, opened 2019-10-29
Improve error messages when installing from manifest+archive

corneliusweig:
Related #358

### DIFF
--- a/pkg/download/fetch.go
+++ b/pkg/download/fetch.go
@@ -47,7 +47,8 @@ var _ Fetcher = fileFetcher{}
 type fileFetcher struct{ f string }
 
 func (f fileFetcher) Get(_ string) (io.ReadCloser, error) {
-	return os.Open(f.f)
+	file, err := os.Open(f.f)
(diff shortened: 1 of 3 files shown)


pkg/download/fetch.go
```diff
@@ -37,7 +37,7 @@ type HTTPFetcher struct{}
 func (HTTPFetcher) Get(uri string) (io.ReadCloser, error) {
 	resp, err := http.Get(uri)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrapf(err, "failed to get the uri %q", uri)
```
ahmetb: hmm do you think we could use wrapping `error` of FileFetcher.Get , like:

    failed to open specified archive file for reading: ..

ahmetb:
/lgtm
/approve
/hold 
have only 1 comment.

pkg/download/fetch.go
corneliusweig: Makes total sense, then this is treated uniformly in both FileFetchers.

PR #362, opened 2019-10-29
Remove trueVerifier from main logic

corneliusweig:
This verifier implementation is only used in tests, so it should be defined there.

### DIFF
--- a/pkg/download/downloader_test.go
+++ b/pkg/download/downloader_test.go
@@ -165,7 +165,7 @@ func TestDownloader_Get(t *testing.T) {
 		{
 			name: "successful get",
 			fields: fields{
-				verifier: NewInsecureVerifier(),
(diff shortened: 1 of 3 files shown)


pkg/download/downloader_test.go
```diff
@@ -444,3 +452,37 @@ func Test_extractArchive(t *testing.T) {
 		})
 	}
 }
+
+func TestTrueVerifier(t *testing.T) {
```
corneliusweig: This test was moved in order to co-locate it with the type definition.

ahmetb: Tbh this test doesn’t make sense. 
Now that trueVerifier is a test-only type, we don’t need a test for this anymore.

corneliusweig: Agreed :)

Issue #356, opened 2019-10-28
Proposal: add structured output formats to krew list


JJJJJones:
Discussed previously in https://github.com/kubernetes-sigs/krew/pull/354.

`krew list` can be improved with structured (and custom) output formats.

The code to implement it can come mostly from [genericclioptions](https://github.com/kubernetes/cli-runtime/tree/master/pkg/genericclioptions).

JJJJJones:
I've taken a stab at adding structured output using genericclioptions as @ahmetb suggested and can't do it without unnecessary bloat.

JSONPrinter's PrintObj() expects the runtime.Object to have an API group, which would bloat `krew list` with useless and potentially misleading information.

ahmetb:
@JJJJJones makes sense. In that case, we can simply start with imitating -o/--output flags and supporting only 'json' and 'yaml'. What do you think?

JJJJJones:
My apologies. I misspoke with that comment and tried to delete it; there was an unrelated culprit behind my problems getting PrintObj() to work correctly.

I've got JSON and YAML printing working using genericclioptions's JSONYamlPrintFlags.

JJJJJones:
I've pushed my code: https://github.com/JJJJJones/krew/tree/356

Testing it now

JJJJJones:
Fixed some linting issues, pushed latest.

ahmetb:
You wanna send a PR for review?
If there's too much complexity involved we don't have to use genericclioptions.

Similarly, people will ask and have been asking for --output flag on `krew info` which is not a List response (but rather a single object). We need to make sure the approach can work with that too.

JJJJJones:
>  If there's too much complexity involved we don't have to use genericclioptions.

I wouldn't say there is too much but we'll have a better idea with the PR.

> Similarly, people will ask and have been asking for --output flag on krew info which is not a List response (but rather a single object). We need to make sure the approach can work with that too.

It should slot in quite easily. Either a new runtime.Object can be created to handle all the fields of `krew info` (URI, SHA256, homepage, description) or the runtime.Object that list.go uses can be expanded with those fields but have the JSON/YAML directives to not print those fields when empty.

ahmetb:
/priority P1
since multiple users are trying to rely on this.

JJJJJones:
@ahmetb i'll get started on a design doc

PR #365, opened 2019-11-05
add structured output formats to krew list

JJJJJones:
JSON/YAML printing. A lot of what's done here can be adapted or re-used for updating `krew info` to offer structured formats

See #356 

CC @ahmetb


cmd/krew/cmd/list.go
```diff
@@ -24,14 +24,92 @@ import (
...
 	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/cli-runtime/pkg/printers"
 
 	"sigs.k8s.io/krew/pkg/installation"
 )
 
+type Entry struct {
```
JJJJJones: For `krew info`, Entry can be expanded...

```golang
	URI         string `yaml:"omitifempty" json:"omitifempty"`
	SHA256      string `yaml:"omitifempty" json:"omitifempty"`
	URL         string `yaml:"omitifempty" json:"omitifempty"`
	Description string `yaml:"omitifempty" json:"omitifempty"`
```

Alternatively, a new data structure.

ahmetb:
I actually reviewed this but can't submit my review. GitHub is http500 ing on me.

cmd/krew/cmd/list.go
```diff
@@ -56,20 +134,55 @@ Remarks:
 				return nil
 			}
 
-			// print table
-			var rows [][]string
-			for p, version := range plugins {
-				rows = append(rows, []string{p, version})
+			if *output.OutputFormat == "wide" {
```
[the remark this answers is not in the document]
JJJJJones: Yup I fixed this so isTerminal(os.Stdout) without the --output flag behaves like normal.

cmd/krew/cmd/list.go
```diff
@@ -56,20 +134,55 @@ Remarks:
...
-			rows = sortByFirstColumn(rows)
-			return printTable(os.Stdout, []string{"PLUGIN", "VERSION"}, rows)
+
+			if inArray(*output.OutputFormat, output.AllowedFormats()) {
+				objs := []Entry{}
+				for plugin, version := range plugins {
+					obj := Entry{plugin, version}
+					objs = append(objs, obj)
+				}
+				objs = sortByName(objs)
```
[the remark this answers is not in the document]
JJJJJones: plugins is a map, as far as I know a map can't be sorted?

cmd/krew/cmd/list.go
```diff
@@ -56,20 +134,55 @@ Remarks:
...
 		PreRunE: checkIndex,
 	}
 
+	listCmd.Flags().StringVarP(&outputFormat, "output", "o", outputFormat, "_")
+
+	output.OutputFormat = &outputFormat
 	rootCmd.AddCommand(listCmd)
 }
 
+func inArray(s string, arr []string) bool {
```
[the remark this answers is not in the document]
JJJJJones: > moreover it's recommended to usually inline these things, though I'm sure one of the packages we import probably already offer this method (something in k8s.io)

I'll go looking, please let me know if you find something.

cmd/krew/cmd/list.go
```diff
@@ -56,20 +134,55 @@ Remarks:
...
+					}
+				}
+				return nil
+			}
+			return errors.New("unsupported output format specified")
 		},
 		PreRunE: checkIndex,
 	}
 
+	listCmd.Flags().StringVarP(&outputFormat, "output", "o", outputFormat, "_")
```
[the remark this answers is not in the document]
JJJJJones: I just needed a placeholder as I didn't have a description for the flag.

> also why is the default value listed as outputFormat which is an empty string at this point?

Seemingly no reason, it can and probably should be changed to "". Changed.

cmd/krew/cmd/list.go
```diff
@@ -56,20 +134,55 @@ Remarks:
...
+					rows = append(rows, []string{p, version})
+				}
+				rows = sortByFirstColumn(rows)
+				return printTable(os.Stdout, []string{"PLUGIN", "VERSION"}, rows)
 			}
-			rows = sortByFirstColumn(rows)
-			return printTable(os.Stdout, []string{"PLUGIN", "VERSION"}, rows)
+
+			if inArray(*output.OutputFormat, output.AllowedFormats()) {
+				objs := []Entry{}
```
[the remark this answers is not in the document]
JJJJJones: Fixed! Thanks for the heads-up, better to use the stack than the heap.

cmd/krew/cmd/list.go
```diff
@@ -24,14 +24,92 @@ import (
...
+func (c Consume) SetGroupVersionKind(kind schema.GroupVersionKind) {
+	c.APIVersion, c.Kind = kind.ToAPIVersionAndKind()
+}
+
+func (c Consume) GroupVersionKind() schema.GroupVersionKind {
+	return schema.FromAPIVersionAndKind(c.APIVersion, c.Kind)
+}
+
+func (e Entry) GetObjectKind() schema.ObjectKind {
+	return Consume{"_", "_"}
```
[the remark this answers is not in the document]
ahmetb: We actually have an APIGroup/Kind, should be in pkg/constants.

cmd/krew/cmd/list.go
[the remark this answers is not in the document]
ahmetb: yeah, but we can do a sortedPluginNames and use it in nearly all these printers, instead of sorting in multiple places etc?

cmd/krew/cmd/list.go
```diff
@@ -56,20 +134,56 @@ Remarks:
...
 		PreRunE: checkIndex,
 	}
 
+	listCmd.Flags().StringVarP(&outputFormat, "output", "o", "", "_")
+
+	output.OutputFormat = &outputFormat
 	rootCmd.AddCommand(listCmd)
 }
 
+func sliceContains(s string, arr []string) bool {
```
ahmetb: found it https://godoc.org/k8s.io/kubernetes/pkg/util/slice

cmd/krew/cmd/list.go
```diff
@@ -56,20 +134,56 @@ Remarks:
...
 			}
-			rows = sortByFirstColumn(rows)
-			return printTable(os.Stdout, []string{"PLUGIN", "VERSION"}, rows)
+
+			return errors.New("unsupported output format specified")
 		},
 		PreRunE: checkIndex,
 	}
 
+	listCmd.Flags().StringVarP(&outputFormat, "output", "o", "", "_")
```
ahmetb: Let's add a description (unless "_" is a magic string that makes it work?)

I'm not sure if genericclioptions help here.

cmd/krew/cmd/list.go
```diff
@@ -24,14 +24,92 @@ import (
...
+}
+
 func init() {
+	outputFormat := ""
+	output := NewListFlags()
+
 	// listCmd represents the list command
 	listCmd := &cobra.Command{
-		Use:   "list",
+		Use:   "list [(-o|--output=)json|yaml|wide]",
```
ahmetb: How about `-o=name`, can we make it work easily, since it's implemented right down there?

cmd/krew/cmd/list.go
```diff
@@ -24,14 +24,92 @@ import (
...
+	default:
+		return nil, genericclioptions.NoCompatiblePrinterError{OutputFormat: f.OutputFormat, AllowedFormats: f.AllowedFormats()}
+	}
+
+	return printer, nil
+}
+
 func init() {
+	outputFormat := ""
+	output := NewListFlags()
```
ahmetb: I think you don't need to initialize this outside `RunE` function.
It can be initialized in `RunE` function and used there? Similarly you can also access outputFormat from there.

cmd/krew/cmd/list.go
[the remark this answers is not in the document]
JJJJJones: Resolved below

cmd/krew/cmd/list.go
JJJJJones: I'll switch it over for the next commit.

cmd/krew/cmd/list.go
JJJJJones: Not a problem, i'll add it.

cmd/krew/cmd/list.go
JJJJJones: I'll commit this description, hope that's okay. `Output format. One of json, yaml or wide`

cmd/krew/cmd/list.go
JJJJJones: They have to be declared outside `RunE` for the flags to work.

cmd/krew/cmd/list.go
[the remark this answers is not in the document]
JJJJJones: Changed

cmd/krew/cmd/list.go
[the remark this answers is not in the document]
JJJJJones: I'll try and figure something out. Part of the problem here is that `names` and `wide` break out of the printer implementations. The other part is that they use `[][]string` instead of `Entry` as the data structure.

JJJJJones:
> Thanks a lot for the hard work!

Your very much welcome. If you need to reach me i'm in #sig-cli of the kubernetes slack.

cmd/krew/cmd/list.go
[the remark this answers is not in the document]
JJJJJones: I can experiment with [NamesFlag](https://github.com/kubernetes/cli-runtime/blob/master/pkg/genericclioptions/name_flags.go) but it won't behave identically to `names` / redirect output without a flag.

The code for [PrintObj](https://github.com/kubernetes/cli-runtime/blob/master/pkg/printers/name.go#L110) requires there to be a non-empty kind and prints it as the prefix to all the plugin names. Output would look like this.

```
plugin/cssh
plugin/krew
plugin/tail
```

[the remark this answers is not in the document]
ahmetb: > I can experiment with [NamesFlag](https://github.com/kubernetes/cli-runtime/blob/master/pkg/genericclioptions/name_flags.go) [](/kubernetes/cli-runtime/blob/HEAD@{2019-11-05T20:34:05Z}/pkg/genericclioptions/name_flags.go) but it won't behave identically to `names` / redirect output without a flag.

makes sense, I just realized. let's leave -o-=name out for now then.

cmd/krew/cmd/list.go
```diff
@@ -24,14 +24,94 @@ import (
...
+
+func (f *ListFlags) ToPrinter() (printers.ResourcePrinter, error) {
+	outputFormat := *f.OutputFormat
+	var printer printers.ResourcePrinter
+
+	switch outputFormat {
+	case "json":
+		printer = &printers.JSONPrinter{}
+	case "yaml":
+		printer = &printers.YAMLPrinter{}
```
ahmetb: with these added, we'll need additional integration tests next to the existing ones to verify the flags works as intended (and returns valid correct json and yaml that can be parsed into an object, say, like an `interface{}{}`).

this would make sure we're not printing a table for yaml/json cases, similarly we can see how these handle "empty" outputs as part of the tests.

cmd/krew/cmd/list.go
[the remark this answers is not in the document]
ahmetb: FWIW for sorting we can do this:

- make `installation.ListInstalledPlugins` return a `PluginList` type which is an alias to map[string]string

- implement a `SortedNames() []string` method on `PluginList` type –and use it here.

cmd/krew/cmd/list.go
JJJJJones: I'll draft up some tests.

JJJJJones:
@ahmetb i've drafted up some integration tests. All of them pass successfully _except_ `TestKrewListYAML()`. I don't yet know why it's failing to unmarshal what should be valid YAML.

I'm pretty sure I split the empty list tests because of some funkiness resulting from `test.WithIndex()`.

JJJJJones:
YAML test fixed by magic(?). The yaml identifiers for Plugin that were crashing stopped crashing.

```golang
Plugin := struct {
		Name    string `yaml:"Name"`
		Version string `yaml:"Version"`
	}{"", ""}```

File: integration_test/list_test.go [hunk 13]
```diff
@@ -42,3 +45,122 @@ func TestKrewList(t *testing.T) {
...
+	if diff := cmp.Diff(eventualList, expected); diff != "" {
+		t.Fatalf("'list' output doesn't match:\n%s", diff)
+	}
+
+	Plugin := struct {
+		Name    string
+		Version string
+	}{"", ""}
+
+	err := json.Unmarshal(eventualList, &Plugin)
```
ahmetb: I am bit confused here. Shouldnt this be []Plugin? Since this is list command I expect a list response.

JJJJJones: Only one plugin (foo) is installed in the test so I chose to just use a singular Plugin.

ahmetb: hmm I think when you do `kubectl get` (list cmd) you always get a PodList resource (unless you request a specific resource by NAME); that's why I expect the genericclioptions would return us a json/yaml array here.

any ideas why it's happening?

does it automatically turn into an array if you install multiple plugins
(btw you can install multiple plugins using validPlugin1 validPlugin2 directly)

JJJJJones: My apologies, I miscommunicated. What's returned is a list, but since only one plugin is installed in the test, I treat it as one object rather than a list.

I'll convert the Unmarshal to an array of Plugins.

ahmetb: Maybe try unmarshaling the output into `[]map[string]interface{}` where map[string]interface{} is a placeholder for Plugin object.

This way, we can (1) validate it's valid yaml/json, (2) validate it's an array. :)

cmd/krew/cmd/list.go
```diff
@@ -24,14 +24,94 @@ import (
...
+}
+
+func sortByName(e []Entry) []Entry {
+	sort.Slice(e, func(a, b int) bool {
+		return e[a].Name < e[b].Name
+	})
+	return e
+}
+
+// Consume produces a junk GroupVersionKin for obj.GetObjectKind().GroupVersionKind().Empty() check to eat in PrintObj()
```
corneliusweig:
```suggestion
// Consume produces a junk GroupVersionKind for obj.GetObjectKind().GroupVersionKind().Empty() check to eat in PrintObj()
```

cmd/krew/cmd/list.go
```diff
@@ -24,14 +24,94 @@ import (
...
+}
+
+func (e Entry) GetObjectKind() schema.ObjectKind {
+	return Consume{constants.CurrentAPIVersion, constants.PluginKind}
+}
+
+func (e Entry) DeepCopyObject() runtime.Object {
+	newObj := Entry{e.Name, e.Version}
+	return newObj
+}
```
corneliusweig: There is a lot of boilerplate involved by funneling the json/yaml printing through genericclioptions and the `PrintObj` machinery. What do we gain from it?

AFAICT this will not help to implement `jsonpath` more easily.

cmd/krew/cmd/list.go
```diff
@@ -56,17 +136,44 @@ Remarks:
...
+
+			if slice.ContainsString(output.AllowedFormats(), *output.OutputFormat, nil) {
+				objs := make([]Entry, 0, len(plugins))
+				for plugin, version := range plugins {
+					obj := Entry{plugin, version}
+					objs = append(objs, obj)
+				}
+				objs = sortByName(objs)
+				for _, obj := range objs {
+					p, err := output.ToPrinter()
```
corneliusweig: I think the proper way is to instantiate the printer once outside this for loop and reuse it for every item.

cmd/krew/cmd/list.go
```diff
@@ -56,17 +136,44 @@ Remarks:
...
-			return printTable(os.Stdout, []string{"PLUGIN", "VERSION"}, rows)
+
+			if slice.ContainsString(output.AllowedFormats(), *output.OutputFormat, nil) {
+				objs := make([]Entry, 0, len(plugins))
+				for plugin, version := range plugins {
+					obj := Entry{plugin, version}
+					objs = append(objs, obj)
+				}
+				objs = sortByName(objs)
+				for _, obj := range objs {
```
corneliusweig: The proper way to print lists with the object printer is to pass it the full list and let the object printer handle lists. This is important when you want to read that output in a structured way.

For example, when printing the list as json, the code will produce several json objects like so:
```json
{ "name": "krew", "version": "v0.3.1" }
{ "name": "cssh", "version": "v1.1.1" }
```
instead of
```json
{
    "apiVersion": "v1",
    "kind": "List",
    "items": [
      { "name": "krew", "version": "v0.3.1" },
      { "name": "cssh", "version": "v1.1.1" }
    ]
}
```
The latter should be the preferred format for machine processing.

For that to work properly, you can do something like this https://github.com/corneliusweig/ketall/blob/07fd92429ca33385012971f9cc2a1408c1db6cdd/pkg/ketall/printer/adapter.go#L56

integration_test/list_test.go
JJJJJones: I'll swap it over.

ahmetb: @JJJJJones I was wrong in my review per @corneliusweig's comment above about the return format from -o=json for lists.

It appears like -o=json for lists returns a **json stream** and not json array. So what you were doing here is correct.

But for **tests**, if you install 2 plugins, you need to do something like this:

```
j := json.NewDecoder(b)
i := 0
for err := j.Decode(); err != io.EOF; {
  i++
}
// assert i == 2 (or however many plugins you installed)
```

cmd/krew/cmd/list.go
```diff
@@ -56,17 +136,44 @@ Remarks:
...
+				}
+				return nil
+			}
+
+			return errors.New("unsupported output format specified")
 		},
 		PreRunE: checkIndex,
 	}
 
+	listCmd.Flags().StringVarP(&outputFormat, "output", "o", "", "Output format. One of json, yaml, wide, names")
```
ahmetb: should be "name", right? did you decide to support "name" format?

integration_test/list_test.go
```diff
@@ -42,3 +45,122 @@ func TestKrewList(t *testing.T) {
...
+
+func TestKrewListYAML(t *testing.T) {
+	skipShort(t)
+
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	test.Krew("install",
+		"--manifest", filepath.Join("testdata", "foo.yaml"),
+		"--archive", filepath.Join("testdata", "foo.tar.gz")).
```
ahmetb: fwiw if you install 2 plugins you can test the ordering (sorted by name)

JJJJJones:
> What do we gain by using the PrintObj machinery? AFAICT that only causes a lot of boilerplate and a lot more dependencies. So my main question is, why should we do it this way? Why is this better than relying on plain json.Marshal and yaml.Marshal?

Pros

- Re-usage of Kubernetes packages
- Reduced drift between behavior expected of Kubernetes and Krew
- I avoided 're-inventing the wheel' which, as Roedy Green, pointed out, "[supervisors love that](https://github.com/Droogans/unmaintainable-code#packratting)".

Cons

- Multiple types of bloat
- Maintainer confusion

I'm not going to be the best of judges as to whether the price paid is worthwhile as I slogged through this and have to content myself with the fact I learned more about Kubernetes and golang in the process. Hopefully, i'll soon be able to content myself with this code being merged and a new release being made.

The behavior of cli-runtime's printers is inexplicable / unfriendly? I'm not sure who I should talk to about that. JSONPrinter's PrintObj() does a check for if the obj is of type *runtime.Unknown which theoretically should fail. *runtime.Unknown doesn't implement any of the methods required to be a runtime.Object but that code is acceptable. If I make a switch in a function that takes a runtime.Object as an argument and check for a runtime.Unknown, Golang is smart enough to figure out that can never be the case as runtime.Unknown doesn't implement the methods required to be a runtime.Object.

> (which will make our binary considerably larger)

I don't know what (if any) flags are used for release builds but hack/make-binary produced a 29MB build of Krew, compared to the 11MB's of Krew release v0.3.1.

> Right now this code sits in a file which can only be tested in integration tests. It would be much better to move this code to a printer package where it can be unit-tested independently of the other krew list logic. Do you think this is doable?

Its definitely possible to abstract the code from list.go to be used in info.go as well, see [this comment](https://github.com/kubernetes-sigs/krew/pull/365#discussion_r342708597). That abstracted code should end up being unit-test friendly.

cmd/krew/cmd/list.go
JJJJJones: Definitely much better, committing the change now.

cmd/krew/cmd/list.go
JJJJJones: `jsonpath` can be implemented using the relevant genericclioptions flags.

ahmetb:
> I'm not going to be the best of judges as to whether the price paid is worthwhile as I slogged through this and have to content myself with the fact I learned more about Kubernetes and golang in the process.

I think you can be the judge and I most certainly appreciate it.

It appears like the genericclioptions aren't giving us much here. 😞 I assumed we could just give them a list of `[]runtime.Object`s and they would do the rest for us.  But you implemented the flag name, allowed values, checks, in fact some of the printing, which doesn't seem that difficult anyway. (What value of genericclioption I'm missing?)

So at this point maybe we should stop using genericclioptions and
- for `json`: in a for-loop print a json object every line
- for `yaml`: create an array, print it as yaml doc like `kubectl get ns -o=yaml` does
- we can also implement `name`, which is the "current" pipe behavior (no KIND/ prefix)

ahmetb:
^ The good part: Now we know how it works, and we can still keep the integration tests intact. :)

JJJJJones:
> I think you can be the judge and I most certainly appreciate it.

If this were my project i'd be somewhat conflicted. Its nice to re-use existing Kubernetes code to ensure consistent behavior but that code was tailored to something which Krew isn't. The upkeep isn't worth the hassle, let alone how bloated the Binary and code becomes. If Krew were a project on the level of kubectl thus benefiting from increased dev and maintenance support to care for specialized printer machinations then sure. However, for now, the price of consistency would be to make Krew significantly less developer- and maintainer- friendly which should only hurt progress going forward by increasing the barriers to change.

I originally intended to add jsonpath, go-template, etc. support, hoping it'd be a relatively minor addition through genericclioptions after getting yaml/json working. That was overly optimistic.

> It appears like the genericclioptions aren't giving us much here. 😞 I assumed we could just give them a list of []runtime.Objects and they would do the rest for us.

It kind of is. But making things into runtime.Objects carries baggage and outside json/yaml printing (which was its own fight) supporting other flags from genericclioptions or printers (like tabwriter) from printer to get the job done but by re-using Kubernetes requires more discrete, supporting code that isn't an easy read. I understand what i'm doing with the objects and methods i've written and somewhat why they're necessary, but to everyone else it can only obfuscate and confuse. Even with a refactor to put those objects and methods in a separate source file its still minutiae that a future developer or maintainer will get lost in, blinded to the forest by the trees.

> So at this point maybe we should stop using genericclioptions and

I'll do a swap over to something generic and self-written. I think there's definitely value in making re-usable code to handle the printing and giving it its own source file.

> ^ The good part: Now we know how it works, and we can still keep the integration tests intact. :)

I'm adding TDD expertise to my resume as we speak.

ahmetb:
@JJJJJones let us know what you wanna do in this PR.
We can move ahead with this genericclioptions patch you prepared, or you can probably also implement functionality from scratch.

> I originally intended to add jsonpath, go-template, etc. support, hoping it'd be a relatively minor addition through genericclioptions after getting yaml/json working. That was overly optimistic.

Namely, jsonpath sounds like the only thing we're sort of missing out. But we didn't wanna have it in the first place. So we can just go with -o=name|json|yaml and do it ourselves ––if you want.

JJJJJones:
> We can move ahead with this genericclioptions patch you prepared, or you can probably also implement functionality from scratch.

I'd say go with the patch i've made.

Later, refactor to either go all genericclioptions/printers or implement functionality from scratch and extend support towards info.

**Implementing things through genericclioptions/printers**

- quite a bit is already done
- what isn't done requires bending (a new wide/names flag with new printer(s?))
- what doesn't have to be bent in is behaviorally consistent with other kubernetes projects
- more bloat to confuse maintainers and devs

**Implementing functionality from scratch**

- less bloated
- won't be behaviorally consistent

I'm leaning towards implementing functionality from scratch for later on.

> Namely, jsonpath sounds like the only thing we're sort of missing out. But we didn't wanna have it in the first place. So we can just go with -o=name|json|yaml and do it ourselves ––if you want.

I won't be adding jsonpath myself.

ahmetb:
Sounds good. Let’s proceed as is.

JJJJJones:
@ahmetb i'll resolve the lingering issue(s) shortly.

cmd/krew/cmd/list.go
JJJJJones: 'Name' is supported as "names", i'll change it from "names" to "name".

integration_test/list_test.go
JJJJJones: I'll expand the tests with validPlugin to ensure this.

cmd/krew/cmd/list.go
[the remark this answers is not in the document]
JJJJJones: I've left `-o=name` in just to leave an option of mimicking redirect output for terminal users.

Want me to remove it?

integration_test/list_test.go
JJJJJones: Implemented sorting tests for all output types.

cmd/krew/cmd/list.go
JJJJJones: I made a new type to encapsulate []Entry and implement runtime.Object so an array of objects is printed.

cmd/krew/cmd/list.go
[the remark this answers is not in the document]
ahmetb: We can keep -o=name it’s cool with me. 
Just need to write tests for it, too.

cmd/krew/cmd/list.go
```diff
@@ -24,14 +24,94 @@ import (
...
+func NewListFlags() *ListFlags {
+	outputFormat := ""
+	return &ListFlags{
+		JSONYamlPrintFlags: genericclioptions.NewJSONYamlPrintFlags(),
+		OutputFormat:       &outputFormat,
+	}
+}
+
+func (f *ListFlags) AllowedFormats() []string {
+	formats := f.JSONYamlPrintFlags.AllowedFormats()
```
corneliusweig: Again, can be one line.

cmd/krew/cmd/list.go
```diff
@@ -24,14 +24,105 @@ import (
...
+}
+
 func init() {
+	outputFormat := ""
+	output := NewListFlags()
+
 	// listCmd represents the list command
 	listCmd := &cobra.Command{
-		Use:   "list",
+		Use:   "list [(-o|--output=)json|yaml|wide|name]",
```
corneliusweig: Cobra already deals with showing possible flags. There is no need to specify those explicitly here.

cmd/krew/cmd/list.go
```diff
@@ -24,14 +24,105 @@ import (
...
+		OutputFormat:       &outputFormat,
+	}
+}
+
+func (f *ListFlags) AllowedFormats() []string {
+	formats := f.JSONYamlPrintFlags.AllowedFormats()
+	return formats
+}
+
+func (f *ListFlags) AddFlags(c *cobra.Command) {}
```
corneliusweig: This looks unused and useless.

JJJJJones: Removed

cmd/krew/cmd/list.go
JJJJJones: Removed

integration_test/list_test.go
```diff
@@ -42,3 +47,208 @@ func TestKrewList(t *testing.T) {
...
+		RunOrFail()
+
+	expected := []byte("[\n    {\n        \"Name\": \"foo\",\n        \"Version\": \"v0.1.0\"\n    }\n]\n")
+
+	eventualList := test.WithIndex().Krew("list", "-o", "json").RunOrFailOutput()
+	if diff := cmp.Diff(eventualList, expected); diff != "" {
+		t.Fatalf("'list' output doesn't match:\n%s", diff)
+	}
+
+	Plugins := []struct {
```
corneliusweig: That seems more idiomatic:
```go
var Plugin struct {
		Name    string
		Version string
	}
```

cmd/krew/cmd/list.go
JJJJJones: Swapped

cmd/krew/cmd/list.go
```diff
@@ -24,14 +24,94 @@ import (
...
+func (c Consume) GroupVersionKind() schema.GroupVersionKind {
+	return schema.FromAPIVersionAndKind(c.APIVersion, c.Kind)
+}
+
+func (e Entry) GetObjectKind() schema.ObjectKind {
+	return Consume{constants.CurrentAPIVersion, constants.PluginKind}
+}
+
+func (e Entry) DeepCopyObject() runtime.Object {
+	newObj := Entry{e.Name, e.Version}
```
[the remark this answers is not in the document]
JJJJJones: Edited

integration_test/list_test.go
corneliusweig: When I run `kubectl get pod -o yaml`, I get:
```
apiVersion: v1
items: []
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```
and for json:
```
{
    "apiVersion": "v1",
    "items": [],
    "kind": "List",
    "metadata": {
        "resourceVersion": "",
        "selfLink": ""
    }
}
```

Therefore, I think we should stay in line with what k8s does and print a list of objects instead of an object stream.

corneliusweig: oops.. nevermind. Just saw that you changed it already :bowing_man:

cmd/krew/cmd/list.go
```diff
@@ -24,11 +24,89 @@ import (
...
+func (c Consume) GroupVersionKind() schema.GroupVersionKind {
+	return schema.FromAPIVersionAndKind(c.APIVersion, c.Kind)
+}
+
+func (e Entries) GetObjectKind() schema.ObjectKind {
+	return Consume{constants.CurrentAPIVersion, constants.PluginKind}
+}
+
+func (e Entries) DeepCopyObject() runtime.Object {
+	return append(Entries{}, e...)
```
corneliusweig: This is quite a lazy deep copy ;)
If Entry would have any object members, this would just be a shallow copy. For here it's fine though.

cmd/krew/cmd/list.go
```diff
@@ -24,11 +24,89 @@ import (
...
+}
+
+type ListFlags struct {
+	JSONYamlPrintFlags *genericclioptions.JSONYamlPrintFlags
+	OutputFormat       *string
```
corneliusweig: `OutputFormat` doesn't need to be a pointer, right? What about
```patch
diff --git a/cmd/krew/cmd/list.go b/cmd/krew/cmd/list.go
index c478981..e592a94 100644
--- a/cmd/krew/cmd/list.go
+++ b/cmd/krew/cmd/list.go
@@ -72,15 +72,11 @@ func (e Entries) DeepCopyObject() runtime.Object {
 
 type ListFlags struct {
 	JSONYamlPrintFlags *genericclioptions.JSONYamlPrintFlags
-	OutputFormat       *string
+	OutputFormat       string
 }
 
 func NewListFlags() *ListFlags {
-	outputFormat := ""
-	return &ListFlags{
-		JSONYamlPrintFlags: genericclioptions.NewJSONYamlPrintFlags(),
-		OutputFormat:       &outputFormat,
-	}
+	return &ListFlags{JSONYamlPrintFlags: genericclioptions.NewJSONYamlPrintFlags()}
 }
 
 func (f *ListFlags) AllowedFormats() []string {
@@ -88,23 +84,21 @@ func (f *ListFlags) AllowedFormats() []string {
 }
 
 func (f *ListFlags) ToPrinter() (printers.ResourcePrinter, error) {
-	outputFormat := *f.OutputFormat
 	var printer printers.ResourcePrinter
 
-	switch outputFormat {
+	switch f.OutputFormat {
 	case "json":
 		printer = &printers.JSONPrinter{}
 	case "yaml":
 		printer = &printers.YAMLPrinter{}
 	default:
-		return nil, genericclioptions.NoCompatiblePrinterError{OutputFormat: f.OutputFormat, AllowedFormats: f.AllowedFormats()}
+		return nil, genericclioptions.NoCompatiblePrinterError{OutputFormat: &f.OutputFormat, AllowedFormats: f.AllowedFormats()}
 	}
 
 	return printer, nil
 }
 
 func init() {
-	outputFormat := ""
 	output := NewListFlags()
 
 	// listCmd represents the list command
@@ -124,7 +118,7 @@ Remarks:
 			}
 
 			// return sorted list of plugin names when piped to other commands or file
-			if *output.OutputFormat == "name" || (*output.OutputFormat == "" && !isTerminal(os.Stdout)) {
+			if output.OutputFormat == "name" || (output.OutputFormat == "" && !isTerminal(os.Stdout)) {
 				var names []string
 				for name := range plugins {
 					names = append(names, name)
@@ -134,7 +128,7 @@ Remarks:
 				return nil
 			}
 
-			if *output.OutputFormat == "wide" || (*output.OutputFormat == "" && isTerminal(os.Stdout)) {
+			if output.OutputFormat == "wide" || (output.OutputFormat == "" && isTerminal(os.Stdout)) {
 				// print table
 				var rows [][]string
 				for p, version := range plugins {
@@ -144,7 +138,7 @@ Remarks:
 				return printTable(os.Stdout, []string{"PLUGIN", "VERSION"}, rows)
 			}
 
-			if slice.ContainsString(output.AllowedFormats(), *output.OutputFormat, nil) {
+			if slice.ContainsString(output.AllowedFormats(), output.OutputFormat, nil) {
 				objs := make(Entries, 0, len(plugins))
 				for plugin, version := range plugins {
 					obj := Entry{plugin, version}
@@ -167,9 +161,8 @@ Remarks:
 		PreRunE: checkIndex,
 	}
 
-	listCmd.Flags().StringVarP(&outputFormat, "output", "o", "", "Output format. One of json, yaml, wide, name")
+	listCmd.Flags().StringVarP(&output.OutputFormat, "output", "o", "", "Output format. One of json, yaml, wide, name")
 
-	output.OutputFormat = &outputFormat
 	rootCmd.AddCommand(listCmd)
 }
 
```

cmd/krew/cmd/list.go
```diff
@@ -24,11 +24,89 @@ import (
...
+func (c Consume) SetGroupVersionKind(kind schema.GroupVersionKind) {
+	c.APIVersion, c.Kind = kind.ToAPIVersionAndKind()
+}
+
+func (c Consume) GroupVersionKind() schema.GroupVersionKind {
+	return schema.FromAPIVersionAndKind(c.APIVersion, c.Kind)
+}
+
+func (e Entries) GetObjectKind() schema.ObjectKind {
+	return Consume{constants.CurrentAPIVersion, constants.PluginKind}
```
corneliusweig: Here, you are re-using the version/kind of the krew manifest. As you are using the k8s machinery, you should not do this here, because the object you are serializing is _not_ the krew manifest. It's ok to reuse the api version, but please use a different kind.

cmd/krew/cmd/list.go
JJJJJones: I prefer to think of it as efficient :)

integration_test/list_test.go
JJJJJones: Much, swapped over

cmd/krew/cmd/list.go
```diff
@@ -56,17 +134,42 @@ Remarks:
...
+				for p, version := range plugins {
+					rows = append(rows, []string{p, version})
+				}
+				rows = sortByFirstColumn(rows)
+				return printTable(os.Stdout, []string{"PLUGIN", "VERSION"}, rows)
 			}
-			rows = sortByFirstColumn(rows)
-			return printTable(os.Stdout, []string{"PLUGIN", "VERSION"}, rows)
+
+			if slice.ContainsString(output.AllowedFormats(), *output.OutputFormat, nil) {
```
corneliusweig: This `if` seems to repeat the logic from `ToPrinter`. It is safe to call `ToPrinter` with arbitrary OutpuFormat, because this will detect a reasonable error. However, you should move the `ToPrinter` call upfront to catch this error early. Like so:
```patch
diff --git a/cmd/krew/cmd/list.go b/cmd/krew/cmd/list.go
index c478981..e6e25c1 100644
--- a/cmd/krew/cmd/list.go
+++ b/cmd/krew/cmd/list.go
@@ -28,8 +28,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/printers"
-	"k8s.io/kubectl/pkg/util/slice"
-
 	"sigs.k8s.io/krew/pkg/installation"
 )
@@ -144,25 +142,21 @@ Remarks:
 				return printTable(os.Stdout, []string{"PLUGIN", "VERSION"}, rows)
 			}
 
-			if slice.ContainsString(output.AllowedFormats(), *output.OutputFormat, nil) {
-				objs := make(Entries, 0, len(plugins))
-				for plugin, version := range plugins {
-					obj := Entry{plugin, version}
-					objs = append(objs, obj)
-				}
-				objs = sortByName(objs)
-				p, err := output.ToPrinter()
-				if err != nil {
-					return err
-				}
-				err = p.PrintObj(objs, os.Stdout)
-				if err != nil {
-					return err
-				}
-				return nil
+			p, err := output.ToPrinter()
+			if err != nil {
+				return err
 			}
-
-			return errors.New("unsupported output format specified")
+			objs := make(Entries, 0, len(plugins))
+			for plugin, version := range plugins {
+				obj := Entry{plugin, version}
+				objs = append(objs, obj)
+			}
+			objs = sortByName(objs)
+			err = p.PrintObj(objs, os.Stdout)
+			if err != nil {
+				return err
+			}
+			return nil
 		},
 		PreRunE: checkIndex,
 	}
```

integration_test/list_test.go
JJJJJones: Sorry, I forgot to make a comment here.

For anyone not caught up to speed, I already swapped from a JSON stream to array.

cmd/krew/cmd/list.go
```diff
@@ -24,11 +24,89 @@ import (
...
+func NewListFlags() *ListFlags {
+	outputFormat := ""
+	return &ListFlags{
+		JSONYamlPrintFlags: genericclioptions.NewJSONYamlPrintFlags(),
+		OutputFormat:       &outputFormat,
+	}
+}
+
+func (f *ListFlags) AllowedFormats() []string {
+	return f.JSONYamlPrintFlags.AllowedFormats()
```
corneliusweig: This should not only return `json`+`yaml`, but also `name` and `wide`.

JJJJJones: Its there just because it needs to be. I chose to leave AllowedFormats() relatively barren because if someone does have some reason for using ListFlags, I don't want to give the impression it handles name/wide.

cmd/krew/cmd/list.go
```diff
@@ -24,11 +24,89 @@ import (
...
+}
+
+func (f *ListFlags) ToPrinter() (printers.ResourcePrinter, error) {
```
corneliusweig: This function is a copy of `k8s.io/cli-runtime/pkg/genericclioptions/json_yaml_flags.go:44`. Why not
```go
func (f *ListFlags) ToPrinter() (printers.ResourcePrinter, error) {
	return f.JSONYamlPrintFlags.ToPrinter(*f.OutputFormat)
}
```

cmd/krew/cmd/list.go
JJJJJones: The version/kind is being used arbitrarily here and could just as easily be substituted with "_"s, which is what I did in the first place. @ahmetb suggested I use the constants.

cmd/krew/cmd/list.go
JJJJJones: Ostensibly there was some reason though its ~pretty much lost to time now~ I don't remember what it is right now. I'll switch it, worst case the commit history has it.

cmd/krew/cmd/list.go
JJJJJones: I kept OutputFormat a pointer for consistency with genericclioptions. This switch looks good to me, commit it?

cmd/krew/cmd/list.go
```diff
@@ -24,11 +24,89 @@ import (
...
+	case "json":
+		printer = &printers.JSONPrinter{}
+	case "yaml":
+		printer = &printers.YAMLPrinter{}
+	default:
+		return nil, genericclioptions.NoCompatiblePrinterError{OutputFormat: f.OutputFormat, AllowedFormats: f.AllowedFormats()}
+	}
+
+	return printer, nil
+}
```
corneliusweig: I'm still very uncomfortable that all this boilerplate code above this point is here in this file. IMO this is very easily testable and can also be reused by other modules. This file is for the `list` command, and should have a very high level abstraction. In contrast, the printing code has pretty low abstraction and does not fit well here.
As a result, you had to write a ton of integration tests for printing, because it was impossible to test the this independently of the `krew` machinery. This is not optimal.

Why do you think is this the best place for the printing details?

cmd/krew/cmd/list.go
JJJJJones: I made it this way to assist in readability; name, wide and json/yaml are all distinctly handled with other options throwing an error.

Also, I realize now this is why AllowedFormats() only lists yaml/json support.

JJJJJones: I'll move the ToPrinter up though.

cmd/krew/cmd/list.go
JJJJJones: > As a result, you had to write a ton of integration tests for printing, because it was impossible to test the this independently of the krew machinery

Printing is hardly my forte but wouldn't these integration tests need to exist regardless? Sorting should be covered (primarily) by a unit test but the integration test would implicitly verify that.

> Why do you think is this the best place for the printing details?

I do and I don't.

Ideally, much of the printing logic would be its own separate file `printer.go` that supports `krew info` as well (I made comments above semi-prototyping how that can go). Since that isn't done at this time, I just left the code in `list.go` to avoid confusing anyone into thinking its particularly re-usable. Would you like me to move it around?

ahmetb:
Sorry for delay. I am at kubecon, probably will review this next week.

integration_test/list_test.go
```diff
@@ -15,11 +15,16 @@
...
+	"encoding/json"
+	"path/filepath"
+	"regexp"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"gopkg.in/yaml.v2"
 )
 
+// FIXME: still necessary?
```
ahmetb: yeah i think we'd have a very preliminary "list" test and this can be it.

integration_test/list_test.go
```diff
@@ -42,3 +47,260 @@ func TestKrewList(t *testing.T) {
...
+
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	test.Krew("install",
+		"--manifest", filepath.Join("testdata", "foo.yaml"),
+		"--archive", filepath.Join("testdata", "foo.tar.gz")).
+		RunOrFail()
+
+	expected := []byte("[\n    {\n        \"Name\": \"foo\",\n        \"Version\": \"v0.1.0\"\n    }\n]\n")
```
ahmetb: why do you compare json output?
maybe just parse and cmp lists?

integration_test/list_test.go
```diff
@@ -42,3 +47,260 @@ func TestKrewList(t *testing.T) {
...
+		t.Fatalf("'list' output doesn't match:\n%s", diff)
+	}
+
+	var Plugin []struct {
+		Name    string
+		Version string
+	}
+
+	err := json.Unmarshal(eventualList, &Plugin)
+	if err != nil || Plugin[0].Name != "foo" || Plugin[0].Version != "v0.1.0" {
```
ahmetb: just do a slice cmp by having an `expected` slice?

integration_test/list_test.go
ahmetb: why aren't we outputting the index.Plugin object?

this bare `{"Name": ... , "Version": ...}` object is something new that doesn't exist in our ecosystem. let's just give them the index.Plugin as much as we can?

integration_test/list_test.go
```diff
@@ -42,3 +47,260 @@ func TestKrewList(t *testing.T) {
...
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	test.Krew("install", validPlugin).RunOrFail()
+	test.Krew("install",
+		"--manifest", filepath.Join("testdata", "foo.yaml"),
+		"--archive", filepath.Join("testdata", "foo.tar.gz")).
+		RunOrFail()
+
+	eventualList := test.Krew("list", "-o", "json").RunOrFailOutput()
```
ahmetb: -o=json output should NOT be an `[]` json array.

PTAL at `-o=json` for pods etc. it's a json stream.

corneliusweig: So with the latest stable `kubectl` I get for `kubectl get pod -ojson`:
```json
{
    "apiVersion": "v1",
    "items": [],
    "kind": "List",
    "metadata": {
        "resourceVersion": "",
        "selfLink": ""
    }
}
```
when the namespace is empty. And if there are pods, they are listed in the `items` field. This is _not_ a json stream, right? However, we should also have the wrapping list object.

However, I totally agree that a list of json objects like a top-level structure as `[{..}, {..}]` is completely wrong.

ahmetb: Actually you're right I missed that. It would be good to stick to that format.

integration_test/list_test.go
JJJJJones: I had no idea if it was wanted or not, so better to throw it in than not. Plus, it helped inform me.

integration_test/list_test.go
JJJJJones: We wanted confirmation that valid JSON was being output (it only seems sane at that point to also test the accuracy)

integration_test/list_test.go
JJJJJones: Shouldn't line 29 be removed then since its only ever run as part of a larger battery?

integration_test/list_test.go
JJJJJones: Also in YAML's(?) case the error result was always nil, regardless of failure.

integration_test/list_test.go
JJJJJones: >why aren't we outputting the index.Plugin object?

I'm not sure I understand what you mean, index.Plugin is output. Lines 69-77 then take it, confirm its valid JSON and that its values are correct.

integration_test/list_test.go
JJJJJones: https://github.com/kubernetes-sigs/krew/pull/365#pullrequestreview-312725338

An objects list was what was discussed previously. I can change it to an object with the list contained within it but that'd require some relatively significant changes.

JJJJJones: Output currently looks like this for -o json.

```json
[
    {
        "Name": "grep",
        "Version": "v1.2.0"
    },
    {
        "Name": "ingress-nginx",
        "Version": "v0.25.0"
    },
    {
        "Name": "krew",
        "Version": "v0.3.1"
    },
    {
        "Name": "view-utilization",
        "Version": "v0.2.2"
    }
]
```

-o yaml

```yaml
- Name: grep
  Version: v1.2.0
- Name: ingress-nginx
  Version: v0.25.0
- Name: krew
  Version: v0.3.1
- Name: view-utilization
  Version: v0.2.2
```

ahmetb: We can't really invent a new type for plugins.
We already have an `index.Plugin` type. I can't get behind introducing a format that is not going to be extensible and incompatible with the existing schema we have.

ahmetb: Also you don't need to use metav1.List whatever.
You can just fake it.
I think we might need to perhaps start over and not use genericclioptions for this.

JJJJJones: I don't use metav1.List?

ahmetb: But that's what's suggested above.
If we merge Krew into kubectl (a likely outcome) we'll need to adhere the standard format, so we can't take a change like this which we'd only have to break later on.

corneliusweig:
@JJJJJones It's really getting hard to follow what's going on in this PR. We have been going back and forth on various topics:

- how to test this
- how to structure the code
- what libraries to use for the job
- what are the goals of the design (just solving the `krew list` part or can we do `krew info` as well?)

It's impossible to get a clear look on this without taking a step back and rethink the approach here. So can you please write a design document about this change? That way we can discuss about the pros and cons of each approach without wasting the time to implement various drafts.

In particular, you should discuss those questions:

* What is the problem with the current state of krew?
* What is your proposed change?
* How does `kubectl` solve a similar problem?
* What is the structure of the output?
* What libraries can be used? (what are the pros and cons here)
* How can this be tested.
* What integration tests are necessary.

/hold

ahmetb:
> So can you please write a design document about this change?

+1. It doesn't have to be an actual doc or a full comprehensive design.

Let's go to #356 and do a whiteboarding of what we're adding, and what the end result (output format) etc are going to look like.

ahmetb:
Closing due to inactivity.
We're still open to pursuing this, but let's keep it a lot simpler.
/close

PR #367, opened 2019-11-08
Clean untracked files while updating index

### LINKED ISSUE #264: Proposal: clean untracked files while updating index
Currently `krew update` (when called directly or indirectly via `install` or
`upgrade` commands) runs the following commands to ensure a _clean_ local
copy, so that we don't have to nuke/re-clone a repo every time.

- `git fetch -v`
- `git reset --hard '@{upstream}'`

However, these commands do not clean up untracked files.
I'm not sure if the _untracked files_ cause any issues yet but it feels like
we can also run `git clean -xfd` to remove untracked files from the local copy
in the `pkg/gitutil#EnsureUpdated()`.

/kind proposal
/priority backlog

ferhatelmas:
fixes #264

### DIFF
--- a/pkg/gitutil/git.go
+++ b/pkg/gitutil/git.go
@@ -51,8 +51,11 @@ func update(destinationPath string) error {
 		return errors.Wrapf(err, "fetch index at %q failed", destinationPath)
 	}
 
-	err := exec(destinationPath, "reset", "--hard", "@{upstream}")
-	return errors.Wrapf(err, "reset index at %q failed", destinationPath)
+	if err := exec(destinationPath, "reset", "--hard", "@{upstream}"); err != nil {
+		return errors.Wrapf(err, "reset index at %q failed", destinationPath)
+	}
+	err := exec(destinationPath, "clean", "-xfd")
+	return errors.Wrapf(err, "clean index at %q failed", destinationPath)
 }
 
 // EnsureUpdated will ensure the destination path exists and is up to date.
--- a/pkg/gitutil/git.go
+++ b/pkg/gitutil/git.go
@@ -51,9 +51,12 @@ func update(destinationPath string) error {
 		return errors.Wrapf(err, "fetch index at %q failed", destinationPath)
 	}
 
-	if err := exec(destinationPath, "reset", "--hard", "@{upstream}"); err != nil {
-		return errors.Wrapf(err, "reset index at %q failed", destinationPath)
-	}
+	err := exec(destinationPath, "reset", "--hard", "@{upstream}")
+	return errors.Wrapf(err, "reset index at %q failed", destinationPath)
+}
+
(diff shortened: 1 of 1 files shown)


pkg/gitutil/git.go
```diff
@@ -51,8 +51,11 @@ func update(destinationPath string) error {
 		return errors.Wrapf(err, "fetch index at %q failed", destinationPath)
 	}
 
-	err := exec(destinationPath, "reset", "--hard", "@{upstream}")
-	return errors.Wrapf(err, "reset index at %q failed", destinationPath)
+	if err := exec(destinationPath, "reset", "--hard", "@{upstream}"); err != nil {
```
ahmetb: probably need to update godoc, and possibly the method name, right? 
this method now does certain things.

Also does this respect `--no-update-index` flags?

pkg/gitutil/git.go
```diff
@@ -51,8 +51,11 @@ func update(destinationPath string) error {
...
+		return errors.Wrapf(err, "reset index at %q failed", destinationPath)
+	}
+	err := exec(destinationPath, "clean", "-xfd")
```
corneliusweig: Argh.. there are no long options for `-xd` (only `--force/-f`). So can you add an explanatory comment what the git command does?

pkg/gitutil/git.go
ferhatelmas: put it under another function for clarity.

flag is already taken care in the call path.

pkg/gitutil/git.go
ferhatelmas: made it a separate function, I guess it's good enough now.

ahmetb: there was no need to make a new function IMO. there doesn't seem to be splitting these, esp. since we aren't testing them.

ferhatelmas: This is exactly what I thought (tests) but then did it anyway because it was somehow addressing two comments and we might add tests later.

ferhatelmas: Should I merge them back or good to keep?

ahmetb: Yeah might as well merge back.
that method is basically 3 execs.

ferhatelmas: Merged.

ferhatelmas: @corneliusweig since godoc is updated, is inline comment needed?

corneliusweig: @ferhatelmas The godoc should be enough :)

ferhatelmas:
@corneliusweig approval?

corneliusweig:
> @corneliusweig approval?

Most of the discussion was lead by @.ahmetb, so I didn't want to interfere too much :)

PR #370, opened 2019-11-09
Rely on standard utils for `IsSubPath`

corneliusweig:
* Add more test cases
* Fix misleading argument naming of `IsSubPath`
       - subPath -> basePath
       - path    -> subPath

### DIFF
--- a/pkg/pathutil/pathutil_test.go
+++ b/pkg/pathutil/pathutil_test.go
@@ -15,6 +15,7 @@
 package pathutil
 
 import (
+	"os"
 	"path/filepath"
 	"reflect"
 	"testing"
@@ -28,7 +29,7 @@ func TestIsSubPathExtending(t *testing.T) {
 	tests := []struct {
 		name            string
 		args            args
-		wantExtending   []string
+		wantExtending   string
 		wantIsExtending bool
 	}{
 		{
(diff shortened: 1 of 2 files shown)


pkg/pathutil/pathutil_test.go
```diff
@@ -55,17 +56,54 @@ func TestIsSubPathExtending(t *testing.T) {
...
 				extendingPath: filepath.Join("a"),
 			},
-			wantExtending:   nil,
-			wantIsExtending: false,
+		},
+		{
+			name: "base path is not clean",
+			args: args{
+				basePath:      filepath.Join("d", "..", "a", "b", "..", "..", "a"),
+				extendingPath: filepath.Join("a", "b"),
```
ahmetb: nit: test cases are hard to read
do you mind quickly replacing these with `filepath.FromSlash()`
maybe not needed so much I'm not sure

ahmetb:
/lgtm
/approve
/hold
feel free to /hold cancel as you wish if you wanna fix nit above

pkg/pathutil/pathutil_test.go
corneliusweig: Oh it's a good idea and easily done.

PR #373, opened 2019-11-10
RELEASING_KREW: Specify the repository to push the tag to

corneliusweig:
When pushing individual tags, specifying the repos is mandatory.

I also wasn't aware of this when reviewing #369 :face_with_head_bandage:

/assign @ahmetb


ahmetb:
Oh yeah I think I corrected this in the shell, but forgot to update docs.
/lgtm
/approve

Thanks for catching.

PR #376, opened 2019-11-11
Update golangci-lint to v1.21.0

ferhatelmas:
Incremental analysis is added.

### DIFF
--- a/hack/run-lint.sh
+++ b/hack/run-lint.sh
@@ -23,7 +23,7 @@ gopath="$(go env GOPATH)"
 if ! [[ -x "$gopath/bin/golangci-lint" ]]; then
   echo >&2 'Installing golangci-lint'
   curl --silent --fail --location \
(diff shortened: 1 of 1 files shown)


ferhatelmas:
There are some new linters; like whitespace (enforces using more whitespace). 

Should we evaluate them if we want to enable or not? To me, we have something stable and new ones have overlapping functionality with existing ones.

corneliusweig:
> To me, we have something stable and new ones have overlapping functionality with existing ones.

If I understand this comment correctly, you are saying that you don't see the need to enable more linters. Is that correct?

ferhatelmas:
Yes. 

In general, we can keep an eye for newcomers but for now, new ones are related to style and not needed, I think.

corneliusweig:
I see. Thanks for keeping this up-to-date!
/lgtm
/approve

PR #375, opened 2019-11-11
Use renameOrCopy for single-file move operation

ahmetb:
It looks like we can generalize copyDir func as "copy" because filepath.Walk
works on regular files, too. Making single file move use this renameOrCopy
method.

Also fixed a typo.

This should fix #357.

PR #377, opened 2019-11-11
Improve logic to determine if receipts migration is necessary

### LINKED ISSUE #374: receiptsmigration detection on windows is faulty if krew itself fails to install
If I try to install krew on windows and fail halfway and receipts/krew.yaml is not written, krew will end up in a situation that requires `kubectl krew system receipts-upgrade`.

This happens because we look at `store/` dir which is not empty. But in this case of a bug (like #357) `store/krew/` is actually an empty dir.

https://github.com/kubernetes-sigs/krew/blob/7bd396984863b9379d76b2bc020da78756bd7a54/pkg/receiptsmigration/migration.go#L42-L63

What if we instead looked at entries under `bin/` directory?

/area platform/windows
/priority P2

corneliusweig:
Fixes #374

### DIFF
--- a/integration_test/testutil_test.go
+++ b/integration_test/testutil_test.go
@@ -84,6 +84,10 @@ func setupKrewBin(t *testing.T, tempDir *testutil.TempDir) string {
 	if err := os.MkdirAll(binPath, 0755); err != nil {
 		t.Fatal(err)
 	}
+
+	// todo(corneliusweig): when the receipts migration logic is removed, this becomes obsolete
+	tempDir.Write("receipts/krew.notyaml", []byte("must be present for receipts migration check"))
+
 	if err := os.Symlink(krewBinary, filepath.Join(binPath, "kubectl-krew")); err != nil {
 		t.Fatalf("cannot link krew binary: %s", err)
 	}
--- a/pkg/receiptsmigration/migration.go
+++ b/pkg/receiptsmigration/migration.go
@@ -44,19 +44,12 @@ func Done(newPaths environment.Paths) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	store, err := ioutil.ReadDir(newPaths.InstallPath())
+	plugins, err := ioutil.ReadDir(newPaths.BinPath())
 	if err != nil {
 		return false, err
 	}
 
-	var hasInstalledPlugins bool
-	for _, entry := range store {
-		if entry.IsDir() {
-			hasInstalledPlugins = true
-			break
-		}
-	}
-
+	hasInstalledPlugins := len(plugins) > 0
 	hasNoReceipts := len(receipts) == 0
 
 	return !(hasInstalledPlugins && hasNoReceipts), nil
--- a/pkg/receiptsmigration/migration_test.go
+++ b/pkg/receiptsmigration/migration_test.go
@@ -39,7 +39,7 @@ func TestIsMigrated(t *testing.T) {
 	}{
 		{
 			name:         "One plugin and receipts",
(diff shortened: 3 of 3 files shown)


pkg/receiptsmigration/migration.go
```diff
@@ -44,14 +44,14 @@ func Done(newPaths environment.Paths) (bool, error) {
...
+	plugins, err := ioutil.ReadDir(newPaths.BinPath())
 	if err != nil {
 		return false, err
 	}
 
 	var hasInstalledPlugins bool
-	for _, entry := range store {
-		if entry.IsDir() {
+	for _, entry := range plugins {
+		if strings.HasPrefix(entry.Name(), "kubectl-") && 0 != entry.Mode()&os.ModeSymlink {
```
ahmetb: what if we didn't look for kubectl- prefix?
similarly, what if we didn't look for them to be symlinks?
we can just ReadDir?

ahmetb:
mostly lgtm
/hold
for comments

pkg/receiptsmigration/migration.go
corneliusweig: That should be fine, unless somebody puts something into the `bin` directory. Doing these checks  makes that much more unlikely. But this should matter in very exceptional cases only.

ahmetb: Yeah I guess it's especially fine because we'll remove this code ≥0.5 or maybe earlier.

ahmetb:
/lgtm
/approve

We can have this + #375 go into a new release, as both of them address Windows installation issues.

corneliusweig:
I simplified the logic as suggested. That way, the tests don't need to be modified as much, which is better.

---

> We can have this + #375 go into a new release, as both of them address Windows installation issues.

Yes, makes sense. Let's wait for confirmation that the copy/move issue is indeed resolved.

pkg/receiptsmigration/migration_test.go
```diff
@@ -39,7 +39,7 @@ func TestIsMigrated(t *testing.T) {
 	}{
 		{
 			name:         "One plugin and receipts",
-			filesPresent: []string{"store/konfig/konfig.sh", "receipts/present"},
+			filesPresent: []string{"bin/konfig/konfig.sh", "receipts/present"},
```
ahmetb: this is not very realistic since it created dirs under bin/
let's just ad /konfig.sh 

also if "konfig" the word isn't relevant, let's say "foo", (ditto below)

but it seems like this has simplified the tests a good deal.

corneliusweig: Yes, makes much more sense!

ahmetb:
/lgtm
/approve
thanks!

ahmetb:
/retitle Improve logic to determine if receipts migration is necessary

integration_test/testutil_test.go
```diff
@@ -84,6 +84,7 @@ func setupKrewBin(t *testing.T, tempDir *testutil.TempDir) string {
 	if err := os.MkdirAll(binPath, 0755); err != nil {
 		t.Fatal(err)
 	}
+	tempDir.Write("receipts/krew.notyaml", []byte("must be present for receipts migration check"))
```
corneliusweig: This is a bit hacky. But the new logic requires that _something_ is present under `receipts/`.

corneliusweig:
@ahmetb If we merge this, we can craft a bugfix release for windows, right?

ahmetb:
I think so.

ahmetb:
> If we merge this, we can craft a bugfix release for windows, right?

Let's try to get other low-hanging small PRs in for a release too! 👍

PR #382, opened 2019-11-13
Integrate archiver to simplify download

### LINKED ISSUE #371: Move tar and zip extraction into a reusable package
/kind cleanup
/priority P3

As shown in [in the security advisory](https://github.com/kubernetes-sigs/krew/security/advisories/GHSA-m9g4-7496-9p6f) it's hard to reason about TAR extraction and more importantly: do the job securely.

Let's move the tar + zip extraction code out of pkg/downloader to its own package somewhere so we can use it in other projects.

It would be great to have some of the options to these methods configurable, this would also make tests easier, such as objects like TarGzExtractOptions and ZipExtractOptions with flags like:

- skip verifying secure paths (default: false)
- (for tarballs) skip pax header (default: true)
- ... what else?

ferhatelmas:
Right now, archiver doesn't handle security issue
but there is an existing PR to handle it.
Added walker to handle until upstream is fixed.

fixes #371


corneliusweig:
@ferhatelmas Can you link the PR you are talking about?

ferhatelmas:
sure, https://github.com/mholt/archiver/pull/169

corneliusweig:
Hmm.. strange. The repo seems pretty active but this CVE fix is from May 12. What's wrong there?

ferhatelmas:
I guess the owner was busy with caddy 2.0 release and there is no other maintainer.

pkg/download/downloader.go
```diff
@@ -170,44 +60,38 @@ func suspiciousPath(path string) error {
...
-	}
+func isSuspiciousArchive(path string) error {
+	return archiver.Walk(path, func(f archiver.File) error {
+		switch h := f.Header.(type) {
+		case *tar.Header:
+			return suspiciousPath(h.Name)
+		case zip.FileHeader:
+			return suspiciousPath(h.Name)
+		default:
+			return errors.Errorf("Unknow header type: %T", h)
```
ahmetb: typo

ahmetb:
Can we please preserve signatures of something like `extractZIP` and `extractTARGZ` and/or write interface wrappers around the mholt/archiver package please?

I know this is not super necessary but for example see how we have wrappers around similar packages such as `semver` package and `yaml` packages. 

This is a pretty critical dependency so I'd love to have an option to flip between implementations if it comes to that.

corneliusweig:
I'm a bit worried that the maintainer has not managed to merge a proposed fix for a CVE in half a year.

Besides, `archiver` just unpacks and creates symlinks:
https://github.com/mholt/archiver/blob/6041b49af61de1075782610aaee8c6c53ba73c45/zip.go#L207-L215
and
https://github.com/mholt/archiver/blob/6041b49af61de1075782610aaee8c6c53ba73c45/tar.go#L232-L245

Thus, it's a piece of cake to escape from the target directory and write to arbitrary paths (you don't even have to chain symlinks).

To me it clearly looks like `archiver` was built with ease in mind, but not security. It's also not possible to configure its security level, so I'm afraid we can't use `archiver`.

pkg/download/downloader.go
erain: Also: Error strings should not be capitalized
Ref: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

corneliusweig:
@erain What is your opinion on the symlink handling in `archiver`?

erain:
@corneliusweig You are talking about the security implication in https://github.com/kubernetes-sigs/krew/issues/359 right?

I am not the expert on the topic. Some research showed that `archiver` has symlink support built-in: https://github.com/mholt/archiver/pull/92

I am not sure about whether the symlink handling is ideal (e.g. only relative symlink and not go beyond the base dir) yet. But we can do more research here.

ahmetb:
I vote -1 for using archiver. It’s not in a good state. We seem to be caring about security more. And we have open PRs like #360 which will rely on our ability to correctly and securely implement symlinks.

erain:
Good to know. Sorry for missing the context on this.

corneliusweig:
@erain no worries at all. There is a history about escaping the unpack directory, so that is something to keep an eye out for.

corneliusweig:
/close
In the current state of `archiver` it does not meet our security requirements.

corneliusweig:
@ferhatelmas Thanks though for investigating and integrating into krew. Maybe there is another better alternative?

ahmetb:
For now, I recommend we consider creating an interface, and extracting unarchive functionality to its new pkg that we can more easily test/maintain.

We can even make tar/zip methods smaller as there are a lot of commonalities during extraction (mkdir, write file, symlink...).

Let's just start with simply moving the code, and keep PRs small/reviewable.

ferhatelmas:
> Thanks though for investigating and integrating into krew.

PR #386, opened 2019-11-14
Make logging more consistent for downloaded vs. local archives

corneliusweig:
This is a follow-up for #361 . I didn't realize that the logging was equally misleading.

Also see https://github.com/kubernetes-sigs/krew/pull/375#issuecomment-553653013

### DIFF
--- a/pkg/download/downloader.go
+++ b/pkg/download/downloader.go
@@ -35,11 +35,11 @@ import (
 func download(url string, verifier Verifier, fetcher Fetcher) (io.ReaderAt, int64, error) {
 	body, err := fetcher.Get(url)
 	if err != nil {
-		return nil, 0, err
+		return nil, 0, errors.Wrapf(err, "failed to obtain plugin archive")
 	}
 	defer body.Close()
 
-	glog.V(3).Infof("Reading archive into memory")
+	glog.V(3).Infof("Reading archive file into memory")
 	data, err := ioutil.ReadAll(io.TeeReader(body, verifier))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/download/downloader.go
```diff
@@ -33,26 +33,25 @@ import (
 // download gets a file from the internet in memory and writes it content
 // to a Verifier.
 func download(url string, verifier Verifier, fetcher Fetcher) (io.ReaderAt, int64, error) {
-	glog.V(2).Infof("Fetching %q", url)
 	body, err := fetcher.Get(url)
 	if err != nil {
-		return nil, 0, errors.Wrapf(err, "could not download %q", url)
+		return nil, 0, err
```
ahmetb: removing wrapping makes me nervous. :| :|

pkg/download/downloader.go
```diff
@@ -33,26 +33,25 @@ import (
...
 	}
 	defer body.Close()
 
-	glog.V(3).Infof("Reading download data into memory")
+	glog.V(3).Infof("Reading archive into memory")
```
ahmetb: Reading archive file [...]

pkg/download/downloader.go
ahmetb: what if we said: `failed to obtain plugin archive`

ahmetb:
mostly good. small nits and I'll lgtm

pkg/download/downloader.go
corneliusweig: sounds good

PR #387, opened 2019-11-14
Refactor: move errorFetcher struct to test file

corneliusweig:
This fetcher was only used in its test.


ahmetb:
It actually helps us have code coverage around error cases in downloader.
That's why it's there. :)
/hold

ahmetb:
In other words, we ensure the errors from downloader implementation are returned correctly.

corneliusweig:
Whoops .. sorry I wasn't thinking straight. Thanks for catching this. I do think however, that it should have been in the test file.
/retitle Refactor: move errorFetcher struct to test file

Issue #388, opened 2019-11-15
Drop GOPROXY default


ferhatelmas:
/kind cleanup

Since their explicitly set values are reitaration of defaults.

https://golang.org/doc/go1.13#modules

`GO111MODULE`:
> The GO111MODULE environment variable continues to default to auto, but the auto setting now activates the module-aware mode of the go command whenever the current working directory contains, or is below a directory containing, a go.mod file — even if the current directory is within GOPATH/src.

`GOPROXY`:
> The GOPROXY environment variable may now be set to a comma-separated list of proxy URLs or the special token direct, and its default value is now https://proxy.golang.org,direct.

ferhatelmas:
If you think this is a good thing to do (still can prefer being explicit), I will send a PR.

ahmetb:
I think we retain GO111MODULE explicitly because travis sets GOPATH and therefore tries to use the old ways. 

GOPROXY can be removed.

ferhatelmas:
OK, let me drop proxy in one PR and we can discuss module separately because there are multiple references to it.

corneliusweig:
@ferhatelmas Can this be closed now?

ferhatelmas:
I think no because according to the first sentence of release notes, it doesn't matter if source under GOPATH or not. 
 
> The GO111MODULE environment variable continues to default to auto, but the auto setting now activates the module-aware mode of the go command whenever the current working directory contains, or is below a directory containing, a go.mod file — even if the current directory is within GOPATH/src.

That's why I think we should drop it too but since it's referenced more, I wanted to split PRs. What do you think?

corneliusweig:
@ferhatelmas Let's try it out :)

ahmetb:
It seems like Travis now sets GO111MODULE=auto and also set GOPATH.
https://docs.travis-ci.com/user/languages/go/

This is weird and I'm not sure why it's assuming =off , because GOPATH is not supposed to be set for =on.

corneliusweig:
Am I the only one who is confused here? I'm beginning to think that it is more apparent and obvious if we re-state the default settings for `GO111MODULE` and `GOPROXY`.
That seems to leave the smallest room for interpretation.

PR #389, opened 2019-11-15
Drop GOPROXY since same with default

ferhatelmas:
related to #388.


ferhatelmas:
I formatted yaml too but it might be done separately or should be reverted (might be our desired format)?

corneliusweig:
The original version is what the `travis-cli` produces. This will avoid further formatting changes in the future, so it would be best if you can make the change without re-formatting in a new way.

ferhatelmas:
OK, reverted formatting.

ahmetb:
I'm not at all sure why but this speeds up builds by 20 seconds.

old:

image


new:

image


/lgtm
/approve

ferhatelmas:
> I'm not at all sure why but this speeds up builds by 20 seconds.

Recent builds are again back to slow speed. I also tried to find an explanation but I couldn't come up with a doc. I guess it was just network variance :open_mouth:

PR #393, opened 2019-11-16
Fail validation when manifest has line break in `shortDescription`

### LINKED ISSUE #391: line wrapped on "kubectl search" output
Pretty weird. Seems to be reproing right now:

```
ssh-jump                        A kubectl plugin to SSH into Kubernetes nodes u...  no
sudo                            Run Kubernetes commands impersonated as group s...  no
support-bundle                  Creates support bundles for off-cluster analysis    no
tail                            Stream logs from multiple pods and containers u...  no
view-secret                     Decode Kubernetes secrets
                                no
view-serviceaccount-kubeconfig  Show a kubeconfig setting to access the apiserv...  no
view-utilization                Shows cluster cpu and memory utilization            no
virt                            Control KubeVirt virtual machines using virtctl     no
warp                            Sync and execute local files in Pod                 no
```

image

/kind bug
/priority P1

corneliusweig:
Fixes #391

### DIFF
--- a/pkg/index/validation/validate.go
+++ b/pkg/index/validation/validate.go
@@ -79,6 +79,9 @@ func ValidatePlugin(name string, p index.Plugin) error {
 	if p.Spec.ShortDescription == "" {
 		return errors.New("should have a short description")
 	}
+	if strings.ContainsAny(p.Spec.ShortDescription, "\r\n") {
+		return errors.New("should not have line breaks in short description")
+	}
 	if len(p.Spec.Platforms) == 0 {
 		return errors.New("should have a platform specified")
 	}
--- a/cmd/krew/cmd/search.go
+++ b/cmd/krew/cmd/search.go
@@ -96,6 +96,7 @@ Examples:
 }
 
 func limitString(s string, length int) string {
+	s = strings.TrimSpace(s)
 	if len(s) > length && length > 3 {
 		s = s[:length-3] + "..."
 	}
--- a/cmd/krew/cmd/search.go
+++ b/cmd/krew/cmd/search.go
@@ -96,7 +96,6 @@ Examples:
 }
 
 func limitString(s string, length int) string {
-	s = strings.TrimSpace(s)
 	if len(s) > length && length > 3 {
 		s = s[:length-3] + "..."
 	}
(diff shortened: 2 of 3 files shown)


ahmetb:
I assume we’re still susceptible to \n though. 
What if Plugin Developer enters multiline string?
Should we disallow \n altogether?

corneliusweig:
Good point. I think we _should_ catch multi-line `shortDescription` during manifest reviews. To be absolutely certain, we should add some validation for that.

However, when thinking in the direction of multi-index, it would be nice to also be fault-tolerant there, because we cannot force downstream indexes to use our validation tool. So I would suggest to do both: validating that `shortDescription` does not contain newlines _and_ replace all newlines by spaces.

corneliusweig:
Oh.. when we reject `shortDescription` with line breaks, this will also work for multi-index krew, because they need to have valid manifests. So that sounds like to proper thing to do.

This means that `view-secret` is no longer a valid manifest and https://github.com/kubernetes-sigs/krew-index/pull/342 needs to be merged too.

corneliusweig:
@ahmetb So this fails because `who-can` also has a trailing new-line (we never noticed because of the ellipsis).

What do you think about the general plan to forbid newlines in `shortDescription`? Then let's remove the newline in `who-can.yaml`, too. Or should we be more lenient with trailing `\n`?

ahmetb:
The cleanest is to prevent \n overall. 
Let’s do it in index validator. 
I think soon we should split structural validity vs cosmetic validity.

corneliusweig:
Ok, then let me fix this in the index and then I'll re-trigger this.

corneliusweig:
Alright, CI is all green now.

pkg/index/validation/validate.go
```diff
@@ -79,6 +79,9 @@ func ValidatePlugin(name string, p index.Plugin) error {
 	if p.Spec.ShortDescription == "" {
 		return errors.New("should have a short description")
 	}
+	if strings.ContainsAny(p.Spec.ShortDescription, "\r\n") {
```
ahmetb: Do you know if we call this on the parse path?

If so it might cause an issue for people who run on this version of krew with an older index. I am not sure how/if that’d happen. Wdyt?

corneliusweig: :face_with_head_bandage: You are totally right. This _is_ called in the parse path. However, I think we are fine here:

* the new krew version will only be installed when the index was updated as well. The only way one could break it, would be by purpose, if somebody manually downloads a newer krew version and tries to work with an old index version.
* old krew versions might have a newer index. However that doesn't really matter, because the new index will also be valid for the old krew version.

PR #396, opened 2019-11-18
Drop GO111MODULE env variable

### LINKED ISSUE #388: Drop GOPROXY default

ferhatelmas:
Drop obsolete GO111MODULE environment variable.

Also, dry tool installation a bit.

fixes #388.


ferhatelmas:
How about release-

hack/install-gox.sh
```diff
@@ -12,31 +12,18 @@
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
-# limitations under the License.
+# limitations under the License
```
corneliusweig: That missing period breaks the boilerplate check ;)

ferhatelmas: thanks :smile:

corneliusweig:
@ferhatelmas thanks for doing this. Could you split this PR in two separate ones? One for the `GO111MODULE` change and one only for the dry improvements?

hack/install-gox.sh
```diff
@@ -24,12 +24,10 @@ cleanup() {
 trap cleanup EXIT
 
 install_gox() {
-  gobin="$(go env GOPATH)/bin"
   cd "${scratch_dir}"
-  env GOPATH="${scratch_dir}" \
```
corneliusweig: Ah.. now I understand why you originally had the DRY patch in this PR. For the install helpers, I would not necessarily have removed the `GO111MODULE` env var, though.

But since you did it, can you also add a version tag to each installed tool?

hack/install-gox.sh
```diff
@@ -24,12 +24,10 @@ cleanup() {
...
-    GO111MODULE=off \
-    GOBIN="${gobin}" \
```
ahmetb: when you remove GOBIN, how do you ensure gox ends up in PATH?

Is the "$(go env GOPATH)/bin" (which is probably /home/travis/go/bin) added to PATH already?

hack/install-gox.sh
ferhatelmas: Sure, it's a better idea.

hack/install-gox.sh
ferhatelmas: yes.

hack/install-gox.sh
ferhatelmas: Added versions.

ahmetb:
I think this is good now.
There was a time when if you explicitly didn't enable go.mod, it wouldn't use it on Travis. Looks like they’ve improved things since go1.11. :)

corneliusweig:
> Generally, tools(golangci-lint, impi, etc.) are installed under global GOBIN but release-notes is installed under hack.

I think we can move `release-notes` out of the hack folder. I didn't want to clutter the system, but :man_shrugging: .

/lgtm
/approve

PR #397, opened 2019-11-18
Drop GZIP env in artifact building

ferhatelmas:
> warning: GZIP environment variable is deprecated; use an alias or script

handle warning with `-I / --use-compress-program=PROG`

### DIFF
--- a/hack/make-release-artifacts.sh
+++ b/hack/make-release-artifacts.sh
@@ -46,7 +46,7 @@ done < <(find "${bin_dir}" -print0)
 echo >&2 "Creating ${krew_tar_archive} archive."
 (
   cd "${bin_dir}"
-  tar -I "gzip --best" -cvf "${SCRIPTDIR}/../out/${krew_tar_archive}" ./*
+  tar --use-compress-program "gzip --no-name" -cvf "${SCRIPTDIR}/../out/${krew_tar_archive}" ./*
 )
 
 checksum_cmd="shasum -a 256"
--- a/hack/make-release-artifacts.sh
+++ b/hack/make-release-artifacts.sh
@@ -46,7 +46,7 @@ done < <(find "${bin_dir}" -print0)
 echo >&2 "Creating ${krew_tar_archive} archive."
 (
   cd "${bin_dir}"
-  tar -I "gzip -${n}" -cvf "${SCRIPTDIR}/../out/${krew_tar_archive}" ./*
+  tar -I "gzip --best" -cvf "${SCRIPTDIR}/../out/${krew_tar_archive}" ./*
 )
 
 checksum_cmd="shasum -a 256"
--- a/hack/make-release-artifacts.sh
+++ b/hack/make-release-artifacts.sh
@@ -46,7 +46,7 @@ done < <(find "${bin_dir}" -print0)
 echo >&2 "Creating ${krew_tar_archive} archive."
 (
   cd "${bin_dir}"
-  GZIP=-n tar czvf "${SCRIPTDIR}/../out/${krew_tar_archive}" ./*
+  tar -I "gzip -${n}" -cvf "${SCRIPTDIR}/../out/${krew_tar_archive}" ./*
 )
 
 checksum_cmd="shasum -a 256"


hack/make-release-artifacts.sh
```diff
@@ -46,7 +46,7 @@ done < <(find "${bin_dir}" -print0)
 echo >&2 "Creating ${krew_tar_archive} archive."
 (
   cd "${bin_dir}"
-  GZIP=-n tar czvf "${SCRIPTDIR}/../out/${krew_tar_archive}" ./*
+  tar -I "gzip -${n}" -cvf "${SCRIPTDIR}/../out/${krew_tar_archive}" ./*
```
corneliusweig: I think the `-n` is just an option and should not be read as an env var:
```suggestion
  tar -I "gzip -n" -cvf "${SCRIPTDIR}/../out/${krew_tar_archive}" ./*
```

Can you also add a `--best` gzip option?

corneliusweig: Also, we need a tester with an OSX system (which uses BSD tar). The dash options are just GNU tar and it could be possible that your change breaks the BSD tar.

ferhatelmas: Right, updated.

ahmetb:
could --best be impacting reproducibility of the archive file when executed multiple times?

hack/make-release-artifacts.sh
```diff
@@ -46,7 +46,7 @@ done < <(find "${bin_dir}" -print0)
 echo >&2 "Creating ${krew_tar_archive} archive."
 (
   cd "${bin_dir}"
-  GZIP=-n tar czvf "${SCRIPTDIR}/../out/${krew_tar_archive}" ./*
+  tar -I "gzip --best" -cvf "${SCRIPTDIR}/../out/${krew_tar_archive}" ./*
```
ahmetb: Definitely opt in for longer flag alias when you can, for readability.

ferhatelmas: sure, good idea.

ferhatelmas:
> could --best be impacting reproducibility of the archive file when executed multiple times?

OK but why are we closing PR? 

We can refactor existing implementation `gzip -n` and handle the warning, needs MAC testing though.

ahmetb:
Sorry I am mobile nowadays. I must’ve closed accidentally. 

I am ok with changing -n flag but we might not need —best, I recall there was an issue with  reproducibility of the archive back in the day with it.

hack/make-release-artifacts.sh
corneliusweig: There should also be
```suggestion
  tar -I "gzip --best --no-name" -cvf "${SCRIPTDIR}/../out/${krew_tar_archive}" ./*
```

This is what guarantees the reproducibility. AFAIK the `--best` option should satisfy reproducibility.

corneliusweig:
> could --best be impacting reproducibility of the archive file when executed multiple times?

I think for reproducibility we just need to enable `-n / --no-name` again.

ahmetb:
I just got curious: what was -n/--no-name doing for us?
We already set timestamps of the archived files to `0`.
What does "not saving the original file name" mean here?

corneliusweig:
Good question! I think the problem is that `gzip` takes the name and timestamp of the original file. However, when creating a `tar.gz` file, what happens is that tar is piped to `gzip`. And when gzip finds a pipe, it seems to always use the current timestamp.

Bottom line is, that we want to have both:

- defined timestamps of the files in the tar
- _and_ the `--no-name` option when creating the archive.

ferhatelmas:
Dropped `--best`since it introduces another functionality which isn't the target of the PR.

Used longer flags.

hack/make-release-artifacts.sh
```diff
@@ -46,7 +46,7 @@ done < <(find "${bin_dir}" -print0)
 echo >&2 "Creating ${krew_tar_archive} archive."
 (
   cd "${bin_dir}"
-  tar -I "gzip --best" -cvf "${SCRIPTDIR}/../out/${krew_tar_archive}" ./*
+  tar --use-compress-program "gzip --no-name" -cvf "${SCRIPTDIR}/../out/${krew_tar_archive}" ./*
```
corneliusweig: You dropped the `--best` option. Was that intentional?

ahmetb: I suspect it was because of my comment about reproducibility.

ferhatelmas: I think we can add it but I prefer it adding separately because

corneliusweig:
Thanks for your persistence!

PR #404, opened 2019-11-25
Prefer the best compression for artifacts

ferhatelmas:
As discussed in https://github.com/kubernetes-sigs/krew/pull/397.

### DIFF
--- a/hack/make-release-artifacts.sh
+++ b/hack/make-release-artifacts.sh
@@ -46,7 +46,7 @@ done < <(find "${bin_dir}" -print0)
 echo >&2 "Creating ${krew_tar_archive} archive."
 (
   cd "${bin_dir}"
-  tar --use-compress-program "gzip --no-name" -cvf "${SCRIPTDIR}/../out/${krew_tar_archive}" ./*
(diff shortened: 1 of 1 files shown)


ahmetb:
/lgtm
/approve
/hold

@ferhatelmas do you get the same checksum from the tarball deterministically when you specify --best and re-encrypt the same file over and over again?

We need reproducibility, for example I should be able to tar the same code on my machine and get the same results with GNU tar (same version as CI). Adding --best might complicate that. 

But we can try. 
Can you also note what % bytes does this compression save?

corneliusweig:
@ferhatelmas would you care about doing some experiments about reproducibility and compression ratio as suggested above?

ahmetb:
Let's stop pursuing this. I see <1% better compression rate, and potentially issues about reproducibility.

```
-rw-r--r--  1 ahmetb primarygroup 17058579 Dec  8 23:16 krew.tar.gz
-rw-r--r--  1 ahmetb primarygroup 16932259 Dec  8 23:16 krew.tar.gz
```

/close

ferhatelmas:
Sorry, I was late to produce any numbers. Closing sounds good to me. 

However. I don't really understand the reproducibility issue intuitively because it's just another level. If it is fine at 6, should be fine at 9 to me. 

Maybe, I need to read man pages of tar/gzip to understand this issue.

PR #422, opened 2019-12-03
Move implementation details to internal

### LINKED ISSUE #291: Hide internal packages from external users
https://godoc.org/sigs.k8s.io/krew shows that pretty much all our packages
are available for public consumption, despite we only use them in Krew. Perhaps
pkg/index and pkg/constants are exempt from this.

If there's a way to hide these, through something like "internal" go packages,
I'd be in favor of implementing it.

/kind cleanup
/priority P3

ferhatelmas:
move all packages into `internal` except `pkg/index` and `pkg/constants`.

fixes #291


cmd/generate-plugin-overview/go.sum
```diff
@@ -122,6 +122,7 @@ sigs.k8s.io/krew v0.2.2-0.20190724210953-8867a94acd6d h1:ftkzey0SeGithDvmwEsUnlD
 sigs.k8s.io/krew v0.2.2-0.20190724210953-8867a94acd6d/go.mod h1:wsheYhp5znrUApov8QcNjBCwXu85fIGjRJvjSkE8pI8=
 sigs.k8s.io/krew v0.3.2 h1:FKBD9xy9GtO5A8FeVk/cbgG6vzPbUg4SwUziZj/TGoU=
 sigs.k8s.io/krew v0.3.2/go.mod h1:G0cvdJtBb0gHBQZtCBumtllvhU6H+ggDr2oLznaLglc=
+sigs.k8s.io/krew v0.3.3 h1:cHv2oReJmFmR0CnT9Fs1c9JCAOC6OBpdq4xKEJis+2E=
```
ferhatelmas: why do we have separate go.mod files ?

ferhatelmas: This diff is intentional to discuss this issue otherwise, it shouldn't be a part of this PR.

ahmetb: This is intentional.
The stuff under cmd/* are essentially like different repos that have separate set of dependencies and they rely on parts of krew.

ahmetb:
HUGE if we can get it in ASAP before it starts introducing merge conflicts everywhere :)

cmd/generate-plugin-overview/go.sum
ferhatelmas: But they will move further altogether, right? We won't release them separately.

ferhatelmas: Right now, it's only added for `generate-plugin-overview` cmd, I suggest dropping it.

ahmetb: Yeah, they move together. We can make them part of the main repo, by removing go.mod.
@corneliusweig wdyt?

ahmetb: BTW I think the main reason these cmd/* tools have their own go.mod is mostly so that we don't download _their_ dependencies while building krew.

ahmetb:
I'll go ahead with this, seeing no issues locally.

I'll send a PR to move cmd/* commands to v0.3.3 later.

cmd/generate-plugin-overview/go.sum
ferhatelmas: I see your point but no extra dependency so it is somehow premature optimization.

ferhatelmas: I sent https://github.com/kubernetes-sigs/krew/pull/424 to remove `go.mod`.

corneliusweig: Yeah, make sense. It also reduces the maintenance overhead of keeping the child module in sync with the current krew version :+1:

PR #424, opened 2019-12-04
Remove sub go.mod from cmd

ferhatelmas:
remove cmd/* gomod since they don't bring the benefits they were supposed to bring.

context of how it was realized: https://github.com/kubernetes-sigs/krew/pull/422#discussion_r353475794


ahmetb:
Actually it helps us split dependencies between krew and these tools. It’s not the worst idea to keep these.

corneliusweig:
Interesting. Our linter doesn't find this issue when the `go.{mod,sum}` files in the submodule are present.

I think the biggest reason for removing the submodule is that we don't need to bother about keeping the krew version in sync with the main module there. However, we could also achieve that with a `replace` directive in `go.mod`.

On the other hand, it's nice that we can add dependencies to the submodules without poisoning the main krew module.

---

Overall, I'm slightly in favor of keeping the status quo. Just because it improves the separation of krew and its associated utils. But I'm open for other improvements which reduce the maintenance overhead.

ahmetb:
Technically, the go.mod at the root should be requiring more pkgs now that the child go.mod's are deleted. But I don't see that happening, weird.

I feel like within travis it might be editing go.mod (during `go mod download`, or `go test ./...`). I wish we could easily globally set `-mod=readonly`, so we don't add new deps on the fly.

ahmetb:
I guess indeed these two tools don't really add new deps to Krew.
So let's go ahead with this.

ferhatelmas:
> Our linter doesn't find this issue when the go.{mod,sum} files in the submodule are present.

Following is from go modules wiki:
> go test ./... from the repository root will no longer test everything in the repository

Similarly, `go/analysis` (underlying package leveraged by golangci-lint) stops too when `./...` is used or target is left out.

> we could also achieve that with a replace directive in go.mod.
Still, for a release, it needs a version bump. 

> I guess indeed these two tools don't really add new deps to Krew.
Exactly. 

> So let's go ahead with this.
Sounds good, I will check what is failing.

ahmetb:
```
cmd/generate-plugin-overview/generate-plugin-overview.go:111:2: ifElseChain: rewrite if-else to switch statement (gocritic)
	if matches := githubRepoPattern.FindStringSubmatch(homepage); matches != nil {
	^
```

stuff like these are intrusive. if we can disable these sometime, I really appreciate. It does the same for simple slice allocations without count specified, there's like no point.

ferhatelmas:
> if we can disable these sometime, I really appreciate. It does the same for simple slice allocations without count specified, there's like no point.

Ok, I will make a separate pass for it.

ahmetb:
/lgtm
/approve

can you please make sure krew-index CI still works in terms of fetching these tools?

corneliusweig:
> >we could also achieve that with a replace directive in go.mod.
>
> Still, for a release, it needs a version bump.

Mental note: need to understand this better.

ahmetb:
Also I think the `replace` directives are meant to be checked in _only_ if they are used to "strictly pin" to a strict commit of a dependency.

Stuff like `replace sigs.k8s.io => ../..` should not be checked in if the module is being go-get'ed (like these cmd/* tools).

ferhatelmas:
> can you please make sure krew-index CI still works in terms of fetching these tools?

Looks good.

PR #425, opened 2019-12-04
Run gofmt via golangci-lint

ferhatelmas:
Drop custom script since golangci-lint can already do it.


.golangci.yml
```diff
@@ -24,6 +26,7 @@ run:
 
 linters:
   enable:
+    - gofmt
```
corneliusweig: I think we don't even need this. Looking at `golangci-lint linters | grep goimports`:
```
goimports: Goimports does everything that gofmt does. Additionally it checks unused imports [fast: true, auto-fix: true]
```

ahmetb:
> I find the latter a little easier to understand, but it's not a big deal.

Agreed.

.golangci.yml
ferhatelmas: not exactly correct.

goimports version of golangci doesn't do simplification.

ferhatelmas: here is the configuration: https://github.com/golangci/golangci-lint/blob/master/pkg/config/config.go#L139

maybe we should suggest adding it then goimports can be used solely.

corneliusweig: Mm.. looks like goimports does no simplifications: https://github.com/golang/go/issues/21476

corneliusweig:
So I guess this is indeed the best simplification for `gofmt`. From my side
/lgtm

.golangci.yml
ferhatelmas: There was a long standing demand, I thought it was already added but it seems it's still missing.

ferhatelmas:
@ahmetb I believe this is waiting for your review/approval.

PR #427, opened 2019-12-05
skip/fail upgrade if no longer matches a platform

### LINKED ISSUE #423: False upgrade success message if an installed plugin no longer matches a platform spec
Discovered through #421

In case of:

> * A user installed a plugin, and its platform is removed from the corresponding manifest.

For example, I removed `darwin` platform of `ctx` plugin and ran:

```
$ kubectl krew upgrade ctx --no-update-index  -v=10
I1203 15:16:51.666127   66174 upgrade.go:81] --no-update-index specified, skipping updating local copy of plugin index
I1203 15:16:51.666209   66174 scanner.go:82] Reading plugin "ctx"
I1203 15:16:51.667079   66174 upgrade.go:65] Upgrading plugin: ctx
I1203 15:16:51.667286   66174 platform.go:43] Matching platform for labels(arch=amd64,os=darwin)
Upgraded plugin: ctx
```

It definitely didn't do an upgrade, but prints a message saying so.
/kind bug

I think we should skip upgrade (no fail) with reason like (plugin doesn't have distribution for platform $GOOS/$GOARCH)

ahmetb:
It turns out when upgrade process looks to a plugin manifest that exists
but does no longer have a matching Platform spec, we were returning nil error,
which was surfacing as "upgraded" (see #423).

This addresses the problem by returning an error in this case. However, the
'upgrade' command fails with this error if a plugin list is given to the cmd.

If 'krew upgrade' is called without args (i.e. upgrade all plugins), ensuring
this doesn't fail the cmd, but prints warnings (just like current 'Skipping'
behavior this cmd has).

I think this 'skipping with a warning' behavior is justified because it can
make upgrade command entirely malfunctional when a plugin manifest is updated
breaking the upgrades for that plugin.

(There are still some cases that make 'upgrade' command fatally fail, such as
an installed plugin gone from manifest, which we should address separately.)

Adding integration tests to capture both cases.

Fixes #423.

### DIFF
--- a/integration_test/upgrade_test.go
+++ b/integration_test/upgrade_test.go
@@ -17,6 +17,7 @@ package integrationtest
 import (
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"sigs.k8s.io/krew/pkg/constants"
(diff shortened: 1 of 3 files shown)


integration_test/upgrade_test.go
```diff
@@ -49,6 +50,31 @@ func TestKrewUpgrade(t *testing.T) {
...
+	defer cleanup()
+
+	test.WithIndex().
+		Krew("install", validPlugin).
+		RunOrFail()
+
+	test.WithEnv("KREW_OS", "somethingelse")
+
+	// if upgrading 'all' plugins, must succeed
+	out := string(test.Krew("upgrade", "--no-update-index").RunOrFailOutput())
```
corneliusweig: :+1: for using the `--no-update-index` option. Should we add this to other integration tests to speed them up?

ahmetb: Yeah not a bad idea, though worth seeing how much it speeds up. 
My motive here was to not to cause an index update b/c I was gonna modify the manifest file, but then i realized I can just override env.

PR #429, opened 2019-12-09
Support reading manifest url from files

ahmetb:
Addressing some technical debt:

- Introduce indexscanner.ReadPlugin(io.ReadCloser)
- Make ReadPluginFromFile use that to simplify
- Move validation.ValidatePlugin from several places to ReadPlugin() method
- Add missing test for "preserves ENOENT" behavior for ReadPluginFromFile
- Fix verify-gofmt.sh reference (now gone) in run-tests.sh

Adding new stuff:

- introduce --manifest-url that is mut.ex. with --manifest and positional args
- add unit tests for url fetching
- add integration tests for --manifest-url argument behavior

Addresses #193, potentially supersedes #350.

### DIFF
--- a/integration_test/install_test.go
+++ b/integration_test/install_test.go
@@ -156,7 +156,7 @@ func TestKrewInstall_ManifestArgsAreMutuallyExclusive(t *testing.T) {
 	srv, close := localTestServer()
 	defer close()
 
-	if err := test.Krew("install", validPlugin,
+	if err := test.Krew("install",
 		"--manifest", filepath.Join("testdata", fooPlugin+constants.ManifestExtension),
 		"--manifest-url", srv+"/"+validPlugin+constants.ManifestExtension).
 		Run(); err == nil {
--- a/integration_test/install_test.go
+++ b/integration_test/install_test.go
@@ -112,8 +112,8 @@ func TestKrewInstall_ManifestURL(t *testing.T) {
 
 	test, cleanup := NewTest(t)
 	defer cleanup()
-	srv, close := localTestServer()
-	defer close()
+	srv, shutdown := localTestServer()
+	defer shutdown()
 
 	test.Krew("install",
 		"--manifest-url", srv+"/"+validPlugin+constants.ManifestExtension).
@@ -153,8 +153,8 @@ func TestKrewInstall_ManifestArgsAreMutuallyExclusive(t *testing.T) {
 
 	test, cleanup := NewTest(t)
 	defer cleanup()
-	srv, close := localTestServer()
-	defer close()
+	srv, shutdown := localTestServer()
+	defer shutdown()
 
 	if err := test.Krew("install",
 		"--manifest", filepath.Join("testdata", fooPlugin+constants.ManifestExtension),
--- a/integration_test/install_test.go
+++ b/integration_test/install_test.go
@@ -15,6 +15,8 @@
 package integrationtest
 
 import (
+	"net/http"
+	"net/http/httptest"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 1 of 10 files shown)


integration_test/install_test.go
```diff
@@ -132,7 +148,23 @@ func TestKrewInstall_OnlyArchive(t *testing.T) {
...
-func TestKrewInstall_PositionalArgumentsAndManifest(t *testing.T) {
+func TestKrewInstall_ManifestArgsAreMutuallyExclusive(t *testing.T) {
+	skipShort(t)
+
+	test, cleanup := NewTest(t)
+	defer cleanup()
+	srv, close := localTestServer()
+	defer close()
+
+	if err := test.Krew("install", validPlugin,
```
corneliusweig: This command does not fail due to incompatible `--manifest` and `--manifest-url`, but because it mixes positional args and `--manifest*`. This is also covered by the next test-case, and I think it will not pass, if `validPlugin` is removed.

internal/index/indexscanner/scanner.go
```diff
@@ -72,38 +72,37 @@ func LoadPluginListFromFS(indexDir string) ([]index.Plugin, error) {
...
+// LoadPluginByName loads a plugins index file by its name. When plugin
 // file not found, it returns an error that can be checked with os.IsNotExist.
-func LoadPluginFileFromFS(pluginsDir, pluginName string) (index.Plugin, error) {
+func LoadPluginByName(pluginsDir, pluginName string) (index.Plugin, error) {
 	if !validation.IsSafePluginName(pluginName) {
 		return index.Plugin{}, errors.Errorf("plugin name %q not allowed", pluginName)
 	}
 
 	klog.V(4).Infof("Reading plugin %q", pluginName)
-	pluginsDir, err := filepath.EvalSymlinks(pluginsDir)
```
corneliusweig: This `EvalSymlinks` is just unnecessary, right? In the git history, I couldn't find any clue why this should be there, so it should be fine to remove.

internal/index/indexscanner/scanner.go
```diff
@@ -72,38 +72,37 @@ func LoadPluginListFromFS(indexDir string) ([]index.Plugin, error) {
...
 	if os.IsNotExist(err) {
+		// TODO(ahmetb): we should use go1.13+ errors.Is construct at call sites to evaluate if an error is os.IsNotExist
 		return index.Plugin{}, err
 	} else if err != nil {
 		return index.Plugin{}, errors.Wrap(err, "failed to open index file")
 	}
+	return ReadPlugin(f)
+}
+
+func ReadPlugin(f io.ReadCloser) (index.Plugin, error) {
```
corneliusweig: A godoc comment would be nice here. Maybe also mention that the input file will be closed -- or is that the idiomatic usage of `ReadCloser` anyway?

ahmetb: It is :) otherwise this would take io.Reader

integration_test/install_test.go
corneliusweig: Ping @ahmetb. If I change the test to this:
```go
if err := test.Krew("install",
		"--manifest", filepath.Join("testdata", fooPlugin+constants.ManifestExtension),
		"--manifest-url", srv+"/"+validPlugin+constants.ManifestExtension,
		"--archive", filepath.Join("testdata", fooPlugin+".tar.gz")).
		Run(); err == nil {
		t.Fatal("expected mutually exclusive arguments to cause failure")
}
```
It does not fail, but it should.

Iow, the `--manifest` and `--manifest-url` options are not mutually exclusive, in contrast to what the test case claims.

internal/index/indexscanner/scanner.go
ahmetb: Yeah it was added as "defensive programming". There was some debate in our internal Gerrit back in the day. Not necessary.

integration_test/install_test.go
ahmetb: thanks, should be fixed now.

integration_test/install_test.go
```diff
@@ -105,6 +107,20 @@ func TestKrewInstall_Manifest(t *testing.T) {
 	test.AssertExecutableInPATH("kubectl-" + validPlugin)
 }
 
+func TestKrewInstall_ManifestURL(t *testing.T) {
+	skipShort(t)
+
+	test, cleanup := NewTest(t)
+	defer cleanup()
+	srv, close := localTestServer()
+	defer close()
```
corneliusweig: I'm not exactly sure what it is shadowing, but gocritic is unhappy about calling it `close`.

ahmetb: shadows close builtin method on go channels :)

Lets use cleanup.

corneliusweig: :bulb: `cleanup` sounds good.

ahmetb: fixed.

ahmetb:
Should be good to go.

corneliusweig:
/lgtm
/approve
Sorry for the delay 🙏

PR #430, opened 2019-12-09
Update the installation instruction to also run krew update.

### LINKED ISSUE #419: Fresh install of krew complains about local plugin index
I followed the installation guide and installed v0.3.3 on a new Mac. After adding krew to path, the first use complains about krew local plugin index not initialized.

```
F1203 10:07:07.784437   15575 root.go:58] krew local plugin index is not initialized (run "kubectl krew update")
```

Had to run `kubectl krew update` to initialize the local plugin index. Expected this to be done as part of the installation process so a new user can start using the plugin immediately.

tony-yang:
For a fresh install, run krew update to initialize the local plugin index as part of the installation process to avoid the `plugin index is not initialized` error for first time user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,8 +53,9 @@ Check out the list of [kubectl plugins available on krew][list] or just run
       set -x; cd "$(mktemp -d)" &&
       curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/download/v0.3.3/krew.{tar.gz,yaml}" &&
       tar zxvf krew.tar.gz &&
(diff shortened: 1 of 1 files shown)


tony-yang:
/remove-do-not-merge/invalid-commit-message

corneliusweig:
@tony-yang Can you remove the `Fixes #419` from your commit message and force-push? Otherwise prow will not let us merge this :)

tony-yang:
Thanks @corneliusweig, originally updated the text in my first comment to remove the FIX # label but didn't realize I need to also remove them from my commit. Force pushed.

Issue #414, opened 2019-12-01
Consider go1.13 errors package


ahmetb:
Go has introduced new error wrapping primitives. Especially the errors.Is check can be useful to krew in certain few locations that we rely on error type (ENOENT etc). 

https://golang.org/doc/go1.13#error_wrapping

corneliusweig:
Recommended read before doing this: https://blog.golang.org/go1.13-errors

Also, https://github.com/fatih/errwrap might be helpful.

erain:
For reference, the blog post @corneliusweig mentioned is https://blog.golang.org/go1.13-errors

ahmetb:
Thanks. It would be good to see if we get Stack Traces like the current err pkg when run with -v 5 etc

corneliusweig:
Thanks @erain. I got confused and thought that the link from @.ahmetb was the same :man_facepalming:

ferhatelmas:
> Also, fatih/errwrap might be helpful.

I [offered help](https://github.com/fatih/errwrap/issues/2#issuecomment-561347099)  to integrate errwrap into golangci-lint. It have it locally, will push upstream accordingly.

corneliusweig:
@ferhatelmas would that make sense? My understanding is that it should be a deliberate choice to wrap or not to wrap (see blog.golang.org/go1.13-errors). A linter would simply report any such occurrence.

ferhatelmas:
I think yes because even if it's decided to wrap, people are error-prone and might forget putting `%w` and might use `%s` or `%q`.

Also, it helps in transition with `--fix` flag of golangci-lint to automatically refactor codebase.

ferhatelmas:
Yes. Here is a reply to it from the famous [blog post](https://blog.golang.org/go1.13-errors#TOC_3.4.).

> When adding additional context to an error, either with fmt.Errorf or by implementing a custom type, you need to decide whether the new error should wrap the original. There is no single answer to this question; it depends on the context in which the new error is created.

That's why it should be opt-in linter, false by default.

ahmetb:
Here's the criteria I expect from this new errors mechanism:

- ability to wrap errors (to add more context/reason) without having to implement a custom type
- preserves syntax while printing the wrapped error messages (`reason: reason: root cause`)
- preserves ability to print errors normally (as above) or with stack trace when `-v`≥1 specified (currently achieved wit %+v)
[...]

Issue #421, opened 2019-12-03
Check for failures when plugins are removed


corneliusweig:
Currently, it is not entirely clear, how `krew` will behave in the following circumstances:

* A user installed a plugin, and its manifest gets removed from `krew-index`
* A user installed a plugin, and its platform is removed from the corresponding manifest.

This issue is to explore if `krew` runs into any errors with one of its sub-commands in one of the above situations.

---

These scenarios can be tested by manipulating the `index` folder in `$KREW_ROOT` and setting the `--no-update-index` CLI option.

/good-first-issue

ahmetb:
We currently print a really weird "Upgraded" message in case of 

> * A user installed a plugin, and its platform is removed from the corresponding manifest.

For example, I removed `darwin` platform of `ctx` plugin and ran:

````
$ kubectl krew upgrade ctx --no-update-index  -v=10
I1203 15:16:51.666127   66174 upgrade.go:81] --no-update-index specified, skipping updating local copy of plugin index
I1203 15:16:51.666209   66174 scanner.go:82] Reading plugin "ctx"
I1203 15:16:51.667079   66174 upgrade.go:65] Upgrading plugin: ctx
I1203 15:16:51.667286   66174 platform.go:43] Matching platform for labels(arch=amd64,os=darwin)
Upgraded plugin: ctx
```

It definitely didn't do an upgrade, but prints a message saying so.
/kind bug

> * A user installed a plugin, and its manifest gets removed from `krew-index`

I think this scenario is pretty straightforward. We treat it like the plugin was installed manually (`--manifest=.., --archive=..`) and wasn't part of the index to begin with.

When you do an `upgrade` on such a plugin (that is removed from index) today, you'll see:

```
F1203 15:19:40.001383   66283 root.go:58] failed to load the plugin manifest for plugin ns: open /Users/ahmetb/.krew/index/plugins/ns.yaml: no such file or directory
```

...which isn't nice, but can we do any better than this? I think this is OK. (Maybe we can treat errors.Is(os.IsNotExist) and say "plugin no longer exists in the index repository"?)

chriskim06:
I had a few questions about this issue:

When I remove linux from the grep plugin I get this:
```
$ KREW_ROOT="$PWD/playground" out/bin/krew-linux_amd64 upgrade grep --no-update-index
[...]

PR #432, opened 2019-12-10
cmd: Friendly error messages when plugin not found

### LINKED ISSUE #431: Improve some of the error messaging when a plugin manifest does not exist or does not support a specific platform
Some of the messaging around "not found" plugins can be confusing. Here are a few examples:

installing a non-existent plugin
```
$ KREW_ROOT="$PWD/playground" out/bin/krew-linux_amd64 install asdf --no-update-index
F1209 09:09:12.024536    3214 root.go:58] failed to load plugin "asdf" from the index: open /home/kimc/go/src/sigs.k8s.io/krew/playground/index/plugins/asdf.yaml: no such file or directory
```

upgrading plugin when manifest is deleted
```
$ KREW_ROOT="$PWD/playground" out/bin/krew-linux_amd64 upgrade grep --no-update-index
F1208 22:47:56.737799    8298 root.go:58] failed to load the plugin manifest for plugin grep: open /home/kimc/go/src/sigs.k8s.io/krew/playground/index/plugins/grep.yaml: no such file or directory
```

installing already installed plugin but plugin manifest removed from index
```
$ KREW_ROOT="$PWD/playground" out/bin/krew-linux_amd64 install grep --no-update-index
F1209 09:11:03.179194    3735 root.go:58] failed to load plugin "grep" from the index: open /home/kimc/go/src/sigs.k8s.io/krew/playground/index/plugins/grep.yaml: no such file or directory
```

The error that gets printed could be updated based on a check like `errors.Is(err, os.ErrNotExist)` for the first two. The last one should probably see that the plugin is already installed and mention something about that rather than trying to load the manifest again.

This issue was created after some discussion in https://github.com/kubernetes-sigs/krew/issues/421

chriskim06:
Fixes #431 

I decided to just use `os.IsNotExist` on the error rather than `errors.Is` since there's a package name conflict with the existing `github.com/pkg/errors` package. I was thinking about using the new %w verb in `fmt.Errorf` but noticed that you lose stack trace info when setting the log level. There's an existing issue, https://github.com/kubernetes-sigs/krew/issues/414, that is open to consider using 1.13 logging so I figured I'd use this for now and we could change it when the other issue gets figured out.

### DIFF
--- a/cmd/krew/cmd/install.go
+++ b/cmd/krew/cmd/install.go
@@ -16,7 +16,6 @@ package cmd
 
 import (
 	"bufio"
-	errs "errors"
 	"fmt"
 	"os"
 
@@ -100,7 +99,7 @@ Remarks:
 				}
 				plugin, err := indexscanner.LoadPluginFileFromFS(paths.IndexPluginsPath(), name)
 				if err != nil {
-					if errs.Is(err, os.ErrNotExist) {
+					if os.IsNotExist(err) {
 						return fmt.Errorf("plugin \"%s\" does not exist in the plugin index", name)
 					}
 					return errors.Wrapf(err, "failed to load plugin %q from the index", name)
--- a/cmd/krew/cmd/install.go
+++ b/cmd/krew/cmd/install.go
@@ -16,6 +16,7 @@ package cmd
 
 import (
 	"bufio"
+	errs "errors"
 	"fmt"
 	"os"
 
@@ -89,6 +90,9 @@ Remarks:
 			for _, name := range pluginNames {
 				plugin, err := indexscanner.LoadPluginFileFromFS(paths.IndexPluginsPath(), name)
 				if err != nil {
+					if errs.Is(err, os.ErrNotExist) {
+						return errors.New(fmt.Sprintf("plugin \"%s\" does not exist in the plugin index", name))
+					}
 					return errors.Wrapf(err, "failed to load plugin %q from the index", name)
 				}
 				install = append(install, plugin)
--- a/cmd/krew/cmd/install.go
+++ b/cmd/krew/cmd/install.go
@@ -90,7 +90,7 @@ Remarks:
 				plugin, err := indexscanner.LoadPluginFileFromFS(paths.IndexPluginsPath(), name)
 				if err != nil {
 					if os.IsNotExist(err) {
-						return errors.New(fmt.Sprintf("plugin %q does not exist in the plugin index", name))
+						return errors.Errorf("plugin %q does not exist in the plugin index", name)
 					}
 					return errors.Wrapf(err, "failed to load plugin %q from the index", name)
 				}
--- a/cmd/krew/cmd/install.go
+++ b/cmd/krew/cmd/install.go
@@ -86,12 +86,22 @@ Remarks:
 				return errors.New("--archive can be specified only with --manifest")
 			}
 
+			installed, err := installation.ListInstalledPlugins(paths.InstallReceiptsPath())
+			if err != nil {
+				return errors.Wrap(err, "failed to find all installed versions")
+			}
+			skipped := false
 			var install []index.Plugin
 			for _, name := range pluginNames {
+				if _, ok := installed[name]; ok {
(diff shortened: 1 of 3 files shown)


ahmetb:
/hold
let's hold off on this until #429 is merged, there's a fair amount of merge conflicts involved.

ahmetb:
> I decided to just use `os.IsNotExist` on the error rather than `errors.Is` since

I think this wouldn't work currently, because (1) github/pkg/errors.Wrap doesn't wrap the errors in a way the new go1.13 errors.Is() can use (2) the rest of the codebase must migrate to the errors package properly before we start incorporating parts of it.

cmd/krew/cmd/install.go
```diff
@@ -85,10 +85,23 @@ Remarks:
...
+			installed, err := installation.ListInstalledPlugins(paths.InstallReceiptsPath())
+			if err != nil {
+				return errors.Wrap(err, "failed to find all installed versions")
+			}
+			skipped := false
 			var install []index.Plugin
 			for _, name := range pluginNames {
+				if _, ok := installed[name]; ok {
+					fmt.Fprintf(os.Stderr, "plugin \"%s\" is already installed\n", name)
+					skipped = true
```
ahmetb: I'm not sure what this part does. We already have the skipping behavior implemented (see below), we can't implement it again like this. I don't see how this changes anything?

```
$ kubectl krew install krew
Installing plugin: krew
W1209 19:07:29.190869   69309 install.go:124] Skipping plugin "krew", it is already installed
```

ahmetb: I guess it helps us not look up the deleted manifest of a plugin that's already installed. I don't think we need this, the error you added below does a proper job.

cmd/krew/cmd/install.go
```diff
@@ -85,10 +85,23 @@ Remarks:
...
+					continue
+				}
 				plugin, err := indexscanner.LoadPluginFileFromFS(paths.IndexPluginsPath(), name)
 				if err != nil {
+					if os.IsNotExist(err) {
+						return fmt.Errorf("plugin \"%s\" does not exist in the plugin index", name)
```
ahmetb: you can use %q.
and also please use `errors.New`, as it creates a stacktrace.

cmd/krew/cmd/install.go
chriskim06: I thought that skipping the manifest lookup for an already installed plugin and displaying a different message was what we wanted for this based on your comment [here](https://github.com/kubernetes-sigs/krew/issues/421#issuecomment-563102188):

> ...  we try to load manifest from fs again, I wonder why. we should be treating it like ErrAlreadyInstalled and say it's already installed

If its not necessary I can totally remove it, but I thought it seemed a little nicer.

Another reason I added it was for the case when the plugin(s) that are being installed don't exist so that the help message doesn't get printed. It felt out of place to show that as well.

ahmetb: Yeah, I think it's making the code unnecessarily more complex and it solves a small problem that rarely occurs if ever (installing an installed plugin directly by name, after its manifest is which occurred after installing it for the first time). Let's abandon this.

ahmetb:
/retitle cmd: Friendly error messages when plugin not found

cmd/krew/cmd/install.go
chriskim06: Sounds good, I'll remove it

cmd/krew/cmd/install.go
```diff
@@ -89,6 +89,9 @@ Remarks:
 			for _, name := range pluginNames {
 				plugin, err := indexscanner.LoadPluginFileFromFS(paths.IndexPluginsPath(), name)
 				if err != nil {
+					if os.IsNotExist(err) {
+						return errors.New(fmt.Sprintf("plugin %q does not exist in the plugin index", name))
```
ahmetb: sorry for back and forth but you can use errors.Errorf for this :)

chriskim06: No worries, I'll make the change tomorrow

ahmetb:
/lgtm
/approve
/hold cancel
I'm cancelling hold b/c now this is simple enough for me to rebase the other pr.

PR #433, opened 2019-12-10
Disallow fmt.Errorf usage through code validation

ahmetb:
Came across during a code review, new contributors can end up using fmt.Errorf
which does not create stack trace at the error site --and also doesn't preserve
stack trace if used to wrap an error.


ferhatelmas:
Related to #414, when it's done, this should be removed too.

ahmetb:
Yeah. We have to wait until Cornelius to come back from vacation to merge things. :)

corneliusweig:
Huh, holidays? Nah, I was just being slow.. :christmas_tree: :santa: time..
/lgtm
/approve

PR #435, opened 2019-12-13
Improve linting and fix found issues

ferhatelmas:
Enable all rules from gocritic except some troublesome rules (see `.golangci.yml` for details).

Here is a part of what new rules bring:
* don't shadow imports or builtins such as
`os` and `close` respectively.
* use wrapper ReplaceAll
* combine types in func calls if same type
* find duplicate imports
* find suspicious regex
* ensure proper comment format

To fix some of the issues, I created a type for os and arch, then exported it.

### DIFF
--- a/cmd/validate-krew-manifest/main.go
+++ b/cmd/validate-krew-manifest/main.go
@@ -163,7 +163,7 @@ func installPlatformSpec(manifestFile string, p index.Platform) error {
 }
 
 // findAnyMatchingPlatform finds an <os,arch> pair matches to given selector
-func findAnyMatchingPlatform(selector *metav1.LabelSelector) installation.GoOSArch {
+func findAnyMatchingPlatform(selector *metav1.LabelSelector) installation.OSArchPair {
 	for _, p := range allPlatforms() {
 		if selectorMatchesOSArch(selector, p) {
 			klog.V(4).Infof("%s MATCHED <%s>", selector, p)
@@ -171,10 +171,10 @@ func findAnyMatchingPlatform(selector *metav1.LabelSelector) installation.GoOSAr
 		}
 		klog.V(4).Infof("%s didn't match <%s>", selector, p)
 	}
-	return installation.GoOSArch{}
+	return installation.OSArchPair{}
 }
 
-func selectorMatchesOSArch(selector *metav1.LabelSelector, env installation.GoOSArch) bool {
+func selectorMatchesOSArch(selector *metav1.LabelSelector, env installation.OSArchPair) bool {
 	sel, err := metav1.LabelSelectorAsSelector(selector)
 	if err != nil {
 		// this should've been caught by validation.ValidatePlatform() earlier
@@ -188,9 +188,9 @@ func selectorMatchesOSArch(selector *metav1.LabelSelector, env installation.GoOS
 }
 
 // allPlatforms returns all <os,arch> pairs krew is supported on.
-func allPlatforms() []installation.GoOSArch {
+func allPlatforms() []installation.OSArchPair {
 	// TODO(ahmetb) find a more authoritative source for this list
-	return []installation.GoOSArch{
+	return []installation.OSArchPair{
 		{OS: "windows", Arch: "386"},
 		{OS: "windows", Arch: "amd64"},
 		{OS: "linux", Arch: "386"},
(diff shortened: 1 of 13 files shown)


ferhatelmas:
@ahmetb It was mostly from my reading but checked linters and added more rules 
 and it found more issues so I added more refactoring.

cmd/validate-krew-manifest/main.go
```diff
@@ -149,55 +148,55 @@ func installPlatformSpec(manifestFile string, p index.Platform) error {
...
-		"arch": arch,
+		"os":   env.OS,
+		"arch": env.Arch,
 	})
 }
 
 // allPlatforms returns all <os,arch> pairs krew is supported on.
-func allPlatforms() [][2]string {
+func allPlatforms() []installation.GoOSArch {
 	// TODO(ahmetb) find a more authoritative source for this list
```
ferhatelmas: `go tool dist list`

corneliusweig: well, now we now how to get a complete list :)

So what shall we do about it?

ahmetb: I am aware of that. But that would require us running a cmd and keeping code up to date periodically. This is ok for now.

internal/installation/platform_test.go
```diff
@@ -27,12 +27,12 @@ import (
 
 func Test_osArch(t *testing.T) {
 	inOS, inArch := runtime.GOOS, runtime.GOARCH
-	outOS, outArch := osArch()
-	if inOS != outOS {
-		t.Errorf("returned OS=%q; expected=%q", outOS, inOS)
+	out := OSArch()
+	if inOS != out.OS {
+		t.Errorf("returned OS=%q; expected=%q", out.OS, inOS)
```
ahmetb: Use cmp.diff here
What if we add another field :)

internal/installation/platform.go
```diff
@@ -55,16 +55,29 @@ func matchPlatform(platforms []index.Platform, os, arch string) (index.Platform,
 	return index.Platform{}, false, nil
 }
 
-// osArch returns the OS/arch combination to be used on the current system. It
+// GoOSArch is wrapper around operating system and architecture
+type GoOSArch struct {
```
ahmetb: Can we call this Platform or is that too generic?
We don’t care about “Go” part that much.

internal/installation/platform_test.go
ferhatelmas: Updated.

internal/installation/platform.go
ferhatelmas: It can be used in index too (since exported I didn't want to change in a lint PR).
They are also go specific so for now, it might leave as it is.

corneliusweig: :thinking: Wasn't `GoOSArch` that just added in this PR? I think we can choose whatever name suits best. Besides, everything is internal now, so we don't have to bother.

ferhatelmas: Yes, I rename to whatever you want. 

What I mean is that with a better name it could be used in non-internal places too.

corneliusweig: Ok :)

So what about `Platform` as proposed by Ahmet?

ferhatelmas: Sounds good to me since it's in the platform package. Renaming unless there is a better suggestion.

ahmetb: ClientPlatform or OSArchPair isn't that bad either.

ahmetb:
> checked linters and added more rules

Is there a way we can add these linters preventing shadowing builtins to the CI validation, so that we don't have to go through this again?

ferhatelmas:
> Is there a way we can add these linters preventing shadowing builtins to the CI validation

I already added, see the diff in [`.golangci.yml`](https://github.com/kubernetes-sigs/krew/pull/435/files#diff-e23b7926f022c85af1665b36d256231bR3).

internal/installation/platform.go
ferhatelmas: I think renaming it to `Platform` could be misleading with `index.Platform`.

It's just OS and Arch so a simple naming might be better. `OSArchPair` sounds somehow better to me.

ferhatelmas:
Anything else is needed?

ahmetb:
Thanks. I don't fully understand the list of linter check in .golangci.yml, but it's ok for now.
/lgtm
/approve

PR #438, opened 2019-12-17
Show "Upgrading plugin %s" message

ahmetb:
Promoting this from log to user-visible message, because some plugins actually
are ~20-30 MiB and takes a significant time to download. During this time
we show no indication of progress while the download is happening.

/assign @corneliusweig


corneliusweig:
Love the change. There seems to be a problem with `gofmt` though.

ahmetb:
friggin goland.

corneliusweig:
/lgtm
/approve

Sorry for the delay.

PR #439, opened 2019-12-18
Disable all first then enable needed linters

ferhatelmas:
First disable all linters then enable what we need. Otherwise, all linters are running then results are filtered. See with `-v` flag.

Also, sort enabled linters in config.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -36,21 +36,22 @@ run:
   tests: true
 
 linters:
+  disable-all: true
   enable:
-    - gofmt
+    - deadcode
+    - errcheck
     - gocritic
+    - gofmt
     - goimports
     - golint
     - gosimple
     - interfacer
     - maligned
     - misspell
-    - unconvert
-    - unparam
-    - stylecheck
+    - prealloc
     - staticcheck
     - structcheck
-    - prealloc
-    - deadcode
+    - stylecheck
+    - unconvert
+    - unparam
     - varcheck
-    - errcheck


corneliusweig:
I'm not sure I understand what is the problem here. I cannot find any filtering of linter results going on when running with `-v`.

Besides, I think the enabled linters are correct and some are active by default whereas others are enabled explicitly. Iow, the list of enabled linters in our `.golangci.yaml` is not exhaustive and does not list what is enabled by default anyways. Does that make sense?

ferhatelmas:
> some are active by default

This is the key part. If we don't disable all first, we might be adding more into already enabled ones or our config is noop. However, if we disable first, then the enable list explicitly specifies what is active.

It's somehow related to silence some linters which are useful in the context of high performant server application, but not in a cli such as predeclared.

corneliusweig:
Alright. Being explicit about the active linters would be nice.

There is some difference between the 17 active linters in this PR (deadcode errcheck gocritic gofmt goimports golint gosimple interfacer maligned misspell prealloc staticcheck structcheck stylecheck unconvert unparam varcheck) vs. the 21 currently active on master (deadcode errcheck gocritic gofmt goimports golint gosimple govet ineffassign interfacer maligned misspell prealloc staticcheck structcheck stylecheck typecheck unconvert unparam unused varcheck). Would you care to clarify why you want to disable those four linters?

ferhatelmas:
> It's somehow related to silence some linters which are useful in the context of high performant server application, but not in a cli such as predeclared.

Sorry for the typo, I had meant prealloc, instead of predeclared.

ferhatelmas:
> Would you care to clarify why you want to disable those four linters?

Sure, actually my intention was to be explicit and then disable one by one with a discussion.

For this goal to happen, I need to ensure that active linter list isn't modified in this PR. Let me update. However, I won't enable typecheck because it's already covered by tests.

ahmetb:
/lgtm
/approve
Let's give it a try.

PR #449, opened 2019-12-30
Bump golangci-lint to v1.22.2

ferhatelmas:
Regular bump of lint tool to the latest.


ahmetb:
We don’t need to bump our CI tool this frequently. :)
Let’s only pick up significant changes from now on. 
/lgtm
/approve

PR #450, opened 2020-01-02
allow linux/arm64 plugins

ahmetb:
This came up in https://github.com/kubernetes-sigs/krew-index/pull/407.
We don't currently release krew for linux/arm64, but apparently linux/arm
distribution we have works there.

This would allow plugins to be distributed for linux/arm64, in addition to
existing linux/arm.

/assign @corneliusweig


corneliusweig:
/lgtm
/approve
Nice :)

PR #452, opened 2020-01-02
Bump shfmt to v3

ferhatelmas:
Stable module supported pure go release.

### DIFF
--- a/hack/run-lint.sh
+++ b/hack/run-lint.sh
@@ -53,7 +53,7 @@ install_shfmt() {
 
   cd "${shfmt_dir}"
   go mod init foo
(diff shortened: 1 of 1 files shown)


ahmetb:
/lgtm
/approve
Do we need to stay up to date this much? I've also commented the same thing on the last golang lint PR. :)

ferhatelmas:
> Do we need to stay up to date this much? I've also commented the same thing on the last golang lint PR. :)

Probably not but it's also a good habit. 

I also saw your previous comment but did this too since it's major bump. I expected we would catch some new issues but the code is clean, nice job :smile:

PR #453, opened 2020-01-02
Ensure go mod is tidied

ferhatelmas:
Sometimes when a dependency is added, it could be forgotten to be tidied.
This ensures it's always tidied.


corneliusweig:
/lgtm
/approve
Thanks for helping us to keep everything clean :D

Issue #405, opened 2019-11-26
Proposal: Simplify installation of krew as it makes adoption harder


laverya:
Today, to install a plugin on a new machine, the user has to install krew and then the plugin. The recommended process involves copying and running a script that downloads a tarball with curl, decompressing it, and then executing a binary from within that tarball. After this completes, the user is then to add krew to their PATH manually, and finally to install the desired plugin via krew.

I’d like to propose a solution where we integrate the krew and plugin install process into a single script that automatically detects the shell and injects the PATH addition into the relevant bashrc/zshrc/etc file.

This isn’t inherently more insecure than the current solution because both fundamentally involve executing untrusted code directly from the internet.

The advantages of this are having an easy, memorable install process (`curl https://krew.sh | bash`) and that’s faster and more extensible than the current setup.

I’ve experimented with a prototype of this idea that’s currently available at [krew.sh](https://krew.sh). There are a number of things that could be added to this - allowing the user to specify a prior krew version, for instance - but it functions at a basic level today.

I don’t think this should be the recommended, default, or only approach to installing krew. I’m proposing that this is created and hosted as a supported alternative to the current methods.

I think this will increase adoption of krew because developers will be able to give their customers a very easy, one line install as opposed to a multi-step process that involves copying scripts into the shell.

corneliusweig:
Hey @laverya,
thanks for bringing the discussion onto this public platform. I have a few concerns about this:

* `curl http://something | sh` is an anti-pattern. We don't want to endorse it.
* The krew maintainers have no control over `krew.sh`. We have no idea what scripts are served there.
* The installation instructions on krew's homepage are already pretty straightforward. Everyone sees what's being executed and there is no execution of an unknown script.
* Krew also has installation instructions for `fish` and Windows. AFAICT, krew.sh only supports `bash` and `zsh`.

I will admit, that a one-line installation is nice. But it's not safe.

laverya:
Addressing each in turn:

- I'd argue that running untrusted, unknown code is the anti pattern. We often see unsafe `curl | sudo bash` scripts as an implementation, but is downloading an untrusted .tar.gz, extracting it, and executing it inherently safer? A bash script at least has the potential to offer more transparency in that regard. I understand that this is probably more of a perception problem because I don't think it fundamentally is any safer or less safe than the current pattern used.

[...]

PR #456, opened 2020-01-03
Make installing krew even simpler

unixorn:
* Eliminate potential copy-pasta errors by adding an install script.
* Make `bashrc`/`zshrc` snippet only add `krew`'s `bin` directory if it exists.
* Update README.md to include new install option.


unixorn:
Signed the CLA

unixorn:
Trying to retrigger the CLA check

ahmetb:
Nice. While it's tempting, it's not a whole lot different than what we already have :| This script can be `curl | sh`'ed (which is being discussed at #405).

On the other hand, we now have to update not only the readme, but another file while releasing a new version.

unixorn:
Maybe double tag releases as x.y.z and also stable? then the script could just refer to stable, and you wouldn't have to update the readme every release either.

ahmetb:
Instead of re-tagging using the same tag name, we might make use of some redirection through a domain name we own, like krew.dev and create a curl|sh endpoint, but for now it's not a desired installation method and usually frowned upon in Kubernetes community. But more and more, it feels unavoidable.

unixorn:
Or have a STABLE file in the repo with a version number. Then install can be `git clone git@github.com:kubernetes-sigs/krew.git && krew/install-krew`. It's tiny bit more complex than curl|sh, but at least then `git` takes care of verifying the source.

And agreed, curl|sh is ugly, especially given the potential shenanigans shown in https://www.idontplaydarts.com/2016/04/detecting-curl-pipe-bash-server-side/

ahmetb:
Specifically why not “double tag”:

* it’s not a good practice to have floating tags, as they are meant to be snapshots
* we use file download counts on each tag to track krew usage. deleting a tag and re tagging would lose that.

unixorn:
Yeah, double tagging feels kludgier the more we discuss it. A `STABLE_RELEASE` file seems cleaner, but still not perfect.

Better than a raw curl|sh though.

ahmetb:
STABLE_RELEASE would actually work. We’d still need to update windows instructions manually, though.

unixorn:
Ok, I'll update to support that so you can see if it's less ugly

unixorn:
BTW, do you prefer the commits squashed?

ahmetb:
No preference. In fact we don’t even need code (we could have discussed a proposal issue) but it’s ok.

install-krew
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+#
+# Install krew
+#
+# License: Apache 2.0 - see https://github.com/kubernetes-sigs/krew/blob/master/LICENSE
+
+set -o pipefail
+set -o errexit
+
+STABLE_RELEASE="$(cat $(dirname $0)/STABLE_RELEASE)"
```
corneliusweig: You need to fetch the current `STABLE_RELEASE` file from remote. Users should not have to clone krew or download the source tarball for install.

corneliusweig:
I'm not sure I follow correctly. AFAICT, you propose that the installation will still go via an installer script. Unless the goal is to enable the `curl|sh` pattern, I don't see any advantage over the current shell install snippet. I'd even say the UX is worse, because a new user needs to download the script, make it executable and so on.

If the `curl|sh` thing is the goal, then let's continue the discussion on #405.

unixorn:
Downloading is `git clone`, and the script is marked executable in the repository. It's not meant to be a `curl|sh` enabler, I added it to the repo so `git` will automatically handle confirming that it's the correct commit without the end user having to do any SHA checks by hand.

corneliusweig:
I don't think this will improve the UX for new users. New users should be onboarded by downloading the krew binary and setting up the env variables to use krew. If the installation now goes via cloning the repository this makes things more complicated, don't you think?

ahmetb:
+1 this is sadly not improving the experience, it's the same thing in an alternative form, plus there are no _copy-pasta issues_ reported as far as I know.

Let's continue brainstorming at #405. It seems like anything short of curl|sh is not going to make a substantial change.
/close

PR #457, opened 2020-01-04
Show updated plugin manifests and available upgrades

### LINKED ISSUE #443: Show what's updated in "update" command (or its invocations indirectly)
This is something Homebrew has, and I've been looking to have it in Krew, too.

When you run `krew update` (or indirectly, through `krew install` or `krew upgrade`), you'll see a message: `Updated the local copy of plugin index.` This message is nice, but it doesn't provide the big picture.

I think for `update` command specifically, it would be good if we can show something like:

```
$ kubectl krew install foo
\
 │ Updating the local copy of plugin index.
 │   New plugins available:
 │     * multiwatch v0.1.0
 │     * view-secret v0.1.9
 │   The following plugins have new version:
 │     * whoami v0.3.2 -> v0.4.0
 │     * rbac-lookup v1.0.2 -> v1.0.3
 │     * get-all v0.8.0 -> v0.9.1  🚨 (a sign to indicate it's installed)
/
Installing plugin "foo".
[...]
```

Some things to consider:

* requires significant change to underlying update mechanism, as we need to parse all index, store in-memory, then compare (worth it?)
* the above example could be too much output for the `install` command, maybe we figure out a one-liner version like:

       New plugins available: multiwatch v0.1.0, multiwatch v0.1.0
       Upgrades available: get-all v0.8.0 -> v0.9.1 <--- 🚨only installed ones

* this should probably be tested only as _unit tests_ as it would be unnecessarily complex to make _integration tests_ for the update scenarios

I'm supportive of this change only if it can be done with small effort.

/priority P3
/priority backlog
/good-first-issue

artmello:
Closes #443


corneliusweig:
First of all, thanks for taking a shot at this!
The function `ensureIndexUpdated` is actually called in three places: `install`, `update`, and `upgrade`. I'm not sure if it really makes sense to show this for each of these cases.
Also, I wonder if we can get away without scanning the manifest index each time. We could also use `git diff --name-only` to restrict the number of scanned manifests. WDYT?

artmello:
Hi @corneliusweig, thanks for taking a look on this!
On the original issue, it was mentioned that for other commands (i.e `install`) we could have a short version of the updated plugins output. Maybe we can have either a verbose/quiet option to control when that is displayed or not. WDYT?

I agree that we should avoid scanning index when not necessary. I tried to implement suggested approach using `diff --name-only` let me know if it does look any better.

This is current output:
```
$ cd -
~/.krew/index

$ git reset --hard dc075b6355c02e30aa0a43bebf4587bab2eafb9d                                                                                     HEAD is now at dc075b6 update neat to v1.1.0 (#399)

$ cd -
~/code/krew

$ ./out/bin/krew-darwin_amd64 update
  New plugins available:
    * tree v0.2.0
  The following plugins have new version:
    * pod-dive v0.1.3 -> v0.1.4 (!)
    * fleet v0.1.3 -> v0.1.4
    * open-svc v2.2.0 -> v2.3.0
    * view-allocations v0.6.0 -> v0.6.2
    * view-serviceaccount-kubeconfig v2.0.3 -> v2.1.1

$ ./out/bin/krew-darwin_amd64 update
```

ahmetb:
For now let’s just show the same output for all (install/update/upgrade). I initially proposed we only show a summary (items combined into single line). Maybe it’s not worth it. Let’s get a feel for it.

aimbot31:
Still ongoing ?

internal/gitutil/git.go
```diff
@@ -69,7 +70,27 @@ func EnsureUpdated(uri, destinationPath string) error {
 	return updateAndCleanUntracked(destinationPath)
 }
 
-func exec(pwd string, args ...string) error {
+// ListModifiedFiles will fetch origin and list modified files
+func ListModifiedFiles(uri, destinationPath string) ([]string, error) {
```
corneliusweig: `ListModifiedFiles` is a slight misnomer, because no files are actually modified. What about `ListUpstreamChanges`?

cmd/krew/cmd/update.go
```diff
@@ -41,12 +57,147 @@ Remarks:
...
+	}
+
+	return plugins, nil
+}
+
+func retrievePluginNameVersionMap(names []string) map[string]string {
+	m := make(map[string]string, len(names))
+	for _, n := range names {
+		plugin, err := indexscanner.LoadPluginByName(paths.IndexPluginsPath(), n)
+		if err != nil {
```
corneliusweig: You should only continue for `os.IsNoExist(err)`. All other errors should abort the function.

cmd/krew/cmd/update.go
```diff
@@ -41,12 +57,147 @@ Remarks:
 	RunE: ensureIndexUpdated,
 }
 
+func retrieveUpdatedPluginList() ([]string, error) {
+	modifiedFiles, err := gitutil.ListModifiedFiles(constants.IndexURI, paths.IndexPluginsPath())
+	if err != nil {
+		return []string{}, err
```
corneliusweig: Please wrap the error with some meaningful message `errors.Wrapf(err, ...)`. Same in other places.

cmd/krew/cmd/update.go
```diff
@@ -41,12 +57,147 @@ Remarks:
...
+
+	for name, version := range updated {
+		oldVersion, ok := old[name]
+		if !ok {
+			newPluginList = append(newPluginList, newPlugin{
+				name:    name,
+				version: version,
+			})
+			continue
+		}
```
corneliusweig: This could also be a
```go
newVersionPlugin{
  ...,
  oldVersion: "",
  installed:  false,
}
```
That could simplify things a little. WDYT?

cmd/krew/cmd/update.go
```diff
@@ -41,12 +57,147 @@ Remarks:
...
+	})
+
+	return newPluginList, newVersionList
+}
+
+func showUpdatedPlugins(newPluginList []newPlugin, newVersionList []newVersionPlugin) {
+	if len(newPluginList) > 0 {
+		fmt.Fprintln(os.Stderr, "  New plugins available:")
+		for _, np := range newPluginList {
+			fmt.Fprintf(os.Stderr, "    * %s %s\n", np.name, np.version)
```
corneliusweig: I think a `TabWriter` would produce more legible output here.

But as far as I understand, a one-line report would be preferable anyways.

cmd/krew/cmd/update.go
```diff
@@ -41,12 +57,147 @@ Remarks:
...
+		return errors.Wrap(err, "failed to load the list of updated plugins from the index")
+	}
+
+	var oldMap map[string]string
+	if len(updatedPlugins) > 0 {
+		oldMap = retrievePluginNameVersionMap(updatedPlugins)
+	}
+
 	klog.V(1).Infof("Updating the local copy of plugin index (%s)", paths.IndexPath())
 	if err := gitutil.EnsureUpdated(constants.IndexURI, paths.IndexPath()); err != nil {
```
corneliusweig: `git fetch` will be executed twice. Here and in line 170. As this involves talking to a remote this is quite slow, and I would like to avoid the second one if possible.

cmd/krew/cmd/update.go
```diff
@@ -41,12 +57,147 @@ Remarks:
...
+}
+
 func ensureIndexUpdated(_ *cobra.Command, _ []string) error {
+	updatedPlugins, err := retrieveUpdatedPluginList()
+	if err != nil {
+		return errors.Wrap(err, "failed to load the list of updated plugins from the index")
+	}
+
+	var oldMap map[string]string
+	if len(updatedPlugins) > 0 {
```
corneliusweig: If `len(updatedPlugins) == 0`, you can return early here, because this means the index is already up-to-date.

internal/gitutil/git.go
```diff
@@ -69,7 +70,27 @@ func EnsureUpdated(uri, destinationPath string) error {
...
+	if _, err := exec(destinationPath, "fetch", "-v"); err != nil {
+		return []string{}, errors.Wrapf(err, "fetch index at %q failed", destinationPath)
+	}
+
+	output, err := exec(destinationPath, "diff", "--name-only", "@{upstream}", "--", ".")
```
corneliusweig: Sorry if I come up with those ideas bit by bit. But you can also use `git diff --name-status` which will add a mark each change as modified/added/deleted...

You can use that to further simplify the logic you need to build the report. I suggest that you play around with a toy git repo to see how this works. Try to offload as much as possible to git :)

corneliusweig:
Hi @aimbot31, only @artmello can answer that :).

artmello:
Hey @corneliusweig thanks a lot for reviewing it. I will work on your comments and let you know when it is good for another round.

Hey @aimbot31, thanks for the interest. I plan to work on Cornelius comments and on adding tests (as mentioned this may need some refactor). Not sure how we could split it, but let me know if you have any suggestion :)

cmd/krew/cmd/update.go
```diff
@@ -41,12 +57,147 @@ Remarks:
...
+		m[n] = plugin.Spec.Version
+	}
+
+	return m
+}
+
+func retrieveInstalledPluginMap() (map[string]bool, error) {
+	plugins, err := installation.ListInstalledPlugins(paths.InstallReceiptsPath())
+	if err != nil {
+		return map[string]bool{}, err
```
ahmetb: you can use nil here.
consider wrapping err. (e.g. "failed during listing installed plugins")

cmd/krew/cmd/update.go
```diff
@@ -41,12 +57,147 @@ Remarks:
...
+			continue
+		}
+
+		m[n] = plugin.Spec.Version
+	}
+
+	return m
+}
+
+func retrieveInstalledPluginMap() (map[string]bool, error) {
```
ahmetb: why's the return type a map here?
you only need to just return names, i.e. `[]string`?

cmd/krew/cmd/update.go
```diff
@@ -41,12 +57,147 @@ Remarks:
...
+
+		return newVersionList[i].name < newVersionList[j].name
+	})
+
+	return newPluginList, newVersionList
+}
+
+func showUpdatedPlugins(newPluginList []newPlugin, newVersionList []newVersionPlugin) {
+	if len(newPluginList) > 0 {
+		fmt.Fprintln(os.Stderr, "  New plugins available:")
```
ahmetb: it would be better if you just accept `out io.Writer` here, so that you don't repeat os.Stderr everywhere.

cmd/krew/cmd/update.go
```diff
@@ -41,12 +57,147 @@ Remarks:
...
+		filename := filepath.Base(f)
+		extension := filepath.Ext(filename)
+		name := filename[0 : len(filename)-len(extension)]
+		plugins = append(plugins, name)
+	}
+
+	return plugins, nil
+}
+
+func retrievePluginNameVersionMap(names []string) map[string]string {
```
ahmetb: I'm not convinced this should be map[string]string.
It's not great for extensibility or readability.

Feel free to either,
- just create a `type pluginInfo struct {name, version string }`, or
- just return the `[]*Plugin` objects.

cmd/krew/cmd/update.go
```diff
@@ -41,12 +57,147 @@ Remarks:
...
+		if newVersionList[i].installed && !newVersionList[j].installed {
+			return true
+		}
+
+		if !newVersionList[i].installed && newVersionList[j].installed {
+			return false
+		}
+
+		return newVersionList[i].name < newVersionList[j].name
+	})
```
ahmetb: pro-tip:

if you don't do maps (which I don't see a reason why you'd need maps here), you don't need to worry about sorting at all.

all the methods we have **already** are returning `*index.Plugin` types **sorted by name**. :)

cmd/krew/cmd/update.go
```diff
@@ -41,12 +57,147 @@ Remarks:
...
+
+	return newPluginList, newVersionList
```
ahmetb: shout-out to my comment above, this looks odd.

just create an internal type (struct) to carry these.

cmd/krew/cmd/update.go
```diff
@@ -41,12 +57,147 @@ Remarks:
...
+	for name := range plugins {
+		m[name] = true
+	}
+
+	return m, nil
+}
+
+func filterAndSortUpdatedPlugins(old, updated map[string]string, installed map[string]bool) ([]newPlugin, []newVersionPlugin) {
+	var newPluginList []newPlugin
+	var newVersionList []newVersionPlugin
```
ahmetb: something I noticed throughout your coding style is how you prefix things with their types (e.g. xxxMap xxxList). let's avoid that. :)

you can simply use `newPlugins`, `updatedPlugins` here.

cmd/krew/cmd/update.go
ahmetb: actually it seems like you have a `type newPlugin struct`, so let's follow suit here. either create a new type or find a way of reusing.

cmd/krew/cmd/update.go
```diff
@@ -17,15 +17,31 @@ package cmd
...
+	"sigs.k8s.io/krew/internal/installation"
 	"sigs.k8s.io/krew/pkg/constants"
 )
 
+type newPlugin struct {
+	name    string
+	version string
+}
+
+type newVersionPlugin struct {
```
ahmetb: rename: updatedPlugin

cmd/krew/cmd/update.go
```diff
@@ -17,15 +17,31 @@ package cmd
...
 	"github.com/spf13/cobra"
 	"k8s.io/klog"
 
 	"sigs.k8s.io/krew/internal/gitutil"
+	"sigs.k8s.io/krew/internal/index/indexscanner"
+	"sigs.k8s.io/krew/internal/installation"
 	"sigs.k8s.io/krew/pkg/constants"
 )
 
+type newPlugin struct {
```
ahmetb: I feel like you can simply have 1 type to carry out everything

```
type pluginInfo struct {
    name, version string
}
```

If you simply return `[]pluginInfo` from everywhere, when the time comes to printing, just diff two `[]pluginInfo` slices to find the "new" and "updated" plugins.

Correct me if I'm wrong but this can even help reduce some of the code.

internal/gitutil/git.go
ahmetb: Let's not touch `gitinfo` if we can.

Instead, let's do this:

1. parse index into []pluginInfo
2. update()
3. parse index into []pluginInfo
4. print diff (new and updated plugins), if any

corneliusweig: The idea to use git for this job is on me. I think it can be helpful because `git diff` can report added/changed/deleted file names.
However, just working with sorted lists is probably the simplest solution of all and does not have to rely on git's output format. I'd be ok either way.

artmello:
@ahmetb / @corneliusweig 

Thanks a lot for the detailed review. I tried to address all your comments, please, let me know if I missed some.
I reverted gitutil changes and used suggested approach of executing a diff on indexes from pre/pos update. Also I moved to use a less verbose output as suggested.

Here is a sample output:
```
$ cd -
~/.krew/index

$ git reset --hard dc075b6355c02e30aa0a43bebf4587bab2eafb9d
HEAD is now at dc075b6 update neat to v1.1.0 (#399)

$ cd -
~/dump/krew

$ ./out/bin/krew-darwin_amd64 update
  New plugins available: advise-psp v1.4.6, deprecations v0.2.4, duck v0.0.1, net-forward v1.0.3, profefe v0.9.0, resource-snapshot v0.1.2, spy v0.3.1, tree v0.3.1
  Upgrades available: pod-dive v0.1.3 -> v0.1.4
```

ahmetb:
Can I just say: I LOVEEEEE how small this PR is. 

I have only 2 concerns atm:
- when first ever “update” is done, does it print everything as new (ideally should not)
- it seems like the indentation in the output isnt that useful if the plugin list is gonna just flow over to the next line. I wonder if we can do something like print max 4-5 on each line and just add \n\t after each 5?

I am on the fence about whether we should add integration tests for this? Probably not in this PR since it’s not critical functionality.

ahmetb:
Also it seems like there are a couple lint and test errors.

ahmetb:
@artmello pinging if you've still got time to pursue. it seems like we're pretty close :)

artmello:
Hi @ahmetb, once again thanks for reviewing it and sorry for the sparse answers. I believe all your last comments are addressed with latest commits. Output has changed, now we are splitting plugins list on chuncks  of 5.

```
$ cd -
~/.krew/index

$ git reset --hard dc075b6355c02e30aa0a43bebf4587bab2eafb9d
HEAD is now at dc075b6 update neat to v1.1.0 (#399)

$ cd -
~/dump/krew

$ ./out/bin/krew-darwin_amd64 update
  New plugins available: advise-psp v1.4.6, deprecations v0.2.4, duck v0.0.1, net-forward v1.0.3, profefe v0.9.0
	resource-snapshot v0.1.2, spy v0.3.3, tree v0.4.0
```

ahmetb:
/lgtm
/approve
Thanks a bunch. This was cool work. 

I might write a few integration tests for this.

corneliusweig:
This is great! Thanks for your contribution @artmello

artmello:
Thanks a lot @ahmetb and @corneliusweig for all the help on it. Looking forward to contribute again soon :)

Issue #454, opened 2020-01-02
Plugins installed via --manifest are blocking upgrades (and others are re-downloaded every time)


ahmetb:
See output below.

There are two problems here:

1. The `snapshot` plugin is installed from `--manifest`, it's not in the krew-index yet, but for some reason, it's blocking the upgrade of other plugins. 

1. If `upgrade` fails (after downloading some plugins), those plugin receipts are not written to FS (for some reason?), so if I re-run this command, all the available updates are being downloaded/installed **again**. (ctrl-f `Installing new version` below).

```
$ kubectl krew upgrade -v=10
I0102 13:15:32.589848   42242 update.go:45] Updating the local copy of plugin index (/Users/ahmetb/.krew/index)
I0102 13:15:32.589940   42242 git.go:73] Going to run git fetch -v
From https://github.com/kubernetes-sigs/krew-index
 = [up to date]      master     -> origin/master
 = [up to date]      v0.2.1     -> origin/v0.2.1
I0102 13:15:33.246522   42242 git.go:73] Going to run git reset --hard @{upstream}
HEAD is now at 9011a4e tree v0.2.0 (#410)
I0102 13:15:33.254537   42242 git.go:73] Going to run git clean -xfd
Updated the local copy of plugin index.
I0102 13:15:33.259907   42242 util.go:38] Found 65 install receipts in /Users/ahmetb/.krew/receipts
I0102 13:15:33.260584   42242 util.go:45] parsed receipt for access-matrix: version=v0.4.2
I0102 13:15:33.260852   42242 util.go:45] parsed receipt for auth-proxy: version=v0.7.0
I0102 13:15:33.261027   42242 util.go:45] parsed receipt for bulk-action: version=v1.1.0-beta
I0102 13:15:33.261179   42242 util.go:45] parsed receipt for ca-cert: version=v0.0.0
I0102 13:15:33.261400   42242 util.go:45] parsed receipt for change-ns: version=v1.0.0
I0102 13:15:33.261633   42242 util.go:45] parsed receipt for config-cleanup: version=v0.5.0
I0102 13:15:33.261791   42242 util.go:45] parsed receipt for cssh: version=v1.0.1
I0102 13:15:33.261939   42242 util.go:45] parsed receipt for ctx: version=v0.7.1
I0102 13:15:33.262106   42242 util.go:45] parsed receipt for custom-cols: version=v0.0.3
[...]

PR #458, opened 2020-01-04
Fix upgrade if a plugin installed from manifest

ferhatelmas:
If a plugin isn't in index, upgrade for all doesn't fail anymore.
Also, added an integration test by modifying valid plugin `konfig` to `konfig-non-index`.

Related to #454 problem 1.


integration_test/testutil_test.go
```diff
@@ -37,8 +37,9 @@ import (
 const (
 	persistentIndexCache = "krew-persistent-index-cache"
 	krewBinaryEnv        = "KREW_BINARY"
-	validPlugin          = "konfig" // a plugin in central index with small size
-	validPlugin2         = "mtail"  // a plugin in central index with small size
+	validPlugin          = "konfig"           // a plugin in central index with small size
+	validPlugin2         = "mtail"            // a plugin in central index with small size
+	validNonIndexPlugin  = "konfig-non-index" // a plugin not in central index with small size
```
ahmetb: instead of doing this let's do this:

1. test.WithIndex()
2. find konfig.yaml in the index path, delete it
3. use upgrade --no-update-index in both `install` and `upgrade` calls in the integration test.

this way:
- you can fake the situation that konfig is not in the index.
- without having to duplicate a testdata file (I really dislike increasing testdata)

ferhatelmas: Sounds much better, updated.

corneliusweig:
/lgtm
/approve
..and sorry for this :snail: response

PR #461, opened 2020-01-07
Migrate to GitHub workflows

### LINKED ISSUE #451: Move away from Travis CI to GitHub Workflows
/priority P3

Travis CI at this point seems to be a ticking time bomb. It would be better to migrate to GitHub Workflows. I don't see any particular downsides (we can test/run neither locally). Migration would be the only cost.

Advantages:
- no need to _allow_ an app to kubernetes-sigs org, GH Actions don't require permissions
- no need for encrypted secrets
- we can probably move to `goreleaser` and simplify some of the asset building, stamping (with version+gitinfo) and uploading (to GH Releases) as part of this.
- we can more easily separate the stages of build (e.g. lint, build, tests, release...) with a compartmentalized UI, compared to Travis.

Up for grabs and discussion.

alonyb:
This PR contains a new pipeline for this project with GitHub Workflows

There are 2 files, one for PUSH/PR and other for release.

PUSH & PR FILE:

- Run all travis commands before deploying stage

RELEASE FILE:

- Run all travis commands including deploying stage
- For creating a new release, it must have a tag in the commit and then push the tag, and the pipeline will create a release automatically.

To see logs about the workflow see the fork on my repo [here](https://github.com/RubenBaez/krew/actions) , both config files are included

This PR will help with this  #451


.github/workflows/release.yml
```diff
@@ -0,0 +1,59 @@
...
+          hack/verify-code-patterns.sh
+          hack/verify-boilerplate.sh
+          hack/run-lint.sh
+          go test -short -coverprofile=coverage.txt -covermode=atomic ./...
+          hack/make-all.sh
+          hack/verify-installation.sh
+          hack/ensure-kubectl-installed.sh
+          hack/run-integration-tests.sh
+          hack/verify-receipts-upgrade-migration.sh
+
```
ahmetb: let's split these steps with proper names. :)

.github/workflows/release.yml
```diff
@@ -0,0 +1,59 @@
...
+      # temporary fix
+      # see https://github.com/actions/setup-go/issues/14
+        run: |
+          echo "##[set-env name=GOPATH;]$(dirname $GITHUB_WORKSPACE)"
+          echo "##[add-path]$(dirname $GITHUB_WORKSPACE)/bin"
+        shell: bash
+    
+      - name: Prepare environment
+        run: |
+          go mod download
```
ahmetb: I think (and maybe I'm wrong) you'll need && between these lines. Otherwise a command will fail but next one will continue to execute.

ahmetb:
cc: @corneliusweig @ferhatelmas thoughts? The execution times are comparable, the logs UI is pretty decent and comparable as well.

.github/workflows/release.yml
```diff
@@ -0,0 +1,59 @@
...
+          prerelease: false
+      
+      - name: Upload binaries to release
+        uses: svenstaro/upload-release-action@v1-release
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: out/krew.*
+          tag: ${{ github.ref }}
+          overwrite: true
+          file_glob: true
```
ahmetb: it seems like this is the only relevant part and the first ~40 lines are duplicated from ci.yml.

is there a way to de-duplicate?

.github/workflows/release.yml
ferhatelmas: From docs:
> Each run keyword represents a new process and shell in the runner environment. When you provide multi-line commands, each line runs in the same shell.

Seems not needed.

.github/workflows/release.yml
```diff
@@ -0,0 +1,59 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@master
+        
+      - name: Set GOPATH
+      # temporary fix
+      # see https://github.com/actions/setup-go/issues/14
+        run: |
+          echo "##[set-env name=GOPATH;]$(dirname $GITHUB_WORKSPACE)"
+          echo "##[add-path]$(dirname $GITHUB_WORKSPACE)/bin"
```
ferhatelmas: Where does it come from?

https://stackoverflow.com/questions/59236611/github-actions-run-golint-on-wihdows/59242962#59242962

.github/workflows/release.yml
```diff
@@ -0,0 +1,59 @@
...
+name: Handle Release
+jobs:
+  build:
+    name: Upload Release Asset
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@master
+        
+      - name: Set GOPATH
```
ferhatelmas: this isn't gopath but gobin setup.

.github/workflows/release.yml
RubenBaez: The problem I found is GitHub actions does not have option to ejecute multiples Jobs, I mean I did not found a way to join build stage and deploy stage in the same file because, it runs in different ways(push, PR, tag), so the solution by GitHub comunity is run it in different files. That's why I had to duplicate that part, to deploy artifacts.

ferhatelmas: if trigger is extended, then `if conditional` should work but not familiar with the exact syntax. 

A suggestion:

`if: github.event_name == 'push' && github.ref == 'v\d+\.\d+.\d+.*'`

.github/workflows/release.yml
ferhatelmas: maybe best, set `GOPATH and GOBIN`

.github/workflows/release.yml
RubenBaez: GitHub actions have some issues in env variables, so there are some solutions wrote by GitHub community, I wrote one of them, your option is another good option. Those are temporary solutions until GitHub takes out a new way to do it or get a solution.

.github/workflows/release.yml
ahmetb: Without seeing proof, I won't be convinced :) If they run in the same shell, it most likely means they're separated with `;` and not `&&`. Try

```
false
date
```

you should be seeing a failure not the date.

.github/workflows/release.yml
ahmetb: I ran to this problem earlier as well

https://github.com/ahmetb/kubectl-tree/blob/master/.github/workflows/ci.yml
https://github.com/ahmetb/kubectl-tree/blob/master/.github/workflows/release.yml

and I don't like the duplication :( Let's keep searching?

ahmetb:
Before

ahmetb:
(sorry for accidentally closing)

Before we invest too much into this PR, it seems like: It is doable, but it's not as clean as I expected, so I'll have other maintainers also express how they feel about it.

.github/workflows/release.yml
RubenBaez: Sure, I'll keep searching for a while.
@ahmetb Anyway, this still needs much to improve but my suggestion is to move forward in parallel with the improvements of GitHub, I could follow this process for a long time. And see if we can improve the other Kubernetes projects with GitHub actions.

.github/workflows/release.yml
corneliusweig: @RubenBaez have you checked how github handles multi-line scripts?

.github/workflows/push.yml
```diff
@@ -0,0 +1,47 @@
...
+      with:
+        go-version: 1.13
+      id: go
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v1
+    
+    - name: Set GOPATH
+      # temporary fix
+      # see https://github.com/actions/setup-go/issues/14
```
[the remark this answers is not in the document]
RubenBaez: I am looking for a solution for this since GitHub has problems with the environment variables, the solution I wrote found it [here](https://github.com/actions/setup-go/issues/12#issuecomment-524631719), as soon as I find a more appropriate solution or GitHub fix that, I fix it in this part

.github/workflows/push.yml
```diff
@@ -0,0 +1,47 @@
...
+    steps:
+
+    - name: Set up Go 1.13
+      uses: actions/setup-go@v1
+      with:
+        go-version: 1.13
+      id: go
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v1
```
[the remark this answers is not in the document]
RubenBaez: Of course, no problem, it's fine.

.github/workflows/push.yml
```diff
@@ -0,0 +1,47 @@
...
+        echo "##[set-env name=GOPATH;]$(dirname $GITHUB_WORKSPACE)"
+        echo "##[add-path]$(dirname $GITHUB_WORKSPACE)/bin"
+      shell: bash
+    
+    - name: Prepare environment
+      run: |
+        pwd
+        echo $GOPATH
+        echo $PATH
+        go env
```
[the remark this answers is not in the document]
RubenBaez: Sure, I missed it. Thank you.

.github/workflows/release.yml
```diff
@@ -0,0 +1,59 @@
+on:
+  push:
+    # Sequence of patterns matched against refs/tags
+    tags:
+    - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
```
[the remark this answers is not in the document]
RubenBaez: Good observation. :)

.github/workflows/release.yml
```diff
@@ -0,0 +1,59 @@
...
+name: Handle Release
+jobs:
+  build:
+    name: Upload Release Asset
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
```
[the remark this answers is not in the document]
RubenBaez: Yes, I added it, no problem :)

ahmetb:
> However, the config duplication between `{push,release}.yaml` worries me. We should try harder to combine the two files, or we need to seriously think if it's really worth it to migrate to GH actions.

+1 I think this will be a deal breaker for me as well, and the temp hacks/workarounds in other places to use this is just the icing on the cake.

.github/workflows/release.yml
RubenBaez: It's done, you can check it.

.github/workflows/release.yml
RubenBaez: Yes, I did. You can check it now :)

RubenBaez:
@ahmetb @corneliusweig @ferhatelmas I've combined the 2 files into one, but in one single stage, you can check logs, when it's not a tag [here](https://github.com/RubenBaez/krew/runs/386140537) and when it's a tag [here](https://github.com/RubenBaez/krew/runs/386141567).
It would be better in 2 stages(for me), `build` and `release`, as [here](https://github.com/RubenBaez/krew/runs/386136242), but that means duplicate steps to generate binaries and upload them. I would like to know your opinion.

ahmetb:
@RubenBaez thanks for your work.
Honestly the two examples (with all-in-one but some steps skipped) and separate look OK from an output consumption perspective. At this point, it's probably just a question of whether we'd be comfortable taking on maintaining the workflow YAMLs.

ahmetb:
@RubenBaez in the meanwhile, I've opened [this issue](https://github.com/kubernetes-sigs/krew-index/issues/430) suggesting we can move `krew-index` to GitHub Workflows. It should be much more straightforward since it's a [smaller workflow](https://github.com/kubernetes-sigs/krew-index/blob/438afa810e98396d88983a2239919a8c28c0f864/.travis.yml). 

Would you have cycles to do that? I think we can merge that much more comfortably.

RubenBaez:
@ahmetb 
I can maintain it for a while and see that there are no problems :), and sure I will check it.

.github/workflows/release.yml
corneliusweig: @RubenBaez Btw, I tried this out and it looks like @ahmetb was correct. If one line in a multi-line run script fails, the following lines _will be exectued regardless_.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,92 @@
...
+    - name: Install gox
+      run: |
+        go mod download
+        hack/install-gox.sh\
+
+    - name: Install dependencies
+      run: go mod tidy && git diff --no-patch --exit-code
+      
+    - name: Verify code quality
+      run: |
```
corneliusweig: If one of the scripts in this multi-line run statement fails, the following statements will be executed regardless.

In this case, this is actually fine (and even desired), so no need for action ;)

corneliusweig: well actually... I this _is_ a problem, because the success status of the whole block will be the return value of the last line. So an error in one of the previous lines will not propagate and halt the pipeline. That's not good :/

ahmetb: Yeah I'm almost never cool with that.

Similarly I can guess if you make a typo in a $variable_name, it won't error out. This is [bash strict mode](http://redsymbol.net/articles/unofficial-bash-strict-mode/) all over again.

For these situations, we should prefer creating hack/*.sh scripts that adhere to the same set of practices.

corneliusweig: Yeah, putting everything in bash scripts sounds like a manageable solution.

alonyb: Well, I read some solutions, run those files in differents steps, run it with a bash statement like this https://github.com/kubernetes-sigs/krew-index/pull/437#pullrequestreview-344230430, or finally create a new file *.sh and run all those steps. I would like to know your opinion. Then I could develop it :)

ahmetb: Yeah we can split all scripts into different `steps[]`. The fact that you have a && in the name suggests that they are different steps.

Bash carries out exit code of "last" command (in this case it would be the "only" command if we do 1 command per step).

But when it doesn't make sense to split multiple commands into two different steps, we should not shy away from using `set -ex`.

corneliusweig: `set -u` seems also advisable (see discussion in krew-index)

corneliusweig: > or finally create a new file *.sh and run all those steps

I personally think that we have too many bash scripts already. So if we can avoid adding another one that is only needed for the CI, I would prefer that.

ahmetb: +1 separating each script into a step (just like they were in .travis.yml) is much better than a single opaque script.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,92 @@
...
+        echo "##[add-path]$(dirname $GITHUB_WORKSPACE)/bin"
+      shell: bash
+    
+    - name: Install gox
+      run: |
+        go mod download
+        hack/install-gox.sh\
+
+    - name: Install dependencies
+      run: go mod tidy && git diff --no-patch --exit-code
```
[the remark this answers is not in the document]
alonyb: It's done and removed `&&` and used `set pipefail`

.github/workflows/ci.yml
```diff
@@ -0,0 +1,92 @@
...
+      # temporary fix
+      # see https://github.com/actions/setup-go/issues/14
+      run: |
+        echo "##[set-env name=GOPATH;]$(dirname $GITHUB_WORKSPACE)"
+        echo "##[add-path]$(dirname $GITHUB_WORKSPACE)/bin"
+      shell: bash
+    
+    - name: Install gox
+      run: |
+        go mod download
```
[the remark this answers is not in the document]
alonyb: I agree

.github/workflows/ci.yml
alonyb: I've separated some steps, and others used `set -euo pipefail` as `krew-index`.

.github/workflows/release.yml
alonyb: It's only GOPATH

corneliusweig:
Ok, to me this looks quite good now.

There are two more things that we need to take care of:
- We used to export code coverage to codecov.io, but the pipeline no longer does it. There is a GH action https://github.com/codecov/codecov-action for this, but it requires a token. Or would the usual `bash <(curl -s https://codecov.io/bash)` way work without a token?

- We have no `GITHUB_TOKEN` configured, but the release step needs one. This is clearly our job :wink:

ahmetb:
We can remove Codecov altogether, I haven't been finding it super useful to be honest.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,94 @@
...
+  push:
+    branches: 
+      - '*'
+    tags: 
+      - 'v*.*.*'
+  pull_request:
+    branches: 
+      - '*'
+  release:
+    types: [published, created, edited]
```
ahmetb: I think we don't need this if we do `push.tags` ?
Can you explain how it helps?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,94 @@
...
+
+jobs:
+
+  build:
+    name: Build and Tags
```
ahmetb: you can drop `name:` since there's only 1 job.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,94 @@
...
+    - name: Set up Go 1.13
+      uses: actions/setup-go@v1
+      with:
+        go-version: 1.13
+      id: go
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+    
+    - name: Set GOPATH
```
ahmetb: the name isn't reflecting what's happening below.
It sounds more like "Add GOPATH/bin to PATH"

.github/workflows/ci.yml
```diff
@@ -0,0 +1,94 @@
...
+        echo "##[set-env name=GOPATH;]$(dirname $GITHUB_WORKSPACE)"
+        echo "##[add-path]$(dirname $GITHUB_WORKSPACE)/bin"
+    
+    - name: Install dependencies
+      run: go mod download
+    
+    - name: Install gox
+      run: hack/install-gox.sh
+      
+    - name: Verify code quality
```
ahmetb: What if we just separated each of these commands to separate `steps`?

Because in this mode, I can't tell (from the output) which one is failing.

In travis we had that.

Since the script names are quite self descriptive, you don't even need to specify `name:` for most.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,94 @@
...
+      run: |
+        set -euo pipefail
+        go mod tidy 
+        git diff --no-patch --exit-code
+        hack/verify-code-patterns.sh
+        hack/verify-boilerplate.sh
+        hack/run-lint.sh
+        
+    - name: Unittest and coverage
```
ahmetb: call it "Run unit tests"

.github/workflows/ci.yml
```diff
@@ -0,0 +1,94 @@
...
+    - name: Make binaries && verify krew installation
+      run: hack/make-all.sh
+        
+    - name: Ensure kubectl installed
+      run: hack/ensure-kubectl-installed.sh
+        
+    - name: Integration tests
+      run: |
+        set -euo pipefail
+        hack/verify-installation.sh
```
ahmetb: again, consider separating into two different steps, as they do different things.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,94 @@
...
+        hack/run-integration-tests.sh
+    
+    - name: Receipts upgrade migrations
```
ahmetb: You don't need names for these, not only they aren't quite accurate, they're just repeating what's in the script name.

.github/workflows/ci.yml
alonyb: When editing the releases either from the GitHub interface, the pipeline will also run, but I also see that it is not necessary. I already removed it.

.github/workflows/ci.yml
alonyb: Sure, It's more clear.

.github/workflows/ci.yml
alonyb: > Because in this mode, I can't tell (from the output) which one is failing.
> 
I don't like the idea of separating all steps for the dead times when it changes of step. But, I agree with this. I just separated all the steps, check it, please.

.github/workflows/ci.yml
alonyb: Sure, np :)

.github/workflows/ci.yml
alonyb: It's done.

.github/workflows/ci.yml
alonyb: You are right, but all steps need a name and it is part of keeping the entire structure of the pipeline a little clear, later you can add some other command, among other things.

alonyb:
> Ok, to me this looks quite good now.
> 
> There are ~two~one more things that we need to take care of:
> 
> * We used to export code coverage to codecov.io, but the pipeline no longer does it. There is a GH action https://github.com/codecov/codecov-action for this, but it requires a token. Or would the usual `bash <(curl -s https://codecov.io/bash)` way work without a token?
> * ~We have no `GITHUB_TOKEN` configured, but the release step needs one. This is clearly our job 😉~ GH actions does this automatically, see https://github.com/actions/create-release#example-workflow---create-a-release

@corneliusweig 
As @ahmetb said 
> We can remove Codecov altogether, I haven't been finding it super useful, to be honest.
I didn't put it in the pipeline. Therefore. If you change your mind. Or then it becomes necessary, I could add that step :)

corneliusweig:
Alright, let's put this on trial. As discussed, we'll run travis alongside for a while, see if the results are identical, and watch out for maintenance overhead.

Thanks for pushing this through! @alonyb 

/lgtm
Since it's quite a delicate change, let's wait for a :+1: from @ahmetb.

ahmetb:
/approve
Thanks for the continuous work here. I hope this works out.

corneliusweig:
Hey @alonyb,

I have one final request: You seem to follow closely how GH actions evolve. Should you learn that the env variable workaround (https://github.com/kubernetes-sigs/krew/pull/461/files#diff-e9f950f17198d3d5e3122a44230a09b9R32) can be done in a cleaner way (or can be skipped), could you open a follow-up PR for that?

alonyb:
@corneliusweig 
Sure, I'm going to do some research and open the PR. No problem.

Issue #328, opened 2019-09-02
Add a link to sample cli plugin repo


ferhatelmas:
As suggested at https://github.com/kubernetes-sigs/krew/pull/323#pullrequestreview-282335218

/kind documentation

corneliusweig:
:+1: Somewhere in the vicinity of `docs/DEVELOPER_GUIDE.md` line 32 would fit nicely.

chriskim06:
I can take care of this if its still needed

/assign

PR #466, opened 2020-01-17
Add link to sample-cli-plugin in developer guide

chriskim06:
Let me know if the wording of this should be changed at all.

Fixes https://github.com/kubernetes-sigs/krew/issues/328


ahmetb:
I think there are some changes we made to the sample-cli-plugin that aren’t merged still after such a long time. In that case we don’t need to link out until they’re merged.

chriskim06:
Sounds good, should I just close this?

ahmetb:
lets keep it open.
I'm thinking a major Developer Guide overhaul. Would be useful to incorporate there perhaps.
/lifecycle frozen

ahmetb:
Honestly I think it's ok sample-cli-plugin doesn't mention Krew in this case.
Krew has enough momentum now.

/lgtm
/approve

PR #471, opened 2020-01-26
Update README.md - install krew with homebrew

arslanbekov:
I think many people be happy to install `krew` with homebrew.
I just add this in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,12 @@ Check out the list of [kubectl plugins available on krew][list] or just run
 
 > :warning: **Warning:** krew is only compatible with kubectl v1.12 or higher.
 
+**Homebrew:**
+
+You can install krew using [brew](https://brew.sh/):
+
+    brew install krew
+
 **macOS and Linux:**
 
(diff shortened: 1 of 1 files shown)


arslanbekov:
Ok, no problem.
Thanks for clarifying @corneliusweig

ahmetb:
Oh no. 
We can’t support that installation method as krew is supposed to install and update itself. 

@corneliusweig can you take up the task of working with Brew maintainers to remove Krew from Homebrew formula list? Maybe we can detect and refuse to work as well but that’s not preferable.

immae1:
@corneliusweig could you please explain how i could update the path for the brew installed krew? I need the brew installation because i install some other plugins for docker with a ansible playbook. I will keep this simple

corneliusweig:
@immae1 Oh no.. looks like homebrew is still distributing the old version v0.3.3.  Otherwise you would have seen a warning what to do.

Basically you need to do step three from our installation instructions https://github.com/kubernetes-sigs/krew/#installation

Issue #23, opened 2018-08-06
Support custom index repositories


ahmetb:
Krew should support customizable index repositories.

This would allow users to distribute plugins without requiring them to make a pull request to the centralized index repo like "krew-index".

Similarly, this would allow vendors to distribute their plugins that are too specific or one-off to be in the centralized index repo.

Work here is probably:
- Determine/propose a user experience journey
- Propose a design
- Implement CLI commands to tap/untap into other index repos
- Refactor `update`/`upgrade` paths to keep taps in mind
- Write user/developer documentation on working with custom repos.

@lbb has showed interest to work on this.

ahmetb:
There's a design at https://docs.google.com/document/d/1fG3ST0QbadQGcxNiFAZ9GhX_RZ8pm3piT99UK3FbT88/edit?ts=5b75a2a6 if anyone wants to review.

juanvallejo:
@melaniecebula I've started a [draft here](https://docs.google.com/document/d/1RMCFAKcJZJbdQGgTtb9_JkBzTwSIkBkmWRdKmt1v9EU/edit?usp=sharing) to begin gathering thoughts on how to begin implementing this. Feel free to contribute or leave feedback on there

ahmetb:
@juanvallejo the link is not available for public access

jekohk:
hey, wondering what the next steps are for pushing this forward?

we've been writing many kubectl plugins internally which are specific to our stack, and being able to distribute them with krew would be a huge improvement over the manual process we use now.

ahmetb:
We are moving ahead with the concrete proposal at #483.

PR #478, opened 2020-01-28
Add support for custom indices

chriskim06:
I'm in the process of adding tests currently. I wanted to open the PR to write down the approach and maybe get some feedback. So here's a summary of the changes:
- added a new `index` subcommand
  - `kubectl krew index`: lists the configured custom indices
  - `kubectl krew index add custom-index https://github.com/custom/index.git`: adds a custom index with the name `custom-index`
  - `kubectl krew index remove custom-index`: removes the custom index with the name `custom-index`
- a file is stored at `$KREW_ROOT/indexconfig.yaml` that contains an object whose key/values are a user defined name and URI of the index being added
- custom indices are cloned to `$KREW_ROOT/custom/{index name}`
- plugins from custom indices are installed in `$KREW_ROOT/store/krew-indices/{index name}/{plugin}`
- plugin receipts for plugins installed from a custom index are stored `$KREW_ROOT/receipts/{index name}/{plugin}.yaml`
- plugins from custom indices need to be explicitly installed: `kubectl krew install custom-index/plugin` (this is similar to how its handled in [homebrew](https://docs.brew.sh/Taps); a follow up issue could be to handle plugin name collisions, but for this I only implemented allowing one or the other)
- list/search add plugins from custom indices to the end of the list:
```
$ kubectl krew list
PLUGIN           VERSION
ctx              v0.7.1
grep             v1.2.2
kudo             v0.10.0
ns               v0.7.1
whoami           v0.0.29
private/warp     v0.0.2

$ kubectl krew search
NAME                                    DESCRIPTION                                         INSTALLED
...
who-can                                 Shows who has RBAC permissions to access Kubern...  no
whoami                                  Show the subject that's currently authenticated...  yes
private/access-matrix                   Show an RBAC access matrix for server resources     no
private/auth-proxy                      Authentication proxy to a pod or service            no
private/warp                            Sync and execute local files in Pod                 yes
...
```
- update/upgrade have been modified to look in the correct directory for plugins
  - plugins from custom indices must be explicit when upgrading them specifically: `kubectl krew upgrade custom-index/plugin`
  - no arg upgrade will look in the correct directory for plugins that were installed from custom indices

I think that covers most things at a high level. I took a lot of ideas from `brew tap` in working on this feature. Its definitely a rough first draft and I would love to get some feedback. I chose to do things in certain ways to try to make it as backwards compatible as possible. This is a really important feature in order for me to use this at work (which I'd love to do).

### DIFF
--- a/cmd/krew/cmd/index.go
+++ b/cmd/krew/cmd/index.go
@@ -16,11 +16,14 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/spf13/cobra"
 
 	"sigs.k8s.io/krew/internal/index/indexoperations"
 )
+
+var indexConfig *indexoperations.IndexConfig
 
 // indexCmd represents the index command
 var indexCmd = &cobra.Command{
@@ -37,12 +40,7 @@ var indexAddCmd = &cobra.Command{
 	Short: "Add a custom index to download plugins from",
 	Long:  "Add a custom index to download plugins from",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		fmt.Printf("index add: %+v\n", args)
-		indexConfig, err := indexoperations.GetIndexConfig()
-		if err != nil {
-			return err
-		}
-		return indexConfig.WriteIndexConfig(args[0], args[1])
+		return indexConfig.AddIndex(args[0], args[1])
 	},
 }
 
@@ -51,10 +49,6 @@ var indexRemoveCmd = &cobra.Command{
 	Short: "Remove a custom index that you've added",
 	Long:  "Remove a custom index that you've added",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		indexConfig, err := indexoperations.GetIndexConfig()
-		if err != nil {
-			return err
-		}
 		return indexConfig.RemoveIndex(args[0])
 	},
 }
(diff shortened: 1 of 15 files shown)


ahmetb:
@chriskim06 While I totally appreciate you writing code, it's most likely that this is the BIGGEST work item since the inception of Krew.

We didn't discuss this at all and it's been on our list to **design** this first. Code is not important (yet). So I will close this PR, and ask you to open an issue with a Design proposal first.

We care about:

- user experience
- developer experience
- migration path

more than the code. Let's discuss a design first, then we can revisit your patch.

In Krew project, it's very unlikely we'll accept non-trivial patches of this size without discussing how it changes the user experience first.

/close

chriskim06:
@ahmetb that makes sense, I'll open up a new issue with a more in depth design proposal. Sorry for doing these things out of order 😅

chriskim06:
@ahmetb I'm not at all surprised, definitely knew it was possible for this to get closed when I opened the PR. I'm really looking forward to being able to use krew at work but custom indices is a must before that can happen (I'm sure that's the case for a lot of people).

I'm still new to open source so sorry again if I'm going about things out of order/in a way that isn't conventional. Lesson learned though, in the future I will try to communicate these things through issues to you and the other maintainers!

ahmetb:
That's not at all a problem.
Please open a new issue with your proposal and refer to #23. We haven't collected enough use cases about this.

PR #479, opened 2020-01-28
integration tests for showing new/upgradeable plugins

ahmetb:
Tests when/how we show "New plugins available" and "Upgrades available for
installed plugins".

Adds ~3 seconds to integration tests.

### DIFF
--- a/integration_test/update_test.go
+++ b/integration_test/update_test.go
@@ -14,7 +14,18 @@
 
 package integrationtest
 
-import "testing"
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"testing"
+
+	"sigs.k8s.io/krew/internal/environment"
+	"sigs.k8s.io/krew/pkg/constants"
+)
 
 func TestKrewUpdate(t *testing.T) {
 	skipShort(t)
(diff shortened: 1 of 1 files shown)


integration_test/update_test.go
```diff
@@ -34,3 +47,62 @@ func TestKrewUpdate(t *testing.T) {
...
 	}
 }
+
+func TestKrewUpdateListsNewPlugins(t *testing.T) {
+	skipShort(t)
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	test = test.WithIndex()
+	if err := os.Remove(filepath.Join(test.Root(), "index", "plugins", validPlugin+constants.ManifestExtension)); err != nil {
```
corneliusweig: Better use `environment.NewPaths(test.Root()).IndexPluginsPath()`.

Same below

integration_test/update_test.go
```diff
@@ -34,3 +47,62 @@ func TestKrewUpdate(t *testing.T) {
...
+func TestKrewUpdateListsUpgradesAvailable(t *testing.T) {
+	skipShort(t)
+
+	test, cleanup := NewTest(t)
+	defer cleanup()
+	test = test.WithIndex()
+
+	// set version of some manifests to v0.0.0
+	modifyManifestVersion(t, filepath.Join(test.Root(), "index", "plugins", validPlugin+constants.ManifestExtension), "v0.0.0")
+	modifyManifestVersion(t, filepath.Join(test.Root(), "index", "plugins", validPlugin2+constants.ManifestExtension), "v0.0.0")
```
corneliusweig: I don't quite understand why you use two plugins here. Is it just to ensure that the output lists all available upgrades? For me that's ok, but not necessary.

integration_test/update_test.go
ahmetb: Normally I'd say this is valid. But tests ideally should rely on these utilities a lot less, as that's sort of what they're testing.

In this case, I guess we can use the utility as that's not what we're testing.

integration_test/update_test.go
ahmetb: honestly no need, but also no harm.

ahmetb:
> You didn't write a test for when the installed plugin has no upgrade available, so it should not be listed in the output. This case is somewhat covered by unit tests, so we probably don't need it.

actually that's valid, maybe I'll refactor the 2-plugin installing test to do that.

ahmetb:
done PTAL.

Issue #364, opened 2019-11-04
Problems running tests in MacOS Mojave


JJJJJones:
hack/run-tests.sh breaks because the Mac grep implementation doesn't behave like GNU. I [wrote some code](https://github.com/JJJJJones/krew/blob/db1ee4d7d1b542060dd72badfef7ab23dd1542e1/hack/verify-code-patterns.sh#L19) which can solve it.
```
➜  krew git:(356) ✗ hack/run-tests.sh
...
Running linter
Check code patterns
grep: warning: recursive search of stdin
```

hack/run-integration-tests.sh breaks because readlink won't handle non-symbolic links. From `man readlink`:

```
     When invoked as readlink, only the target of the symbolic link is printed.  If the given argument is
     not a symbolic link, readlink will print nothing and exit with an error.
```

To solve this I replace [this line](https://github.com/kubernetes-sigs/krew/blob/master/hack/run-integration-tests.sh#L41) with below.

```bash
if [[ $(uname) == "Darwin" ]]; then
  krew_binary_realpath="$(stat -f "%N" "${KREW_BINARY}")"
else
  krew_binary_realpath="$(readlink -f "${KREW_BINARY}")"
fi
```

ahmetb:
Since as developers of Krew we rely on GNU grep (even on macOS), I wish there was an easier solution here (smh).

Is there a chance "stat -f" does exactly that on linux, too? It would be really good to have 1 way of doing things.

/priority P3

JJJJJones:
> Since as developers of Krew we rely on GNU grep (even on macOS), I wish there was an easier solution here (smh).

Then GNU grep (ggrep) is a requirement for Mac users.

> Is there a chance "stat -f" does exactly that on linux, too? It would be really good to have 1 way of doing things.

Unfortunately no. Macs uses `-f` and `--format` isn't an accepted flag. [Linux](https://linux.die.net/man/1/stat) uses `--format` and `-f` is a separate flag. Both can use the same `"%N"` for formatting though.

ahmetb:
So here's normally what we do in krew repo when we encounter a situation like this:

We implement this "readlink" functionality in one place. Then we update `verify-code-patterns.sh` to make sure nobody can use `readlink`/`stat` anywhere else but that particular file.

Otherwise, we'll see other scripts running making the same mistake and this issue repeating itself. If you want to go that route, that's the best, otherwise I recommend we leave it as-is (and maybe make a note that gstat is the one that's needed).

[...]

PR #482, opened 2020-01-29
GNU implementation's in scripts to support MacOS

JJJJJones:
#364 

MacOS grep/readlink/stat are insufficient for certain scripts so this wraps calls to those with a function that ensures the existence of and calls the GNU variant for Mac clients.

### DIFF
--- a/hack/resolve-grep.sh
+++ b/hack/resolve-grep.sh
@@ -3,7 +3,7 @@ function resolvegrep () {
     if [[ $(whereis ggrep) -eq 0 ]]; then
       ggrep $@
     else
-      echo "GNU grep not found!"
+      echo "GNU grep not found - please install GNU coreutils from brew.sh (brew install coreutils)"
       exit 1
     fi
   else
--- a/hack/resolve-grep.sh
+++ b/hack/resolve-grep.sh
@@ -1,0 +1,12 @@
+function resolvegrep () {
+  if [[ $(uname) == "Darwin" ]]; then
+    if [[ $(whereis ggrep) -eq 0 ]]; then
+      ggrep $@
+    else
+      echo "GNU grep not found!"
+      exit 1
+    fi
+  else
+    grep $@
+  fi
+}
--- a/hack/resolve-link.sh
+++ b/hack/resolve-link.sh
@@ -1,0 +1,12 @@
+function resolvelink () {
+  if [[ $(uname) == "Darwin" ]]; then
+    if [[ $(whereis greadlink) -eq 0 ]]; then
+      greadlink -f $@
+    else
+      echo "GNU readlink not found"
+      exit 1
+    fi
+  else
+    readlink -f $@
+  fi
+}
--- a/hack/resolve-link.sh
+++ b/hack/resolve-link.sh
@@ -3,7 +3,7 @@ function resolvelink () {
     if [[ $(whereis greadlink) -eq 0 ]]; then
       greadlink -f $@
     else
-      echo "GNU readlink not found"
+      echo "GNU readlink not found - please install GNU coreutils from brew.sh (brew install coreutils)"
       exit 1
     fi
   else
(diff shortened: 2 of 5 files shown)


ahmetb:
thanks for being thoughtful @JJJJJones how about we do this:

Instead of writing code to support two implementations, we can write a `ensure-gnu-tools.sh` that we `source [...]` in scripts where readlink etc are used.
Since number of Krew contributors using macOS is <100, we can just ask them to `brew install coreutils findutils grep [...]` whatever in an error message if we don't detect `ggrep`, `greadlink` etc in PATH.

JJJJJones:
@ahmetb I realize now the PR is misleadingly named because it does more or less what you said.

Calls to grep/stat/readlink are prevented by `hack/verify-code-patterns.sh` in favor of two functions, resolvegrep and resolvelink, that ensure for MacOS users the GNU-equivalent programs are checked for and used if found, otherwise an error is thrown.

corneliusweig:
If I'm not mistaken there is an option to  `brew install coreutils` which creates symlinks without the g for all binaries. So you can call GNU sed simply as `sed` instead of `gsed`.

Maybe this is what we want to recommend? For example
```sh
function ensureGNUUtils () {
  if [[ $(uname) == "Darwin" ]]; then
    if ! hash sed 2>/dev/null; then
      echo "Please install the GNU coreutils with `brew install coreutils --with-secret-option`"
      exit 1
    fi
  fi
}
```

JJJJJones:
> If I'm not mistaken there is an option to `brew install coreutils` which creates symlinks without the g for all binaries. So you can call GNU sed simply as `sed` instead of `gsed`.

I don't see such an option.

```
✗ brew info coreutils
coreutils: stable 8.31 (bottled), HEAD
GNU File, Shell, and Text utilities
https://www.gnu.org/software/coreutils
Conflicts with:
  aardvark_shell_utils (because both install `realpath` binaries)
  b2sum (because both install `b2sum` binaries)
  ganglia (because both install `gstat` binaries)
  gegl (because both install `gcut` binaries)
  idutils (because both install `gid` and `gid.1`)
  md5sha1sum (because both install `md5sum` and `sha1sum` binaries)
  truncate (because both install `truncate` binaries)
/usr/local/Cellar/coreutils/8.31 (476 files, 8.8MB) *
  Poured from bottle on 2019-12-11 at 15:24:35
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/coreutils.rb
==> Options
--HEAD
	Install HEAD version
==> Caveats
Commands also provided by macOS have been installed with the prefix "g".
If you need to use these commands with their normal names, you
can add a "gnubin" directory to your PATH from your bashrc like:
  PATH="/usr/local/opt/coreutils/libexec/gnubin:$PATH"
==> Analytics
install: 39,048 (30 days), 118,368 (90 days), 617,935 (365 days)
install-on-request: 35,515 (30 days), 108,162 (90 days), 560,614 (365 days)
build-error: 0 (30 days)```

ahmetb:
This is the key:

    PATH="/usr/local/opt/coreutils/libexec/gnubin:$PATH"

e.g. https://github.com/ahmetb/dotfiles/blob/78e2c944056223fb55dfd1e5c272b4cb112acdae/.zshrc#L69

JJJJJones:
I think that'd be a mistake.

Installing coreutils from homebrew and referencing "ggrep" or "greadlink" is idempotent. Modifying the PATH would cause unknown side-effects for developers and maintainers who rely on the MacOS variant of commands (such as me) and requires an extra step of setup for people to work on Krew.

ahmetb:
I've been practically developing on macOS for years in Docker and Kubernetes ecosystem and never hit an issue with this PATH configuration.

It's a reasonable expectation from a developer in Kubernetes ecosystem to have an environment with GNU coreutils.

Anecdata aside, people will forget this and add scripts that directly invoke `readlink` –and nobody will remember this patch. Assume tomorrow all maintainers have changed, there's nothing to enforce this coding style. That's why I don't think this is the right approach. Hope that makes sense.

corneliusweig:
Kubernetes repo contains 344 times `sed` and 9 times `gsed`. So it's clear that kubernetes as a project has decided to use GNU coreutils.
As @.ahmetb said, using GNU coreutils has worked fine for him over years on OSX. I'd suggest to try the same and report back if you have trouble.

/close

PR #486, opened 2020-01-30
don't print security notice for krew itself

ahmetb:
The security notice was primarily meant for plugins installed from
krew-index, but this doesn't apply to krew as Krew is a sub-project
of SIG CLI and its source code has proper code reviews and audit trail.

/assign @corneliusweig


corneliusweig:
Good idea!
/lgtm
/approve

PR #488, opened 2020-02-01
Resolve todo and add unit test

corneliusweig:
Related issue: #270

### DIFF
--- a/internal/testutil/tempdir.go
+++ b/internal/testutil/tempdir.go
@@ -21,6 +21,10 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+
+	"sigs.k8s.io/yaml"
+
+	"sigs.k8s.io/krew/pkg/index"
 )
 
 type TempDir struct {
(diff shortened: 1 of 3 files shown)


internal/testutil/tempdir.go
```diff
@@ -72,3 +76,11 @@ func (td *TempDir) Write(file string, content []byte) *TempDir {
 	}
 	return td
 }
+
+func (td *TempDir) WritePlugin(file string, plugin index.Plugin) *TempDir {
```
corneliusweig: I figure this will be useful if we want to write more tests that involve the filesystem

ahmetb: eh, it's ok. i suspect we might forget to use it. :)

ahmetb: pls add `td.t.Helper()` and it's lgtm

corneliusweig: Done.

PR #489, opened 2020-02-01
Check whether PATH has been set up correctly

### LINKED ISSUE #485: Print warning on every command if PATH does not have KREW_ROOT/bin
## Scenario 1: Homebrew

This came up in the Homebrew formula (which we aren't supporting officially at the moment). https://github.com/Homebrew/homebrew-core/pull/49547

Homebrew maintainers refuse to have Caveats for `krew` formula that tells users to:

```
      To complete the installation, you need to add the following to your
      ~/.bash_profile, ~/.bashrc or ~/.zshrc:

          export PATH="${PATH}:${HOME}/.krew/bin"
```

But it does link `kubectl-krew` in PATH. So, users will do `kubectl krew install foo` but `kubectl foo` mysteriously won't work (and `kubectl plugin list` won't show `foo` despite `kubectl krew list` does).

User is left confused.

## Scenario 2: Incomplete krew installations

User followed instructions from Krew's README to install krew. But they did not follow the PATH update (or they did not restart their shell).

So they type `./krew install foo`, it works. But `kubectl foo` mysteriously won't work (and `kubectl plugin list` won't show `foo` despite `kubectl krew list` does).

So the user is left confused.

## Proposed solution

In both problems, user does not have $KREW_ROOT/"bin" directory in their PATH.

So, let's add a PreRun **warning** to root krew command that informs the user about this situation, with the message above.

In addition, we can determine which of `~/.bash_profile`, `~/.bashrc` or `~/.zshrc` or Windows instructions we should be mentioning from environment variables (using $SHELL + GOOS).

**Corner case:** This should not print when `krew` running first time ever, to installing itself (e.g. for the first time) using the instructions in the readme.

/good-first-issue
/priority P2

corneliusweig:
Fixes #485
Related https://github.com/Homebrew/homebrew-core/pull/49547

Homebrew refuses to add a post-install message for setting up
the PATH variable with krew's plugin store. Therefore we need
to check this on our side to warn users about an incomplete
setup. The goal is to
- print specific instructions for each shell/OS.
- not print the instructions when krew installs itself.

### DIFF
--- a/cmd/krew/cmd/internal/setup_check_test.go
+++ b/cmd/krew/cmd/internal/setup_check_test.go
@@ -1,0 +1,104 @@
+// Copyright 2020 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package internal
+
+import (
+	"os"
+	"strings"
+	"testing"
+
+	"sigs.k8s.io/krew/internal/environment"
+	"sigs.k8s.io/krew/internal/testutil"
+)
+
+func TestIsBinDirInPATH_firstRun(t *testing.T) {
+	tempDir, cleanup := testutil.NewTempDir(t)
+	defer cleanup()
+
+	paths := environment.NewPaths(tempDir.Path("does-not-exist"))
+	res := IsBinDirInPATH(paths)
+	if res == false {
+		t.Errorf("Expected positive result on first run")
+	}
+}
+
+func TestIsBinDirInPATH_secondRun(t *testing.T) {
+	tempDir, cleanup := testutil.NewTempDir(t)
+	defer cleanup()
+	if err := os.Mkdir(tempDir.Path("exists"), os.ModePerm); err != nil {
+		t.Fatal(err)
+	}
+	paths := environment.NewPaths(tempDir.Path("exists"))
+	res := IsBinDirInPATH(paths)
+	if res == true {
+		t.Errorf("Expected negative result on second run")
+	}
+}
+
+func TestSetupInstructions_windows(t *testing.T) {
+	const instructionsContain = "add the\n\"%USERPROFILE%\\.krew\\bin\" directory to your PATH environment variable"
+	os.Setenv("KREW_OS", "windows")
+	defer func() { os.Unsetenv("KREW_OS") }()
+	instructions := SetupInstructions()
+	if !strings.Contains(instructions, instructionsContain) {
+		t.Errorf("expected %q\nto contain %q", instructions, instructionsContain)
+	}
+}
+
+func TestSetupInstructions_unix(t *testing.T) {
+	tests := []struct {
+		name                string
+		shell               string
+		instructionsContain string
+	}{
+		{
+			name:                "When the shell is zsh",
+			shell:               "/bin/zsh",
+			instructionsContain: "~/.zshrc",
+		},
+		{
+			name:                "When the shell is bash",
+			shell:               "/bin/bash",
+			instructionsContain: "~/.bash_profile or ~/.bashrc",
+		},
+		{
+			name:                "When the shell is fish",
+			shell:               "/bin/fish",
+			instructionsContain: "config.fish",
+		},
+		{
+			name:                "When the shell is unknown",
+			shell:               "other",
+			instructionsContain: "~/.bash_profile, ~/.bashrc, or ~/.zshrc",
+		},
+	}
+
+	// always set KREW_OS, so that tests succeed on windows
+	os.Setenv("KREW_OS", "linux")
+	defer func(origShell string) {
+		os.Unsetenv("KREW_OS")
+		os.Setenv("SHELL", origShell)
+	}(os.Getenv("SHELL"))
+
+	for _, test := range tests {
+		t.Run(test.name, func(tt *testing.T) {
+			os.Setenv("SHELL", test.shell)
+			instructions := SetupInstructions()
+			if !strings.Contains(instructions, test.instructionsContain) {
+				tt.Errorf("expected %q\nto contain %q", instructions, test.instructionsContain)
+			}
+		})
+	}
+}
--- a/cmd/krew/cmd/internal/setup_check_test.go
+++ b/cmd/krew/cmd/internal/setup_check_test.go
@@ -30,20 +30,17 @@ func TestIsBinDirInPATH_firstRun(t *testing.T) {
 	paths := environment.NewPaths(tempDir.Path("does-not-exist"))
 	res := IsBinDirInPATH(paths)
 	if res == false {
-		t.Errorf("Expected positive result on first run")
+		t.Errorf("expected positive result on first run")
 	}
 }
 
 func TestIsBinDirInPATH_secondRun(t *testing.T) {
 	tempDir, cleanup := testutil.NewTempDir(t)
 	defer cleanup()
-	if err := os.Mkdir(tempDir.Path("exists"), os.ModePerm); err != nil {
-		t.Fatal(err)
-	}
-	paths := environment.NewPaths(tempDir.Path("exists"))
+	paths := environment.NewPaths(tempDir.Root())
 	res := IsBinDirInPATH(paths)
 	if res == true {
-		t.Errorf("Expected negative result on second run")
+		t.Errorf("expected negative result on second run")
 	}
 }
 
--- a/cmd/krew/cmd/internal/setup_check.go
+++ b/cmd/krew/cmd/internal/setup_check.go
@@ -1,0 +1,85 @@
+// Copyright 2020 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package internal
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"k8s.io/klog"
+
+	"sigs.k8s.io/krew/internal/environment"
+	"sigs.k8s.io/krew/internal/installation"
+)
+
+const (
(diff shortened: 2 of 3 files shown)


internal/assertion/setup.go
```diff
@@ -0,0 +1,85 @@
...
+	case strings.HasSuffix(shell, "/fish"):
+		instruction = instructionFish
+	default:
+		instruction = instructionGeneric
+	}
+
+	fmt.Fprintf(w, "%s: %s%s\n\n", red("ATTENTION"), unixMessage, instruction)
+}
+
+func isSetupGood(paths environment.Paths) bool {
```
ahmetb: IsBinDirInPATH

internal/assertion/setup.go
```diff
@@ -0,0 +1,85 @@
...
+
+    set -gx PATH $PATH $HOME/.krew/bin`
+	instructionGeneric = `~/.bash_profile, ~/.bashrc, or ~/.zshrc:
+
+    export PATH="${PATH}:${HOME}/.krew/bin"`
+)
+
+func CheckSetup(w io.Writer, paths environment.Paths) {
+	if isSetupGood(paths) {
+		klog.V(4).Info("Krew setup looks good")
```
ahmetb: Too broad, lets be more specific

internal/assertion/setup.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/fatih/color"
+	"k8s.io/klog"
+
+	"sigs.k8s.io/krew/internal/environment"
+	"sigs.k8s.io/krew/internal/installation"
+)
+
+const (
+	instructionWindows = `To complete the installation, you need to add the "%USERPROFILE%\.krew\bin" directory to your PATH environment variable`
```
ahmetb: Will need 80 char wrap for printed output.

internal/assertion/setup.go
```diff
@@ -0,0 +1,85 @@
...
+		return
+	}
+
+	red := color.New(color.FgRed).SprintfFunc()
+	if installation.IsWindows() {
+		fmt.Fprintf(w, "%s: %s\n\n", red("ATTENTION"), instructionWindows)
+		return
+	}
+
+	unixMessage := `To complete the installation, you need to add the following to your `
```
ahmetb: To complete installation → To be able to run kubectl plugins

internal/assertion/setup.go
```diff
@@ -0,0 +1,85 @@
...
+
+func CheckSetup(w io.Writer, paths environment.Paths) {
+	if isSetupGood(paths) {
+		klog.V(4).Info("Krew setup looks good")
+		return
+	}
+
+	red := color.New(color.FgRed).SprintfFunc()
+	if installation.IsWindows() {
+		fmt.Fprintf(w, "%s: %s\n\n", red("ATTENTION"), instructionWindows)
```
ahmetb: Maybe WARNING is better

internal/assertion/setup.go
```diff
@@ -0,0 +1,85 @@
...
+    export PATH="${PATH}:${HOME}/.krew/bin"`
+	instructionFish = `config.fish:
+
+    set -gx PATH $PATH $HOME/.krew/bin`
+	instructionGeneric = `~/.bash_profile, ~/.bashrc, or ~/.zshrc:
+
+    export PATH="${PATH}:${HOME}/.krew/bin"`
+)
+
+func CheckSetup(w io.Writer, paths environment.Paths) {
```
ahmetb: tbh the "assert" in the package name could make one think this works like "assert" keyword, but it doesn't. maybe pkg/internal/ platform?

ahmetb: func WarnIfBinDirIsNotInPATH is better :)

internal/assertion/setup.go
ahmetb: looks like we're duplicating this line below. maybe come up with a `msg string` and print that at the end of the method?

internal/assertion/setup.go
ahmetb: also add `and restart your shell.`

internal/assertion/setup_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"strings"
+	"testing"
+
+	"sigs.k8s.io/krew/internal/environment"
+	"sigs.k8s.io/krew/internal/testutil"
+)
+
+func TestCheckSetup_firstRun(t *testing.T) {
+	paths := environment.NewPaths("/does-not-exist")
+	buf := &bytes.Buffer{}
```
ahmetb: `var b bytes.Buffer` is enough. you can pass `&b` below.

internal/assertion/setup_test.go
```diff
@@ -0,0 +1,99 @@
...
+	// always set KREW_OS, so that tests succeed on windows
+	os.Setenv("KREW_OS", "linux")
+	defer func(origShell string) {
+		os.Unsetenv("KREW_OS")
+		os.Setenv("SHELL", origShell)
+		cleanup()
+	}(os.Getenv("SHELL"))
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
```
ahmetb: please make the inner tests use `tt` per convention

internal/assertion/setup.go
corneliusweig: Yeah, let's stick to one style for our security warning and this one.

internal/assertion/setup.go
corneliusweig: I'm struggling with a good package name. `installation` and `environment` would make most sense to me, but those are taken already. `platform` doesn't quite ring the right bells to me (although it's better than `assertion`).
What do you think about `internal/setup.Check`?

corneliusweig: I went with `setup` for now. Let me know your thoughts.

corneliusweig:
Thanks for your comments! PTAL

internal/assertion/setup.go
ahmetb: Maybe shouldnt be in a separate pkg :)

internal/setup/warning.go
```diff
@@ -0,0 +1,91 @@
...
+func isBinDirInPATH(paths environment.Paths) bool {
+	// For the first run we don't want to show a warning.
+	_, err := os.Stat(paths.BasePath())
+	if err != nil {
+		klog.V(4).Info("Assuming this is the first run")
+		return os.IsNotExist(err)
+	}
+
+	krewPluginPath := paths.BinPath()
+	return strings.Contains(os.Getenv("PATH"), krewPluginPath)
```
ahmetb: Let's use if there's an utility in kubectl plugin PATH scanning logic.

If not, please separate by `os.PathListSeparator`, and do a `==` check; not `Contains`.

internal/setup/warning_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"bytes"
+	"os"
+	"strings"
+	"testing"
+
+	"sigs.k8s.io/krew/internal/environment"
+	"sigs.k8s.io/krew/internal/testutil"
+)
+
+func TestCheckSetup_firstRun(t *testing.T) {
```
ahmetb: rm test func names :)

internal/setup/warning_test.go
```diff
@@ -0,0 +1,99 @@
...
+	if !strings.Contains(buf.String(), expectedOutput) {
+		t.Errorf("expected %q\nto contain %q", buf.String(), expectedOutput)
+	}
+}
+
+func TestCheckSetup_unix(t *testing.T) {
+	tests := []struct {
+		name   string
+		shell  string
+		output string
```
ahmetb: `outputContains` is more accurate.

internal/setup/warning_test.go
```diff
@@ -0,0 +1,99 @@
...
+	}{
+		{
+			name:   "When the shell is zsh",
+			shell:  "/bin/zsh",
+			output: "add\nthe following to your ~/.zshrc:\n\n    export PATH=\"${PATH}:${HOME}/.krew/bin\"",
+		},
+		{
+			name:   "When the shell is bash",
+			shell:  "/bin/bash",
+			output: "add\nthe following to your ~/.bash_profile or ~/.bashrc:\n\n    export PATH=\"${PATH}:${HOME}/.krew/bin\"",
```
ahmetb: since it's string.Contains, can we just say `"~/.bash_profile or ~/.bashrc"` here?
and `"config.fish"` below?

internal/assertion/setup.go
corneliusweig: :smiley: That could simplify the naming. How about the `environment` package?

internal/setup/warning.go
corneliusweig: Doesn't really look like they have a dedicated util for that: https://github.com/kubernetes/kubernetes/blob/5cc945ad0d2fad2bdde7591490fcf0401fa335aa/staging/src/k8s.io/kubectl/pkg/cmd/plugin/plugin.go#L105

But checking for equality sounds right.

internal/setup/warning_test.go
corneliusweig: I guess you meant the forgotten rename?

internal/setup/warning_test.go
corneliusweig: Yeah, makes sense

internal/setup/warning_test.go
ahmetb: I don't know what I was saying above.
I think I meant make sure Test names are the same as func name.

ahmetb:
This PR is getting too large.
We're just supposed to split env and print a message. :) We don't even need this level of excessive testing. Plus now the `environment` package got refactored further.

Can we split that out to a different PR so that I don't lose context?

IMO this functionality can be in `main` package, doesn't warrant its own pkg.

corneliusweig:
Err.. I agree. This has gotten a little out of hand. It should be simpler now. PTAL

cmd/krew/cmd/internal/setup_check_test.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+func TestIsBinDirInPATH_firstRun(t *testing.T) {
+	tempDir, cleanup := testutil.NewTempDir(t)
+	defer cleanup()
+
+	paths := environment.NewPaths(tempDir.Path("does-not-exist"))
+	res := IsBinDirInPATH(paths)
+	if res == false {
+		t.Errorf("Expected positive result on first run")
```
ahmetb: nit: start error messages lowercase

cmd/krew/cmd/internal/setup_check_test.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+}
+
+func TestIsBinDirInPATH_secondRun(t *testing.T) {
+	tempDir, cleanup := testutil.NewTempDir(t)
+	defer cleanup()
+	if err := os.Mkdir(tempDir.Path("exists"), os.ModePerm); err != nil {
+		t.Fatal(err)
+	}
```
ahmetb: can we just pass tempDir as path, instead of creating a dir called "exists"?

ahmetb:
/lgtm
just some nits
/hold
feel free to cancel if not worth fixing.

I think it's gotten to 200 lines mostly because we added shell-specific handling, which is a better UX anyway.

cmd/krew/cmd/internal/setup_check_test.go
corneliusweig: Sure. I wanted to make it more explicit that the directory is there vs. the non-existent case. But I guess this will do as well.

ahmetb:
/lgtm
/approve
thanks much!

PR #491, opened 2020-02-02
Set new GOPATH env var to github actions

alonyb:
This fixes the old way to set GOPATH to PATH.

This way is used by Minikube too.

I tested it and it works perfectly. [logs](https://github.com/alonyb/krew/runs/421428433?check_suite_focus=true)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,13 +24,14 @@ jobs:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
     
+    - name: Add GOPATH
+      run: echo "::add-path::$(go env GOPATH)/bin"
+    
     - name: Install dependencies
       run: go mod download
     
     - name: Install gox
-      run: | 
-        export PATH=${PATH}:"$(go env GOPATH)/bin"
-        hack/install-gox.sh
+      run: hack/install-gox.sh
       
     - name: Verify code quality
       run: go mod tidy && git diff --no-patch --exit-code
@@ -48,9 +49,7 @@ jobs:
       run: go test -short -coverprofile=coverage.txt -covermode=atomic ./...
         
     - name: Make binaries && verify krew installation
-      run: |
-        export PATH=${PATH}:"$(go env GOPATH)/bin"
-        hack/make-all.sh
+      run: hack/make-all.sh
         
     - name: Ensure kubectl installed
       run: hack/ensure-kubectl-installed.sh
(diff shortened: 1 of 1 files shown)


ahmetb:
Could $() work instead of backticks?

.github/workflows/ci.yml
```diff
@@ -24,19 +24,13 @@ jobs:
...
-        echo "##[set-env name=GOPATH;]$(dirname $GITHUB_WORKSPACE)"
-        echo "##[add-path]$(dirname $GITHUB_WORKSPACE)/bin"
-    
     - name: Install dependencies
       run: go mod download
     
     - name: Install gox
-      run: hack/install-gox.sh
+      run: | 
+        export PATH=${PATH}:`go env GOPATH`/bin
```
corneliusweig:
```suggestion
        export PATH=${PATH}:"$(go env GOPATH)/bin"
```

Please also quote. Just for good measure :)

alonyb:
> Could $() work instead of backticks?

Sure, looks better :) it's done

.github/workflows/ci.yml
alonyb: > Please also quote. Just for good measure :)

Thanks, @corneliusweig I will keep it in mind :)

alonyb:
> Shouldn't we add the `export ...` to _all_ commands involving some go tools? For example, `go mod download` will currently use the system go installation, correct?

It is not necessary, that `PATH` is only used in 2 steps and I would not like to add the same line in all the steps, it does not look good :)

ahmetb:
Is there a way to augment PATH for all steps? Maybe go path is always at $HOME/go/bin just like the default.

alonyb:
> Is there a way to augment PATH for all steps? Maybe go path is always at $HOME/go/bin just like the default.

@ahmetb 
I just invented this.
how about?

ahmetb:
@alonyb that looks perfect tbh and makes the steps easier.

.github/workflows/ci.yml
```diff
@@ -24,13 +24,8 @@ jobs:
...
       uses: actions/checkout@v2
     
-    - name: Add GOPATH/bin to PATH
-      # temporary fix
-      # see https://github.com/actions/setup-go/issues/14
-      run: |
-        set -euo pipefail
-        echo "##[set-env name=GOPATH;]$(dirname $GITHUB_WORKSPACE)"
-        echo "##[add-path]$(dirname $GITHUB_WORKSPACE)/bin"
+    - name: Add GOPATH
```
ferhatelmas: This name seems wrong to me. 

`GOPATH` is already set but we're adding its bin to `PATH`.

ahmetb: Yeah the name could be better like Add GOPATH/bin to PATH.

ferhatelmas: sent #493, let me know what you think.

PR #496, opened 2020-02-06
fix version stamping in build script

ahmetb:
Fixes #...
Related issue: #...

### DIFF
--- a/hack/make-binaries.sh
+++ b/hack/make-binaries.sh
@@ -26,7 +26,7 @@ if ! command -v "gox" &>/dev/null; then
 fi
 
 supported_platforms="darwin/amd64 windows/amd64 linux/amd64 linux/arm"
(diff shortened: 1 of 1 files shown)


ahmetb:
cc: @tony-yang it seems like we're doing cmp.Diff here which is map equality, though it would probably be better to just make sure if the specified set of keys/values exist in the output (instead of exact match)

ahmetb:
/hold
until #497; then rebase.

corneliusweig:
Thanks for finding this bug.
/lgtm
/approve

tony-yang:
> cc: @tony-yang it seems like we're doing cmp.Diff here which is map equality, though it would probably be better to just make sure if the specified set of keys/values exist in the output (instead of exact match)

Fair point. Thanks Ahmet.

PR #497, opened 2020-02-06
make TestKrewVersion a substring check

ahmetb:
Instead of full map comparison, relax it to be "map has all these keys with
the values that contain these strings".

This way, we can test krew version stamping (as it's not deterministic at
testing-time) as well as add more 'krew version' values without needing to
strictly match test.

Related issue: #496


corneliusweig:
Yeah, I think the substring checking is good enough here. Thanks!
/lgtm
/approve

PR #503, opened 2020-02-12
Swap order of install and store receipt

### LINKED ISSUE #502: Upgrade failed; but receipt still written when symlinking fails
/kind bug
/priority P3

Repro steps:

1. Install krew v0.3.3 normally
2. Build a Krew binary locally.
3. Replace ~/.krew/bin symlink of krew itself with the binary :D 

     `mv out/bin/krew-darwin_amd64 $(which kubectl-krew)`

4. `kubectl krew update`  && `kubectl krew upgrade` to krew v0.3.4

5. Observe output:

```
~ $ kubectl krew upgrade

Upgrading plugin: krew
WARNING: failed to upgrade plugin "krew", skipping (error: failed to install new version: failed to link installed plugin: failed to remove old symlink: file "/Users/ahmetb/.krew/bin/kubectl-krew" is not a symlink (mode=-rwxr-xr-x))
Upgrading plugin: ns
Skipping plugin ns, it is already on the newest version
WARNING: Some plugins failed to upgrade, check logs above.
A newer version of krew is available (v0.3.3-47-g14f129f -> v0.3.4).
Run "kubectl krew upgrade" to get the newest version!
```

6. `kubectl krew upgrade krew`  says **"can't upgrade, the newest version is already installed"**

7. But `kubectl krew version` still shows `v0.3.3-47-g14f129f`  and `krew list` shows v0.3.4

Clearly the upgrade receipt was written despite upgrade operation actually failing.

I'm not saying we should support this case (e.g. user breaking the symlink) but if an upgrade fails, install receipt should not be changed.

corneliusweig:
Install is much more likely to fail than storing the receipt.
Before, when the install failed, the new receipts was already
written to disk and not rolled back. Now, the receipt is only
stored when install has succeeded.

Fixes #502

### DIFF
--- a/internal/installation/upgrade.go
+++ b/internal/installation/upgrade.go
@@ -65,11 +65,6 @@ func Upgrade(p environment.Paths, plugin index.Plugin) error {
 	}
 	klog.V(1).Infof("Plugin needs upgrade (%s < %s)", curv, newv)
 
-	klog.V(2).Infof("Upgrading install receipt for plugin %s", plugin.Name)
-	if err = receipt.Store(plugin, p.PluginInstallReceiptPath(plugin.Name)); err != nil {
-		return errors.Wrap(err, "installation receipt could not be stored, uninstall may fail")
-	}
-
 	// Re-Install
 	klog.V(1).Infof("Installing new version %s", newVersion)
 	if err := install(installOperation{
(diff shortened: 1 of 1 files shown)


ahmetb:
Nice.
/lgtm

PR #504, opened 2020-02-12
Use tagged temp dirs to simplify cleanup

corneliusweig:
Fixes #...
Related issue: #...

### DIFF
--- a/docs/RELEASING_KREW.md
+++ b/docs/RELEASING_KREW.md
@@ -13,11 +13,11 @@
     ```sh
     krew=out/bin/krew-darwin_amd64 # assuming macOS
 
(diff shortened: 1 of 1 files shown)


ahmetb:
/tmp shouldn't really need cleanup? :D Isn't that OS responsibility?

corneliusweig:
It is, but usually (at least on my system) it's only done on reboot. And I don't boot frequently O_o

ahmetb:
why don't you just add `&& rm -rf $KREW_ROOT` to the script? that's much easier?

corneliusweig:
I think `KREW_ROOT` is not available anymore in the command after `&&`?

ahmetb:
eh, you're right.

Issue #483, opened 2020-01-30
Proposal nr.2: Support for custom index repositories


chriskim06:
Allowing users to specify custom indices to download plugins from would allow people to distribute plugins without having to go through the main krew-index repo. This would be useful at companies where people want to share plugins that are specific to their workflows or aren't appropriate for krew-index.

I wanted to share the proposal doc I had thrown together here https://docs.google.com/document/d/1Im5gmBA6yuKRWgJWbaBjTTFNvkH0me05V3wEeNPU7fE/edit?usp=sharing and opened a new issue for it based on [this](https://github.com/kubernetes-sigs/krew/pull/478#issuecomment-579107437). There's a lot of similarities to the docs in https://github.com/kubernetes-sigs/krew/issues/23 so if this doesn't add enough I'd be happy to close this.

A lot of ideas came from https://docs.brew.sh/Taps, mainly around how to handle duplicate plugin names.

ahmetb:
Thanks for writing this up.
I'm calling this "proposal number two" for index repos, as you've said there are similarities.
I think both Cornelius and I should read both how Taps work and how your design works. It might take some time but hang in there.

corneliusweig:
@chriskim06 can you ping me on kubernetes slack?

ahmetb:
Ahead of our online meeting about this on Feb 18th, I've documented my wishlist of what I'd love to see from the solution + implementation process:

https://docs.google.com/document/d/1_xYsxtwtQTboKOyp920pb1OWLTQEdU6gF7-QvUA6S0Y/edit?usp=sharing

chriskim06:
Sounds good I'll take a look through that doc and add comments if I have any questions

ahmetb:
Both @corneliusweig and @ahmetb (myself) have discussed the proposal with @chriskim06 and approved it. 

/lgtm
/approve

jdolitsky:
Late to the game here, but would like to point to #580 regarding alternative approach to distribution via OCI

jdolitsky:
The fact that this issue has come up is actually itself testament to "why use OCI" - all these concepts (multitenancy, distributed repos, auth) have already been solved for in the context of Docker push/pull which is the basis for OCI distribution-spec: https://github.com/opencontainers/distribution-spec

jdolitsky:
[...]

PR #505, opened 2020-02-21
Index migration

chriskim06:
This only adds the migration code to change the directory structure that will be needed to support custom indexes. The other changes will be in subsequent PRs.

I'm using the existence of the ~~`index/default/`~~`index/.git` directory to indicate whether or not a migration needs to be performed.

Related issue: #483

### DIFF
--- a/cmd/krew/cmd/system.go
+++ b/cmd/krew/cmd/system.go
@@ -15,8 +15,11 @@
 package cmd
 
 import (
+	"os"
+
 	"github.com/spf13/cobra"
 
+	"sigs.k8s.io/krew/internal/indexmigration"
 	"sigs.k8s.io/krew/internal/receiptsmigration"
 )
 
@@ -50,7 +53,24 @@ This command will be removed without further notice from future versions of krew
 	PreRunE: ensureIndexUpdated,
 }
 
+var indexUpgradeCmd = &cobra.Command{
+	Use:   "index-upgrade",
+	Short: "Perform a migration of the krew index",
+	Long: `Krew became more awesome! To use the new features, you need to run this
+one-time migration, which will enable installing plugins from custom indexes.
+
+This command will be removed without further notice from future versions of krew.
+`,
+	Args: cobra.NoArgs,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return indexmigration.Migrate(paths)
+	},
+}
+
 func init() {
+	if _, ok := os.LookupEnv("X_KREW_ENABLE_MULTI_INDEX"); ok {
+		systemCmd.AddCommand(indexUpgradeCmd)
+	}
 	systemCmd.AddCommand(receiptsUpgradeCmd)
 	rootCmd.AddCommand(systemCmd)
 }
--- a/internal/environment/environment.go
+++ b/internal/environment/environment.go
@@ -59,9 +59,6 @@ func (p Paths) BasePath() string { return p.base }
 // e.g. {BasePath}/index/
 func (p Paths) IndexPath() string { return filepath.Join(p.base, "index") }
 
-// DefaultIndexPath returns the base directory where the default plugin index repository is cloned.
-func (p Paths) DefaultIndexPath() string { return filepath.Join(p.base, "index", "default") }
-
 // IndexPluginsPath returns the plugins directory of the index repository.
 //
 // e.g. {BasePath}/index/plugins/
--- a/internal/environment/environment.go
+++ b/internal/environment/environment.go
@@ -59,6 +59,9 @@ func (p Paths) BasePath() string { return p.base }
 // e.g. {BasePath}/index/
 func (p Paths) IndexPath() string { return filepath.Join(p.base, "index") }
 
+// DefaultIndexPath returns the base directory where the default plugin index repository is cloned.
+func (p Paths) DefaultIndexPath() string { return filepath.Join(p.base, "index", "default") }
+
 // IndexPluginsPath returns the plugins directory of the index repository.
 //
 // e.g. {BasePath}/index/plugins/
--- a/internal/indexmigration/migration.go
+++ b/internal/indexmigration/migration.go
@@ -57,6 +57,6 @@ func Migrate(paths environment.Paths) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to clone index")
 	}
-	klog.Infof("Migration completed succesfully.")
+	klog.Infof("Migration completed successfully.")
 	return nil
 }
--- a/internal/indexmigration/migration.go
+++ b/internal/indexmigration/migration.go
@@ -16,7 +16,9 @@ package indexmigration
 
 import (
 	"os"
+	"path/filepath"
 
+	"github.com/pkg/errors"
 	"k8s.io/klog"
 
 	"sigs.k8s.io/krew/internal/environment"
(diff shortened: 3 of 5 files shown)


internal/indexmigration/migration.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"sigs.k8s.io/krew/internal/environment"
+	"sigs.k8s.io/krew/internal/gitutil"
+	"sigs.k8s.io/krew/pkg/constants"
+)
+
+// Done checks if the krew installation requires a migration to support multiple indexes.
+// A migration is necessary when the index directory doesn't contain a "default" directory.
+func Done(paths environment.Paths) (bool, error) {
+	f, err := os.Stat(paths.DefaultIndexPath())
```
ahmetb: I think this is not a good check.
$KREW_ROOT/index/default not existing might also mean someone just installed krew but there is no default index.

I propose **$KREW_ROOT/index/.git** exists (doesn't even have to be a dir) is a quite good check.

internal/indexmigration/migration_test.go
```diff
@@ -0,0 +1,64 @@
...
+			expected: false,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			tmpDir, cleanup := testutil.NewTempDir(t)
+			defer cleanup()
+
+			_ = os.MkdirAll(tmpDir.Path(test.dirPath), os.ModePerm)
```
ahmetb: handle err :)

internal/indexmigration/migration.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// Migrate removes the index directory and then clones krew-index to the new default index path.
+func Migrate(paths environment.Paths) error {
+	isMigrated, err := Done(paths)
+	if err != nil {
+		return err
+	}
+	if isMigrated {
+		klog.Infoln("Already migrated")
```
ahmetb: please use leveled logs (e.g. klog.v(2) ) for stuff users won't see.

also "already migrated" is not a good log line by itself.

internal/indexmigration/migration.go
```diff
@@ -0,0 +1,59 @@
...
+
+	err = os.RemoveAll(paths.IndexPath())
+	if err != nil {
+		return err
+	}
+	err = gitutil.EnsureCloned(constants.IndexURI, paths.DefaultIndexPath())
+	if err != nil {
+		return err
+	}
+	return nil
```
ahmetb: this simply can be returned as "errors.Wrap"

internal/indexmigration/migration.go
```diff
@@ -0,0 +1,59 @@
...
+		return false, nil
+	}
+	return false, err
+}
+
+// Migrate removes the index directory and then clones krew-index to the new default index path.
+func Migrate(paths environment.Paths) error {
+	isMigrated, err := Done(paths)
+	if err != nil {
+		return err
```
ahmetb: we rarely do `return err` in this codebase.
try `errors.Wrapf(err, "failed to check if index migration is complete")`

internal/environment/environment.go
```diff
@@ -59,6 +59,9 @@ func (p Paths) BasePath() string { return p.base }
 // e.g. {BasePath}/index/
 func (p Paths) IndexPath() string { return filepath.Join(p.base, "index") }
 
+// DefaultIndexPath returns the base directory where the default plugin index repository is cloned.
+func (p Paths) DefaultIndexPath() string { return filepath.Join(p.base, "index", "default") }
```
ahmetb: I feel like we don't need a method like this. Something like we eventually need to refactor this to `IndexPath(name string)`.

The method above (`IndexPath()`) ideally should not be necessary (except for some tests, e.g. below).

internal/indexmigration/migration.go
chriskim06: That's a good idea. I'll make the change.

internal/indexmigration/migration.go
chriskim06: Sounds good. I've seen that throughout the code base, is there some guide for what levels should be used in certain scenarios? I briefly looked through the klog repo but didn't see anything immediately.

internal/environment/environment.go
chriskim06: Ya that makes sense, I'll remove it since this will check for the `.git` directory.

internal/indexmigration/migration.go
ahmetb: Non-leveled logs: always visible to users
Leveled logs: never visible to users, unless `-v LEVEL` is specified.

We frequently use `-v 2` or `-v 10` etc for debugging.
So use judgement from the rest of the repo. :) If it's too off, we'll comment.

internal/environment/environment.go
ahmetb: SGTM.

internal/indexmigration/migration.go
chriskim06: Actually wouldn't we want to display to the user that the migration has already been performed?

ahmetb: 🤔 you’re right but this actually sounds like an error.
But both are fine since this is just a one off command.

chriskim06: Ok I've already changed it to use leveled logging, so I'll just leave it that way if you think either way is cool.

cmd/krew/cmd/system.go
```diff
@@ -50,7 +53,24 @@ This command will be removed without further notice from future versions of krew
...
+This command will be removed without further notice from future versions of krew.
+`,
+	Args: cobra.NoArgs,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return indexmigration.Migrate(paths)
+	},
+}
+
 func init() {
+	if _, ok := os.LookupEnv("X_KREW_ENABLE_MULTI_INDEX"); ok {
```
ahmetb: Maybe add the migration check behind this feature gate to the PreRunE of root as well?

chriskim06: Sure I can add that, I didn't include the check in root but it makes sense to include it in this PR

internal/indexmigration/migration.go
```diff
@@ -0,0 +1,61 @@
...
+
+	err = os.RemoveAll(paths.IndexPath())
+	if err != nil {
+		return errors.Wrapf(err, "could not remove index directory %q", paths.IndexPath())
+	}
+	err = gitutil.EnsureCloned(constants.IndexURI, filepath.Join(paths.IndexPath(), "default"))
+	if err != nil {
+		return errors.Wrap(err, "failed to clone index")
+	}
+	return nil
```
ahmetb: can we quickly have a `klog.Infof("Migration completed succesfully.")` here?

ahmetb:
/lgtm
just one nit.

ahmetb:
/lgtm
/approve
Thanks.

internal/indexmigration/migration.go
```diff
@@ -0,0 +1,62 @@
...
+	if err == nil {
+		return false, nil
+	} else if os.IsNotExist(err) {
+		return true, nil
+	}
+	return false, err
+}
+
+// Migrate removes the index directory and then clones krew-index to the new default index path.
+func Migrate(paths environment.Paths) error {
```
corneliusweig: @chriskim06 This is currently untested. I think this is mostly because the migration code has external dependencies. What about renaming the directory `index -> index/default`? Then you could also test it :)

chriskim06: I'm not sure I totally understand what you mean about renaming the directory. If you could elaborate a little more I'd love to add a unit test for this. I was thinking that this would be covered more in integration tests after looking at the receiptmigration test.

Since this has the call to `gitutil.EnsureCloned` I didn't want that to be called during the unit test. Normally I would do something like using an interface for that function and a different implementation in my test that doesn't actually use git to mock that behavior, but I thought that might overcomplicate this PR.

corneliusweig: So what I was trying to say is that the function is hard to test, because it has an external dependency due to the hidden git clone. After the current function exits, the result is the same as if  `$ mv index index/default` was called (this is just pseudo-mv, it doesn't work like that :) ).

So if your implementation would rely on `mv` instead of `git-clone`, we could write a unit test for it, because file system operations are fine in unit tests. I think I'll give it a try to see if it works.

chriskim06: Whoops, I meant to reply to this yesterday. But ya, that makes sense, I didn't think to do it that way. It does seem more testable in your PR.

PR #506, opened 2020-02-21
add helper for inferring index name from input

ahmetb:
To be used in multi-index work.
cc: @chriskim06

### DIFF
--- a/internal/pathutil/pathutil_test.go
+++ b/internal/pathutil/pathutil_test.go
@@ -184,3 +184,27 @@ func TestReplaceBase(t *testing.T) {
 		})
 	}
 }
+
+func TestCanonicalPluginName(t *testing.T) {
+	tests := []struct {
+		in        string
+		wantIndex string
+		wantName  string
+	}{
+		{"foo", "default", "foo"},
+		{"", "default", ""}, // despite unsupported
+		{"a/b", "a", "b"},
+		{"a/b/c", "a", "b/c"}, // despite unsupported
+	}
+	for _, tt := range tests {
+		t.Run(tt.in, func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


chriskim06:
I can use `constants.DefaultIndexName` in https://github.com/kubernetes-sigs/krew/pull/505, so when this gets merge I'll rebase the other one.

ahmetb:
Migration codes ideally should not use external variables like this (we want them to do a rather static job). 
So let’s hardcode that one.

internal/pathutil/pathutil_test.go
```diff
@@ -184,3 +184,27 @@ func TestReplaceBase(t *testing.T) {
...
+		wantIndex string
+		wantName  string
+	}{
+		{"foo", "default", "foo"},
+		{"", "default", ""}, // despite unsupported
+		{"a/b", "a", "b"},
+		{"a/b/c", "a", "b/c"}, // despite unsupported
+	}
+	for _, tt := range tests {
+		t.Run(tt.in, func(t *testing.T) {
```
corneliusweig: We are not entirely consistent about how test cases and the inner `*testing.T` is called. In some places, the inner is  `tt *testing.T` in others just `t`.

I personally would prefer to have consistency here an only use `tt` for the inner `*testing.T`. However, we can do that on the whole repo in a follow-up PR.

corneliusweig: Opened  #508 to clarify this.

ahmetb: This is mostly because we've used different IDEs over time (vscode-go and Goland).

Similarly you might notice I added a `=` after the `...%q, want =` here. Most other unit-tests don't have the `=`. :)

PR #510, opened 2020-02-22
Refactor existing index path functions to support multiple indexes later on

chriskim06:
Add a name parameter to the index path functions (and behind the `X_KREW_ENABLE_MULTI_INDEX` variable) to return the path for the specified index, with "default" being implicit. Refactor the existing calls to use an empty string for now that gets ignored when the variable is unset.

The integration tests fail when `X_KREW_ENABLE_MULTI_INDEX` is set since the other functionality doesn't exist yet.

Related issue: #483

### DIFF
--- a/cmd/krew/cmd/install.go
+++ b/cmd/krew/cmd/install.go
@@ -91,7 +91,7 @@ Remarks:
 
 			var install []index.Plugin
 			for _, name := range pluginNames {
-				plugin, err := indexscanner.LoadPluginByName(paths.IndexPluginsPath(""), name)
+				plugin, err := indexscanner.LoadPluginByName(paths.IndexPluginsPath("default"), name)
 				if err != nil {
 					if os.IsNotExist(err) {
 						return errors.Errorf("plugin %q does not exist in the plugin index", name)
--- a/cmd/krew/cmd/install.go
+++ b/cmd/krew/cmd/install.go
@@ -91,7 +91,7 @@ Remarks:
 
 			var install []index.Plugin
 			for _, name := range pluginNames {
-				plugin, err := indexscanner.LoadPluginByName(paths.IndexPluginsPath(), name)
+				plugin, err := indexscanner.LoadPluginByName(paths.IndexPluginsPath(""), name)
 				if err != nil {
 					if os.IsNotExist(err) {
 						return errors.Errorf("plugin %q does not exist in the plugin index", name)
--- a/cmd/krew/cmd/install.go
+++ b/cmd/krew/cmd/install.go
@@ -27,6 +27,7 @@ import (
 	"sigs.k8s.io/krew/cmd/krew/cmd/internal"
 	"sigs.k8s.io/krew/internal/index/indexscanner"
 	"sigs.k8s.io/krew/internal/installation"
+	"sigs.k8s.io/krew/pkg/constants"
 	"sigs.k8s.io/krew/pkg/index"
 )
 
@@ -91,7 +92,7 @@ Remarks:
 
 			var install []index.Plugin
 			for _, name := range pluginNames {
-				plugin, err := indexscanner.LoadPluginByName(paths.IndexPluginsPath("default"), name)
+				plugin, err := indexscanner.LoadPluginByName(paths.IndexPluginsPath(constants.DefaultIndexName), name)
 				if err != nil {
 					if os.IsNotExist(err) {
 						return errors.Errorf("plugin %q does not exist in the plugin index", name)
--- a/internal/environment/environment.go
+++ b/internal/environment/environment.go
@@ -57,12 +57,22 @@ func (p Paths) BasePath() string { return p.base }
 // IndexPath returns the base directory where plugin index repository is cloned.
 //
 // e.g. {BasePath}/index/
-func (p Paths) IndexPath() string { return filepath.Join(p.base, "index") }
+func (p Paths) IndexPath(name string) string {
+	if _, ok := os.LookupEnv("X_KREW_ENABLE_MULTI_INDEX"); ok {
+		if name == "" {
+			return filepath.Join(p.base, "index", constants.DefaultIndexName)
+		}
+		return filepath.Join(p.base, "index", name)
+	}
+	return filepath.Join(p.base, "index")
+}
 
 // IndexPluginsPath returns the plugins directory of the index repository.
 //
 // e.g. {BasePath}/index/plugins/
-func (p Paths) IndexPluginsPath() string { return filepath.Join(p.base, "index", "plugins") }
+func (p Paths) IndexPluginsPath(name string) string {
+	return filepath.Join(p.IndexPath(name), "plugins")
+}
 
 // InstallReceiptsPath returns the base directory where plugin receipts are stored.
 //
--- a/internal/environment/environment.go
+++ b/internal/environment/environment.go
@@ -65,9 +65,6 @@ func (p Paths) IndexBase() string {
 // e.g. {BasePath}/index/
 func (p Paths) IndexPath(name string) string {
 	if _, ok := os.LookupEnv(constants.EnableMultiIndexSwitch); ok {
-		if name == "" {
-			return filepath.Join(p.base, "index", constants.DefaultIndexName)
-		}
 		return filepath.Join(p.base, "index", name)
 	}
 	return p.IndexBase()
--- a/internal/environment/environment.go
+++ b/internal/environment/environment.go
@@ -54,7 +54,8 @@ func NewPaths(base string) Paths {
 // BasePath returns krew base directory.
 func (p Paths) BasePath() string { return p.base }
 
-// IndexBase returns the krew index directory.
+// IndexBase returns the krew index directory. This directory contains the default
+// index and custom ones.
 func (p Paths) IndexBase() string {
 	return filepath.Join(p.base, "index")
 }
@@ -63,7 +64,7 @@ func (p Paths) IndexBase() string {
 //
 // e.g. {BasePath}/index/
 func (p Paths) IndexPath(name string) string {
-	if _, ok := os.LookupEnv(constants.EnableMultiIndexFlag); ok {
+	if _, ok := os.LookupEnv(constants.EnableMultiIndexSwitch); ok {
 		if name == "" {
 			return filepath.Join(p.base, "index", constants.DefaultIndexName)
 		}
--- a/internal/environment/environment.go
+++ b/internal/environment/environment.go
@@ -60,9 +60,10 @@ func (p Paths) IndexBase() string {
 	return filepath.Join(p.base, "index")
 }
 
-// IndexPath returns the base directory where plugin index repository is cloned.
-//
-// e.g. {BasePath}/index/
+// IndexPath returns the directory where a plugin index repository is cloned.
+// When constants.EnableMultiIndexSwitch var is unset it just returns the krew
+// index base.
+// e.g. {BasePath}/index/default or {BasePath}/index
 func (p Paths) IndexPath(name string) string {
 	if _, ok := os.LookupEnv(constants.EnableMultiIndexSwitch); ok {
 		return filepath.Join(p.base, "index", name)
@@ -70,9 +71,10 @@ func (p Paths) IndexPath(name string) string {
 	return p.IndexBase()
 }
 
-// IndexPluginsPath returns the plugins directory of the index repository.
-//
-// e.g. {BasePath}/index/plugins/
+// IndexPluginsPath returns the plugins directory of an index repository.
+// When constants.EnableMultiIndexSwitch var is unset it just returns the old
+// structure krew-index plugins.
+// e.g. {BasePath}/index/default/plugins/ or {BasePath}/index/plugins/
 func (p Paths) IndexPluginsPath(name string) string {
 	return filepath.Join(p.IndexPath(name), "plugins")
 }
--- a/internal/environment/environment.go
+++ b/internal/environment/environment.go
@@ -54,17 +54,22 @@ func NewPaths(base string) Paths {
 // BasePath returns krew base directory.
 func (p Paths) BasePath() string { return p.base }
 
+// IndexBase returns the krew index directory.
+func (p Paths) IndexBase() string {
+	return filepath.Join(p.base, "index")
+}
+
 // IndexPath returns the base directory where plugin index repository is cloned.
 //
 // e.g. {BasePath}/index/
 func (p Paths) IndexPath(name string) string {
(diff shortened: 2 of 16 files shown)


internal/environment/environment_test.go
```diff
@@ -54,11 +54,26 @@ func TestPaths(t *testing.T) {
...
+			t.Fatalf("IndexPluginsPath(\"\")=%s; expected=%s", got, expected)
+		}
+		if got, expected := p.IndexPath("test"), filepath.FromSlash("/foo/index/test"); got != expected {
+			t.Fatalf("IndexPath()=%s; expected=%s", got, expected)
+		}
+		if got, expected := p.IndexPluginsPath("test"), filepath.FromSlash("/foo/index/test/plugins"); got != expected {
+			t.Fatalf("IndexPluginsPath(\"\")=%s; expected=%s", got, expected)
+		}
+	} else {
+		if got, expected := p.IndexPath(""), filepath.FromSlash("/foo/index"); got != expected {
```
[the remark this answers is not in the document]
chriskim06: That's a good point, I didn't think of that. I'll add it in a bit.

internal/environment/environment_test.go
```diff
@@ -54,11 +54,26 @@ func TestPaths(t *testing.T) {
 	if got, expected := p.BinPath(), filepath.FromSlash("/foo/bin"); got != expected {
 		t.Fatalf("BinPath()=%s; expected=%s", got, expected)
 	}
-	if got, expected := p.IndexPath(), filepath.FromSlash("/foo/index"); got != expected {
-		t.Fatalf("IndexPath()=%s; expected=%s", got, expected)
-	}
-	if got, expected := p.IndexPluginsPath(), filepath.FromSlash("/foo/index/plugins"); got != expected {
-		t.Fatalf("IndexPluginsPath()=%s; expected=%s", got, expected)
+	if _, ok := os.LookupEnv("X_KREW_ENABLE_MULTI_INDEX"); ok {
```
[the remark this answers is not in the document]
chriskim06: Good call! I'll add it in

internal/indexmigration/migration_test.go
```diff
@@ -20,11 +20,12 @@ import (
 
 	"sigs.k8s.io/krew/internal/environment"
 	"sigs.k8s.io/krew/internal/testutil"
+	"sigs.k8s.io/krew/pkg/constants"
 )
 
 func TestIsMigrated(t *testing.T) {
-	if _, ok := os.LookupEnv("X_KREW_ENABLE_MULTI_INDEX"); !ok {
-		t.Skip("Set X_KREW_ENABLE_MULTI_INDEX variable to run this test")
+	if _, ok := os.LookupEnv(constants.EnableMultiIndexFlag); !ok {
```
corneliusweig: There is actually no reason to skip this test. Also see my PR #511.

pkg/constants/constants.go
```diff
@@ -24,4 +24,7 @@ const (
 	IndexURI = "https://github.com/kubernetes-sigs/krew-index.git"
 	// DefaultIndexName is a magic string that's used for a plugin name specified without an index.
 	DefaultIndexName = "default"
+	// EnableMultiIndexFlag is the name of the environment variable that needs to be set to use
+	// the features around multiple indexes (this will be removed later on).
+	EnableMultiIndexFlag = "X_KREW_ENABLE_MULTI_INDEX"
```
corneliusweig: Is this really a flag? Or is `EnableMultiIndexSwitch` better?
I find  it hard to  come up with a perfect name though. `EnableMultiIndexEnvVarName` is a bit long.

internal/environment/environment.go
```diff
@@ -54,17 +54,22 @@ func NewPaths(base string) Paths {
 // BasePath returns krew base directory.
 func (p Paths) BasePath() string { return p.base }
 
+// IndexBase returns the krew index directory.
+func (p Paths) IndexBase() string {
```
corneliusweig: I think this is a good addition. Most callers of `IndexPath` will in the future call this function instead. I think it would make sense to switch all the other calling sites which look like `IndexPath("")` to this function.

Could you also mention in the godoc that the base includes all the indexes (default + custom ones).

internal/environment/environment_test.go
```diff
@@ -54,27 +55,36 @@ func TestPaths(t *testing.T) {
...
-			t.Fatalf("IndexPath()=%s; expected=%s", got, expected)
-		}
-		if got, expected := p.IndexPluginsPath(""), filepath.FromSlash("/foo/index/plugins"); got != expected {
-			t.Fatalf("IndexPluginsPath(\"\")=%s; expected=%s", got, expected)
-		}
+
+	// Test index path functions with EnableMultiIndexFlag
+	os.Setenv(constants.EnableMultiIndexFlag, "1")
+	if got, expected := p.IndexPath(""), filepath.FromSlash("/foo/index/default"); got != expected {
+		t.Fatalf("IndexPath()=%s; expected=%s", got, expected)
```
corneliusweig: None of these errors is fatal (this is not your fault ;) ). Can you switch to `Errorf` please?
```suggestion
		t.Errorf("IndexPath()=%s; expected=%s", got, expected)
```

internal/environment/environment_test.go
```diff
@@ -54,27 +55,36 @@ func TestPaths(t *testing.T) {
...
-	} else {
-		if got, expected := p.IndexPath(""), filepath.FromSlash("/foo/index"); got != expected {
-			t.Fatalf("IndexPath()=%s; expected=%s", got, expected)
-		}
-		if got, expected := p.IndexPluginsPath(""), filepath.FromSlash("/foo/index/plugins"); got != expected {
-			t.Fatalf("IndexPluginsPath(\"\")=%s; expected=%s", got, expected)
-		}
+
+	// Test index path functions with EnableMultiIndexFlag
+	os.Setenv(constants.EnableMultiIndexFlag, "1")
```
corneliusweig: It's a bit dangerous to _not_ clean up with `defer os.Unsetenv(...)`. However, in this case I think the code is simple enough so that we can be sure that the `Unsetenv`  is always reached. Provided you switch to `Errorf` :)

internal/environment/environment_test.go
```diff
@@ -54,27 +55,36 @@ func TestPaths(t *testing.T) {
...
+	if got, expected := p.IndexPluginsPath(""), filepath.FromSlash("/foo/index/default/plugins"); got != expected {
+		t.Fatalf("IndexPluginsPath(\"\")=%s; expected=%s", got, expected)
+	}
+	if got, expected := p.IndexPath("test"), filepath.FromSlash("/foo/index/test"); got != expected {
+		t.Fatalf("IndexPath()=%s; expected=%s", got, expected)
+	}
+	if got, expected := p.IndexPluginsPath("test"), filepath.FromSlash("/foo/index/test/plugins"); got != expected {
+		t.Fatalf("IndexPluginsPath(\"\")=%s; expected=%s", got, expected)
+	}
+	// Test index path functions without EnableMultiIndexFlag
```
corneliusweig: It may be worthwhile to group those tests via `t.Run("without EnableMultiIndexFlag", func (t *testing.T){ ...}). WDYT?

pkg/constants/constants.go
chriskim06: I couldn't really come up with a great name either. I think `EnableMultiIndexSwitch` is fine, or what about `EnableMultiIndexEnvVar`?

internal/environment/environment.go
chriskim06: Ya I was thinking about that. I think though that most uses after the multi index changes would still use `IndexPath(name)` right?

I'll update the godoc for it.

internal/environment/environment_test.go
chriskim06: Ya that seems much cleaner, I'll make the change :+1:

pkg/constants/constants.go
corneliusweig: Let's go with switch.

internal/environment/environment.go
corneliusweig: It depends. I think all those usages in connection to `LoadPluginListFromFS` will use the index base. And  there are a few of those :)

chriskim06: Wouldn't those still use `IndexPath(name)`? It would need to look in `$KREW_ROOT/index/.../plugins` so I'd think the calls to `LoadPluginListFromFS` would still be passed `IndexPluginsPath(name)`.

Let me know if I've misunderstood something.

ahmetb:
> Refactor the existing calls to use an empty string for now that gets ignored when the variable is unset.

In general, if you use `"default"` string everywhere, you can finish refactoring those pieces later. Whereas `""` is harder to find and replace.

cmd/krew/cmd/install.go
```diff
@@ -91,7 +91,7 @@ Remarks:
 
 			var install []index.Plugin
 			for _, name := range pluginNames {
-				plugin, err := indexscanner.LoadPluginByName(paths.IndexPluginsPath(), name)
+				plugin, err := indexscanner.LoadPluginByName(paths.IndexPluginsPath(""), name)
```
ahmetb: to  reinforce my previous comment, you might as well use `"default"` here. Later on, you'll have to scan every single usage of these methods and replace them all over again. Using `"default"` can be make your job as it's easier to find.

internal/environment/environment.go
```diff
@@ -54,15 +54,31 @@ func NewPaths(base string) Paths {
...
+	return filepath.Join(p.base, "index")
+}
+
 // IndexPath returns the base directory where plugin index repository is cloned.
 //
 // e.g. {BasePath}/index/
-func (p Paths) IndexPath() string { return filepath.Join(p.base, "index") }
+func (p Paths) IndexPath(name string) string {
+	if _, ok := os.LookupEnv(constants.EnableMultiIndexSwitch); ok {
+		if name == "" {
```
ahmetb: if you use `"default"` in the call sites instead of `""` you can simply eliminate this `name=""` check, as well. :)

ahmetb:
LGTM but one major point surfaced up in all the patch.

internal/environment/environment.go
corneliusweig: Yeah, that's right. But we need to discover the indices by listing the index base directory.

Anyway, I don't think it matters too much right now. We have to revisit later anyway.

internal/indexmigration/migration.go
```diff
@@ -49,11 +49,11 @@ func Migrate(paths environment.Paths) error {
...
 	}
 
-	err = os.RemoveAll(paths.IndexPath(""))
+	err = os.RemoveAll(paths.IndexPath("default"))
 	if err != nil {
-		return errors.Wrapf(err, "could not remove index directory %q", paths.IndexPath(""))
+		return errors.Wrapf(err, "could not remove index directory %q", paths.IndexPath("default"))
 	}
-	err = gitutil.EnsureCloned(constants.IndexURI, filepath.Join(paths.IndexPath(""), "default"))
+	err = gitutil.EnsureCloned(constants.IndexURI, filepath.Join(paths.IndexBase(), "default"))
```
corneliusweig: Can you add a todo comment here that `filepath.Join(paths.IndexBase(), "default")` be replaced with `paths.IndexPath("default")` at some point?

Or even better, can we use that right now? After  all, migration should only be called when the env var is set and then `IndexPath("default")` will give the correct result.

internal/environment/environment.go
```diff
@@ -65,9 +65,6 @@ func (p Paths) IndexBase() string {
 // e.g. {BasePath}/index/
 func (p Paths) IndexPath(name string) string {
 	if _, ok := os.LookupEnv(constants.EnableMultiIndexSwitch); ok {
-		if name == "" {
-			return filepath.Join(p.base, "index", constants.DefaultIndexName)
```
corneliusweig: Now we are not using `constants.DefaultIndexName` in all the places with a literal `"default"`. I think it would be better to do so, because it makes it even easier to find all usages.

internal/environment/environment_test.go
```diff
@@ -59,31 +59,31 @@ func TestPaths(t *testing.T) {
...
+			t.Errorf("IndexPluginsPath(\"default\")=%s; expected=%s", got, expected)
 		}
 		if got, expected := p.IndexPath("test"), filepath.FromSlash("/foo/index/test"); got != expected {
-			t.Errorf("IndexPath()=%s; expected=%s", got, expected)
+			t.Errorf("IndexPath(\"test\")=%s; expected=%s", got, expected)
 		}
 		if got, expected := p.IndexPluginsPath("test"), filepath.FromSlash("/foo/index/test/plugins"); got != expected {
-			t.Errorf("IndexPluginsPath(\"\")=%s; expected=%s", got, expected)
+			t.Errorf("IndexPluginsPath(\"test\")=%s; expected=%s", got, expected)
 		}
```
corneliusweig: There's on special treatment of `""` anymore, so the above two test cases do the same as these.

internal/environment/environment_test.go
```diff
@@ -59,31 +59,31 @@ func TestPaths(t *testing.T) {
...
-		if got, expected := p.IndexPluginsPath(""), filepath.FromSlash("/foo/index/plugins"); got != expected {
-			t.Errorf("IndexPluginsPath(\"\")=%s; expected=%s", got, expected)
+		if got, expected := p.IndexPluginsPath("default"), filepath.FromSlash("/foo/index/plugins"); got != expected {
+			t.Errorf("IndexPluginsPath(\"default\")=%s; expected=%s", got, expected)
 		}
 		if got, expected := p.IndexPluginsPath("test"), filepath.FromSlash("/foo/index/plugins"); got != expected {
-			t.Errorf("IndexPluginsPath(\"\")=%s; expected=%s", got, expected)
+			t.Errorf("IndexPluginsPath(\"test\")=%s; expected=%s", got, expected)
 		}
```
corneliusweig: Similar to above, these test cases are now redundant.

internal/environment/environment.go
```diff
@@ -54,15 +54,28 @@ func NewPaths(base string) Paths {
...
 
+// IndexBase returns the krew index directory. This directory contains the default
+// index and custom ones.
+func (p Paths) IndexBase() string {
+	return filepath.Join(p.base, "index")
+}
+
 // IndexPath returns the base directory where plugin index repository is cloned.
 //
 // e.g. {BasePath}/index/
```
corneliusweig: It's worth to bring godoc up-to-date for this function here. Especially the env var switch is worth being mentioned.

internal/environment/environment.go
chriskim06: I totally forgot about that while I was adding this. I'll make that change.

corneliusweig:
Ok, let's get this in. Once the conflict is solved, I'll merge.

PR #511, opened 2020-02-22
Simpler migration without external dependency

corneliusweig:
This migrates the krew home to the new layout without an external dependency.
Related issue: #505

Also see https://github.com/kubernetes-sigs/krew/pull/505#discussion_r382856984

cc @chriskim06

### DIFF
--- a/internal/indexmigration/migration.go
+++ b/internal/indexmigration/migration.go
@@ -22,8 +22,6 @@ import (
 	"k8s.io/klog"
 
 	"sigs.k8s.io/krew/internal/environment"
-	"sigs.k8s.io/krew/internal/gitutil"
-	"sigs.k8s.io/krew/pkg/constants"
 )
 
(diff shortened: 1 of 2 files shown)


internal/indexmigration/migration.go
```diff
@@ -48,15 +44,22 @@ func Migrate(paths environment.Paths) error {
 		klog.V(2).Infoln("Already migrated")
 		return nil
 	}
+	indexPath := paths.IndexPath()
+	tmpPath := filepath.Join(paths.BasePath(), "tmp_index_migration")
```
ahmetb: this wouldn't be better as os.TempDir?

ahmetb:
/lgtm
/approve
/hold 
just a question.

internal/indexmigration/migration.go
corneliusweig: TempDir would be safer of course, but I wanted to avoid moving across file systems.

Issue #507, opened 2020-02-21
Add a new receipt type


chriskim06:
Plugin receipts currently use the `index.Plugin` type. In order to support the changes to support multiple indexes outlined in https://github.com/kubernetes-sigs/krew/issues/483, it was decided that the information for which index a plugin is installed from would be stored in the receipt. In order to support this, a new receipt type should be added. Here's a pretty simple proposed structure of what a receipt type might look like:
```
// Receipt describes a receipt manifest file.
type Receipt struct {
        Plugin   `json:",inline" yaml:",inline"`

        Status ReceiptStatus `json:"status,omitempty"`
}

// ReceiptStatus contains information about the installed plugin.
type ReceiptStatus struct {
        Source SourceIndex `json:"source"`
}

// SourceIndex contains information about the index a plugin was installed from.
type SourceIndex struct {
        // Name is the configured name of an index a plugin was installed from.
        Name string `json:"name"`
}
```

corneliusweig:
We probably also want `omitempty`.

I think it's best to start with a minimal version for the Receipt type. So
```go
type Receipt Plugin
```
should be enough for the initial migration.

---

After the migration is done, we can add fields as required. I think we should only add those fields that are necessary. At the moment we only have a use-case for `SourceIndex`, so what you suggest looks great.
However, we need to think about the semantics of that field. Is it the tap name at the time of install (which might change)? Or maybe the git remote url (which should be more static)? This btw should go into the GoDoc for that field.

chriskim06:
Agreed, that sounds like the best way to introduce the receipt type at first.

I don't really have a strong opinion on what `SourceIndex` should be. I think the index name will probably be easier to work with from a code perspective but I don't know off the top of my head how we would deal with a user adding a different index with the same name and having krew not try to update a plugin installed from the deleted index with the same name. Does that use case fall under the category of edge cases we agreed would have intentionally undefined behavior?

ahmetb:
+1 design looks good. However it’s not extensible. I recommend this:

- status:
  - sourceIndex:
    - name:

If the name-based soln doesn’t work out long-term we can migrate much easier this way. 

We’d probably need a ReceiptFromPlugin method to convert plugin to a receipt in a central place ideally.

chriskim06:
Makes sense, I'll update the design.

PR #512, opened 2020-02-25
Initial receipt type

chriskim06:
This refactors the receipts to use a new `Receipt` type. The tests for it are pretty much identical to the old ones but using the new type. Once the status object gets added to receipts the tests will need to be updated.

I decided to not do `type Receipt Plugin`, only because I figured this would be changed very soon and would save a bit of work down the line. If you'd like me to change it and start out without the embedded type just let me know.

Related issues: #483, #507


internal/index/validation/validate.go
```diff
@@ -99,6 +99,45 @@ func ValidatePlugin(name string, p index.Plugin) error {
 	return nil
 }
 
+// ValidateReceipt checks for structural validity of the Receipt object with given
+// name.
+func ValidateReceipt(name string, r index.Receipt) error {
```
ahmetb: This is pure duplication, we shouldn't do this.

I'm not sure if/why we need to validate `Receipt`s. `Plugin`s are authored by developer community; `Receipt`s aren't. Therefore it should be **undefined/unsupported behavior** trying to catch validation errors (stemmed from `Plugin`) in the `Receipt` type.

internal/testutil/receipt.go
```diff
@@ -0,0 +1,35 @@
...
+func NewReceipt() *Receipt {
+	return &Receipt{v: index.Receipt{Plugin: NewPlugin().V()}}
+}
+
+func (r *Receipt) WithName(s string) *Receipt                 { r.v.ObjectMeta.Name = s; return r }
+func (r *Receipt) WithShortDescription(v string) *Receipt     { r.v.Spec.ShortDescription = v; return r }
+func (r *Receipt) WithTypeMeta(v metav1.TypeMeta) *Receipt    { r.v.TypeMeta = v; return r }
+func (r *Receipt) WithPlatforms(v ...index.Platform) *Receipt { r.v.Spec.Platforms = v; return r }
+func (r *Receipt) WithVersion(v string) *Receipt              { r.v.Spec.Version = v; return r }
+func (r *Receipt) V() index.Receipt                           { return r.v }
```
ahmetb: let's not copy it like this.

Something that is much smaller like the following would be enough:

```
NewReceipt().WithPlugin(p).WithStatus()
```

internal/index/validation/validate.go
chriskim06: Ya I was pretty iffy on this. I only added it because the original receipt loading called `ValidatePlugin`, I can remove it though.

pkg/index/types.go
```diff
@@ -58,3 +58,8 @@ type FileOperation struct {
 	From string `json:"from,omitempty"`
 	To   string `json:"to,omitempty"`
 }
+
+// Receipt describes a plugin receipt file.
```
ahmetb: `plugin receipt` &rarr; `plugin installation receipt`.

Honestly we can call the type `InstallReceipt` as well. @corneliusweig thoughts?

internal/index/validation/validate.go
ahmetb: Yeah we haven't gotten a whole lot out of this.
The reason we have it on `Plugin` is because it's actually an "input" to our program. However `Receipt` is something our program writes/reads itself to store "state".

internal/index/indexscanner/scanner.go
```diff
@@ -98,26 +98,50 @@ func ReadPluginFromFile(path string) (index.Plugin, error) {
...
+// returns an error that can be checked with os.IsNotExist.
+func ReadReceiptFromFile(path string) (index.Receipt, error) {
+	f, err := os.Open(path)
+	if os.IsNotExist(err) {
+		// TODO(ahmetb): we should use go1.13+ errors.Is construct at call sites to evaluate if an error is os.IsNotExist
+		return index.Receipt{}, err
+	} else if err != nil {
+		return index.Receipt{}, errors.Wrap(err, "failed to open index file")
+	}
+	return ReadReceipt(f)
```
ahmetb: This is pure duplication.

Please implement something like `decodeFile()` that works for both Receipt and Plugin to deduplicate implementation. This should be just:

```
func ReadReceiptFromFile(f file) (Receipt, error) {
    var v Receipt
    return decodeFile(f, &v)
}

File: internal/index/indexscanner/scanner.go [hunk 5]
```diff
@@ -98,26 +98,50 @@ func ReadPluginFromFile(path string) (index.Plugin, error) {
...
+
+func ReadReceipt(f io.ReadCloser) (index.Receipt, error) {
+	defer f.Close()
+	receipt := &index.Receipt{}
+	err := decodeFile(f, receipt)
+	r := *receipt
+	if err != nil {
+		return r, errors.Wrap(err, "failed to decode plugin manifest")
+	}
+	return r, errors.Wrap(validation.ValidateReceipt(r.Name, r), "receipt manifest validation error")
```
ahmetb: ditto: de-duplicate implementation.

internal/index/indexscanner/scanner.go
ahmetb: Actually I've just noticed you've done ReadReceiptFromFile &rarr; ReadReceipt &rarr; decodeFile.

However, the core logic seems still very much copypasta from the implementation for Plugin type. So let's try to minimize those if they both work the same way.

internal/testutil/receipt.go
chriskim06: I could just remove instead, since I only added this for the validation test that I was unsure of. What do you think about just removing this for now? It will probably come up later but its not needed once I remove the validation stuff.

internal/index/indexscanner/scanner.go
```diff
@@ -98,26 +98,50 @@ func ReadPluginFromFile(path string) (index.Plugin, error) {
...
+	if os.IsNotExist(err) {
+		// TODO(ahmetb): we should use go1.13+ errors.Is construct at call sites to evaluate if an error is os.IsNotExist
+		return index.Receipt{}, err
+	} else if err != nil {
+		return index.Receipt{}, errors.Wrap(err, "failed to open index file")
+	}
+	return ReadReceipt(f)
+}
+
+func ReadReceipt(f io.ReadCloser) (index.Receipt, error) {
```
ahmetb: If you look at usages of `ReadPlugin` you'll see that there's only 1 place it's used outside this package (`krew install --manifest` argument). That's why this method was necessary and exported.

I don't see `ReadReceipt` to exist or be exported.

internal/testutil/receipt.go
ahmetb: Yeah I don't see enough reasons for this to exist yet.

internal/index/indexscanner/scanner.go
chriskim06: Ya the main problem is that `ReadPluginFromFile`/`ReadReceiptFromFile` return a concrete type `Plugin`/`Receipt`; both of them just read a file and unmarshal to one or the other. I minimized the duplication as much as I could but you're right that it is pretty much a copypasta of the plugin implementation. I'll take another pass at it to see what else I can do, like moving the `ReadReceipt` function into `ReadReceiptFromFile`.

internal/index/indexscanner/scanner.go
```diff
@@ -98,26 +98,45 @@ func ReadPluginFromFile(path string) (index.Plugin, error) {
 
 func ReadPlugin(f io.ReadCloser) (index.Plugin, error) {
 	defer f.Close()
-	p, err := DecodePluginFile(f)
+	plugin := &index.Plugin{}
```
ahmetb: you can simply do this to avoid explicit zero-declaration and `*dereference`s below (ditto for ReadReceipt from file).

```
var p index.Plugin
decode(f,&p)
```

internal/index/indexscanner/scanner.go
```diff
@@ -98,26 +98,45 @@ func ReadPluginFromFile(path string) (index.Plugin, error) {
...
+// ReadReceiptFromFile loads a file from the FS. When receipt file not found, it
+// returns an error that can be checked with os.IsNotExist.
+func ReadReceiptFromFile(path string) (index.Receipt, error) {
+	f, err := os.Open(path)
+	if os.IsNotExist(err) {
+		// TODO(ahmetb): we should use go1.13+ errors.Is construct at call sites to evaluate if an error is os.IsNotExist
+		return index.Receipt{}, err
+	} else if err != nil {
+		return index.Receipt{}, errors.Wrap(err, "failed to open index file")
+	}
```
ahmetb: as far as I can tell this can still be deduplicated (with the same de-dupe technique as `decodeFile` you've done).

corneliusweig:
Interesting. Travis builds seem to be broken right now. I guess we either need to disable travis or this test for now :)

internal/index/indexscanner/scanner_test.go
```diff
@@ -90,6 +90,59 @@ func TestReadPluginFile(t *testing.T) {
 	}
 }
 
+func TestReadReceiptFile(t *testing.T) {
+	type args struct {
+		receiptFilePath string
+	}
+	tests := []struct {
```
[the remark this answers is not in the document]
chriskim06: Ya I was thinking about that. I plan to add more test cases once we add status to a receipt, so this is just to save a bit of work down the line. I can change it though if you'd prefer.

ahmetb:
@corneliusweig do you mind Making the error more Verbose in  Test_fetchLatestTag_GitHubAPI? I suspect it’s hitting API rate limits of travis servers.

internal/index/indexscanner/scanner_test.go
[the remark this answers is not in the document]
corneliusweig: It' fine. Let's leave as-is if you intend to add more tests later.

internal/index/indexscanner/scanner_test.go
```diff
@@ -91,53 +94,50 @@ func TestReadPluginFile(t *testing.T) {
...
 	neverMatch := labels.Set{}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got, err := ReadReceiptFromFile(tt.args.receiptFilePath)
+			tmpDir, cleanup := testutil.NewTempDir(t)
+			defer cleanup()
+
+			receipt := testutil.NewPlugin().WithName("foo").V()
+			receiptFile := filepath.Join(tmpDir.Root(), "receipts", tt.receiptFileName)
```
corneliusweig: Why not simply

```suggestion
			receiptFile := tmpDir.Path("receipts/"+tt.receiptFileName)
```

internal/index/indexscanner/scanner_test.go
```diff
@@ -91,53 +94,50 @@ func TestReadPluginFile(t *testing.T) {
...
-				return
 			}
 			if err != nil {
 				t.Error(err)
-				return
 			}
-			if got.Name != "foo" && got.Kind != "Plugin" {
-				t.Errorf("ReadReceiptFromFile() has not parsed the metainformations %v", got)
-				return
+			if cmp.Diff(got.Plugin, receipt) != "" {
```
corneliusweig: `cmp.Diff` already formats the differences in an easy to read way. You should do
```go
if diff := cmp.Diff(got.Plugin, receipt); diff != "" {
   t.Errorf("Parsed receipt has unexpected content: %s", diff)
}
```

internal/index/indexscanner/scanner.go
```diff
@@ -86,38 +86,59 @@ func LoadPluginByName(pluginsDir, pluginName string) (index.Plugin, error) {
...
+	defer f.Close()
+	err = decodeFile(f, &as)
+	if err != nil {
+		return errors.Wrap(err, "failed to decode plugin manifest")
+	}
+	return nil
+}
+
+// decodeFile tries to decode a plugin/receipt
+func decodeFile(r io.Reader, as interface{}) error {
```
corneliusweig: IMHO, this should be 
```suggestion
func decodeFile(r io.ReadCloser, as interface{}) error {
  defer r.Close()
  ...
}
```

The reading happens here, so we should also close here. Then you can delegate all the `defer r.Close()` calls to this function.

internal/index/indexscanner/scanner.go
```diff
@@ -86,38 +86,59 @@ func LoadPluginByName(pluginsDir, pluginName string) (index.Plugin, error) {
...
+	f, err := os.Open(path)
+	if os.IsNotExist(err) {
+		return err
+	} else if err != nil {
+		return errors.Wrapf(err, "failed to open file %s", path)
+	}
+	defer f.Close()
+	err = decodeFile(f, &as)
+	if err != nil {
+		return errors.Wrap(err, "failed to decode plugin manifest")
```
corneliusweig: This function is no longer exclusively used for plugin manifests.
```suggestion
		return errors.Wrap(err, "failed to parse yaml file")
```

internal/index/indexscanner/scanner.go
```diff
@@ -86,38 +86,59 @@ func LoadPluginByName(pluginsDir, pluginName string) (index.Plugin, error) {
...
+func readFromFile(path string, as interface{}) error {
+	f, err := os.Open(path)
+	if os.IsNotExist(err) {
+		return err
+	} else if err != nil {
+		return errors.Wrapf(err, "failed to open file %s", path)
+	}
+	defer f.Close()
+	err = decodeFile(f, &as)
+	if err != nil {
```
corneliusweig: `errors.Wrap` will do nothing if `err == nil`, so you can do simpler
```go
return errors.Wrap(err, "...")
```

internal/index/indexscanner/scanner.go
```diff
@@ -86,38 +86,59 @@ func LoadPluginByName(pluginsDir, pluginName string) (index.Plugin, error) {
...
 
-// DecodePluginFile tries to decodes a plugin manifest from r.
-func DecodePluginFile(r io.Reader) (index.Plugin, error) {
-	var plugin index.Plugin
+// ReadReceiptFromFile loads a file from the FS. When receipt file not found, it
+// returns an error that can be checked with os.IsNotExist.
+func ReadReceiptFromFile(path string) (index.Receipt, error) {
+	var receipt index.Receipt
+	err := readFromFile(path, &receipt)
+	return receipt, err
```
corneliusweig: Please also wrap the error here, like
```suggestion
	return receipt, errors.Wrap(err, "failed to read receipt file")
```

internal/index/indexscanner/scanner.go
```diff
@@ -86,38 +86,59 @@ func LoadPluginByName(pluginsDir, pluginName string) (index.Plugin, error) {
...
-		return index.Plugin{}, err
-	} else if err != nil {
-		return index.Plugin{}, errors.Wrap(err, "failed to open index file")
+	var plugin index.Plugin
+	err := readFromFile(path, &plugin)
+	if err != nil {
+		return plugin, err
 	}
-	return ReadPlugin(f)
+	return plugin, errors.Wrap(validation.ValidatePlugin(plugin.Name, plugin), "plugin manifest validation error")
```
corneliusweig: Oh no.. looks like we swapped the duplication of boilerplate code for opening the file with duplication of our business logic. IMHO it's better if this function continues to call `ReadPlugin` and we duplicate the file-open boilerplate.

This also means, that `readFromFile` becomes obsolete and should be inlined in `ReadReceiptFromFile`.

@ahmetb what do you think?

internal/index/indexscanner/scanner_test.go
chriskim06: Ah I didn't see that function. I'll make the change

internal/index/indexscanner/scanner.go
chriskim06: I tried this earlier but it fails the `TestLoad_preservesNonExistsError` in `receipts_test` since `os.IsNotExist` seems to fail after wrapping the error. Is it enough to check the contents of the error message in the test to make sure the "no such file or directory" message exists?

internal/index/indexscanner/scanner.go
```diff
@@ -86,38 +86,59 @@ func LoadPluginByName(pluginsDir, pluginName string) (index.Plugin, error) {
...
-	f, err := os.Open(path)
-	if os.IsNotExist(err) {
-		// TODO(ahmetb): we should use go1.13+ errors.Is construct at call sites to evaluate if an error is os.IsNotExist
-		return index.Plugin{}, err
-	} else if err != nil {
-		return index.Plugin{}, errors.Wrap(err, "failed to open index file")
+	var plugin index.Plugin
+	err := readFromFile(path, &plugin)
+	if err != nil {
+		return plugin, err
```
ahmetb: since you aren't wrapping err, just do:

err := readFromFile
return plugin, err

internal/index/indexscanner/scanner.go
ahmetb: no, let's not check error message substrings.
instead, you can directly return the error here.

we wrote that test for a specific purpose: we rely on os.IsNotExist at the call site of these methods.

internal/index/indexscanner/scanner.go
```diff
@@ -86,38 +86,59 @@ func LoadPluginByName(pluginsDir, pluginName string) (index.Plugin, error) {
...
+	return receipt, err
+}
+
+// readFromFile loads a file from the FS into the provided object.
+func readFromFile(path string, as interface{}) error {
+	f, err := os.Open(path)
+	if os.IsNotExist(err) {
+		return err
+	} else if err != nil {
+		return errors.Wrapf(err, "failed to open file %s", path)
```
ahmetb: you actually don't need to wrap this.
most "file" errors in Go are actually quite friendly. They already contain the `path`, and return messages like:

    open: /foo/bar is not readable

internal/index/indexscanner/scanner_test.go
```diff
@@ -90,6 +93,56 @@ func TestReadPluginFile(t *testing.T) {
 	}
 }
 
+func TestReadReceiptFile(t *testing.T) {
```
ahmetb: I'm not particularly happy with the test below, as it exercises only a few things. For example, it doesn't check whether the status fields are parsed correctly.

We just need 1 test case that makes sure the fields are parsed decently. and the ``json:"foo,omitempty"`` field-tags are accurate.

So I suggest we do this:

1. Add a few receipt test files in `testdata/receipts/foo.yaml`

2. Parse the receipt file from disk.

3. Build as `NewReceipt().WithXxx(...)....V()` and use `cmp.Diff` to do equality check.

@corneliusweig wdyt?

internal/index/indexscanner/scanner.go
ahmetb: It seems ok to me now.

pkg/index/types.go
ahmetb: Yeah it's fine + we can refactor easily.

internal/testutil/tempdir.go
```diff
@@ -77,11 +75,11 @@ func (td *TempDir) Write(file string, content []byte) *TempDir {
 	return td
 }
 
-func (td *TempDir) WritePlugin(file string, plugin index.Plugin) *TempDir {
+func (td *TempDir) WriteYaml(file string, obj interface{}) *TempDir {
```
ahmetb: abbrevations are all-caps in go: WriteYAML

internal/index/indexscanner/scanner.go
corneliusweig: Ok, then let's settle with the current version.

internal/index/indexscanner/scanner_test.go
corneliusweig: Well, so far `Receipt` does not have any extra fields beside the ones from `Plugin`. So we are cutting corners by parsing a plugin yaml as a receipt. I think that's ok for now, but we need to ensure that we extend this test as suggested by Ahmet.

And I'm in for the receipt builder pattern. The sooner we have it, the better.

ahmetb: Yeah in the next PR, we should add the `status` field and start working with it to develop tests like this (where we can roundtrip the type i.e. write/parse a YAML file).

chriskim06: So is this ok as is for now while status doesn't exist?

ahmetb: yeah not so useful but helps us pass the tests. 
feel free to remove it altogether or add TODO that captures comment above, but this should be the next PR anyway.

chriskim06: Ok I'll just remove it from this PR and add the tests in the next one when they can be more meaningful with the new status field.

ahmetb:
We can go ahead with this if you merge the small outstanding style comments.

corneliusweig:
/lgtm
/approve
Thanks, @chriskim06, for bearing with us. It always takes some time to get used to the coding style of a new project. But we're getting there! And having two reviewers is also challenging :)

corneliusweig:
/lgtm
:confetti_ball:

PR #513, opened 2020-02-26
Use static URL to download assets from latest release

corneliusweig:
This also simplifies the release process for krew, because the version
number does not need to be updated when releasing a new version.

Fixes #...
Related issue: #...

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Check out the list of [kubectl plugins available on krew][list] or just run
     ```sh
     (
       set -x; cd "$(mktemp -d)" &&
-      curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/download/v0.3.4/krew.{tar.gz,yaml}" &&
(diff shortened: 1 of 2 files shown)


corneliusweig:
@ahmetb was there a particular reason why the version should be hardcoded and thus visible that I'm not aware of?

ahmetb:
I think we didn’t know “latest” worked?
That said I don’t understand removing the tagging instructions in Releasing.md.

corneliusweig:
Oh it's because for a release we needed to adjust the hardcoded version number in `README.md`. That change needed to be committed, pushed, built, and fetched. All of these steps can now be skipped. The tagging step is still there I hope :).

ahmetb:
Hmm but we still need this:

           git commit -am "Release ${TAG:?TAG required}"	

This ensures we have a clean commit that tags the release (we can add `--allow-empty`). 

I frequently follow this pattern in my repos, it helps a lot in the GitHub web UI, as users go to a "commit" that explains the tag; not some irrelevant patch.

corneliusweig:
Ok, that makes sense. Added the instructions for creating a release commit back in.

PR #520, opened 2020-02-27
site: Add netlify.toml

ahmetb:
From: https://github.com/kubernetes/community/blob/master/github-management/subproject-site-requests.md#example-netlify-configuration

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,0 +1,16 @@
+[build]
+base = "site/"
+publish = "site/public/"
+command = "hugo --gc"
+
+[build.environment]
+HUGO_VERSION = "0.65.1"
+
+[context.production.environment]
+HUGO_ENV = "production"
+
+[context.deploy-preview]
+command = "hugo --enableGitInfo --buildFuture -b $DEPLOY_PRIME_URL"
+
+[context.branch-deploy]
+command = "hugo --enableGitInfo --buildFuture -b $DEPLOY_PRIME_URL"


netlify.toml
```diff
@@ -0,0 +1,17 @@
...
+publish = "site/public/"
+command = "hugo --gc"
+
+[build.environment]
+HUGO_VERSION = "0.65.1"
+
+[context.production.environment]
+HUGO_ENV = "production"
+HUGO_ENABLEGITINFO = "true"
+
```
corneliusweig: Wouldn't it make sense to add this already?
```
HUGO_BASEURL = "https://krew.sigs.k8s.io/"
```

netlify.toml
```diff
@@ -0,0 +1,17 @@
...
+base = "site/"
+publish = "site/public/"
+command = "hugo --gc"
+
+[build.environment]
+HUGO_VERSION = "0.65.1"
+
+[context.production.environment]
+HUGO_ENV = "production"
+HUGO_ENABLEGITINFO = "true"
```
corneliusweig: I suspect that we don't need both the env var and the flag? Let's choose one.

netlify.toml
ahmetb: it's already in config.yaml, so I don't think it's needed.

netlify.toml
ahmetb: u r right.

netlify.toml
corneliusweig: I see! makes sense

PR #524, opened 2020-02-28
site: robots.txt

ahmetb:
Prevent Netlify preview environments from being indexed by search bots.

If HUGO_ENV!=production, we return Disallow:/


corneliusweig:
Good catch!
/lgtm
/approve

PR #525, opened 2020-02-28
site: add gitinfo and edit links to footer

ahmetb:
* pass page variables to footer

* use .GitInfo and .File variables to show something like this in the footer
      for each page:

        [Edit Page] - Built from commit [xxxx]

* enableGitInfo both locally and on Netlify builds


corneliusweig:
Looks really nice! Thanks for your love to all the little details that make a great docs site!

/lgtm
/approve

PR #526, opened 2020-02-29
Add receipt status

### LINKED ISSUE #507: Add a new receipt type

chriskim06:
This adds the status object to the receipt type that was added earlier. The changes are pretty small, most of the additions are from tests implemented in the way outlined [here](https://github.com/kubernetes-sigs/krew/pull/512#discussion_r384674035).

Fixes #507 
Related issue: #483


internal/installation/receipt/receipt.go
```diff
@@ -44,6 +44,7 @@ func Load(path string) (index.Receipt, error) {
 	return indexscanner.ReadReceiptFromFile(path)
 }
 
+// receiptFromPlugin creates a receipt object using the plugin and index name.
 func receiptFromPlugin(indexName string, plugin index.Plugin) index.Receipt {
```
ahmetb: Use capitalize Plugin in godoc 
Also let’s move Plugin to be first param. It’s likely we’ll add other params, so it’d better if it is first parameter.

internal/index/indexscanner/testdata/testindex/receipts/custom.yaml
```diff
@@ -0,0 +1,35 @@
...
+      to: .
+    selector:
+      matchLabels:
+        arch: amd64
+        os: linux
+    sha256: deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef
+    uri: http://example.com/
+  shortDescription: in-memory test plugin object
+  version: v1.0.0-test.1
+Status:
```
ahmetb: This shouldnt be parsing, right?

internal/index/indexscanner/testdata/testindex/receipts/default-explicit.yaml
```diff
@@ -0,0 +1,35 @@
...
+# limitations under the License.
+
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: default-explicit-plugin
+spec:
+  platforms:
+  - bin: kubectl-test-plugin
+    files:
```
ahmetb: Lets remove this as it has a default value and we aren’t testing this field’s parsed value.

internal/index/indexscanner/testdata/testindex/receipts/custom.yaml
chriskim06: I didn't actually notice that I capitalized those. Let me look into it.

chriskim06: I thought the json struct tags would affect how it was parsed but it doesn't seem to be that way. When the receipt is written it uses lower case letters for those keys. Am I missing something?

ahmetb: This just shouldn’t be parsing accurately.
We currently ignore unknown fields intentionally.

chriskim06: I'm a little confused I think. What do you mean by unknown fields in this case? If you're referring specifically to the test receipt that has a custom index set in its status, I mainly added that even though only default can be set in a receipt because `receipt.Load` doesn't validate the index that is parsed. Let me know if I misunderstood.

ahmetb: All our fields are case-sensitive.
So as you can imagine, there can't be a field capitalized like `Status`.
So typing `Status` here should be the same as `Asadklfjdksfj` (which would be ignored).

internal/index/indexscanner/scanner_test.go
```diff
@@ -90,6 +95,63 @@ func TestReadPluginFile(t *testing.T) {
...
+		receiptFilePath string
+		pluginName      string
+		indexName       string
+	}
+	tests := []struct {
+		name string
+		args args
+	}{
+		{
+			name: "read receipt of plugin from default index without status",
```
ahmetb: `from default index` part is superfluous.

this just a receipt without `status` field.

internal/index/indexscanner/scanner_test.go
```diff
@@ -90,6 +95,63 @@ func TestReadPluginFile(t *testing.T) {
...
+			args: args{
+				receiptFilePath: filepath.Join(testdataPath(t), "testindex", "receipts", "default.yaml"),
+				pluginName:      constants.DefaultIndexName + "-plugin",
```
ahmetb: using this const here actually doesn't achieve anything.

just call it "foo"?

internal/index/indexscanner/scanner_test.go
```diff
@@ -90,6 +95,63 @@ func TestReadPluginFile(t *testing.T) {
 	}
 }
 
+func TestReadReceiptFromFile(t *testing.T) {
+	type args struct {
+		receiptFilePath string
+		pluginName      string
+		indexName       string
```
ahmetb: these aren't args.

these are `wantPluginName` `wantIndexName`

You should actually make this `wantStatus` and use cmp.Diff to compare parsed status vs wantStatus.

ahmetb: you can see how other table tests do this in other files.

internal/index/indexscanner/scanner_test.go
```diff
@@ -90,6 +95,63 @@ func TestReadPluginFile(t *testing.T) {
...
+			testPlugin := testutil.NewPlugin().WithName(tt.args.pluginName).V()
+			receipt := testutil.NewReceipt().WithPlugin(testPlugin)
+			if tt.args.indexName != "" {
+				receipt = receipt.WithStatus(index.ReceiptStatus{
+					Source: index.SourceIndex{
+						Name: tt.args.indexName,
+					},
+				})
+			}
+			testReceipt := receipt.V()
```
ahmetb: just merge .V() to the statement before? you aren't using the `receipt` var

internal/index/indexscanner/testdata/testindex/receipts/default-explicit.yaml
```diff
@@ -0,0 +1,35 @@
+# Copyright 2019 The Kubernetes Authors.
```
ahmetb: I must admit, I kinda dislike we add 3 nearly identical test files just for this. :(
But I see no other way to test this either.

pkg/index/types.go
```diff
@@ -62,4 +62,17 @@ type FileOperation struct {
 // Receipt describes a plugin receipt file.
 type Receipt struct {
 	Plugin `json:",inline" yaml:",inline"`
+
+	Status ReceiptStatus `json:"status,omitempty"`
```
ahmetb: why `,omitempty`? we want receipts to always have `status` (except today's receipts, which currently don't have, but once they are updated, they should).

internal/index/indexscanner/testdata/testindex/receipts/custom.yaml
chriskim06: I tried changing `version` to `Version` in my receipts in my normal krew installation and was still able to parse receipts in `kubectl krew list` (where it loads the receipts and gets `.Spec.Version` from them). Maybe I'm doing something wrong. I also thought they were case sensitive and made a mistake by accidentally making those capitalized in the test file, but didn't notice it since the tests were passing and they were still parsing correctly.

pkg/index/types.go
chriskim06: This was just me misremembering what we talked about before, I wrote up the receipt issue before you and I cleared that up and I copied over the struct definitions I wrote there and forgot to remove this. I'll get rid of it :+1:

internal/index/indexscanner/testdata/testindex/receipts/custom.yaml
ahmetb: That's very odd. Let's continue thinking they're case-sensitive.

internal/index/indexscanner/scanner_test.go
```diff
@@ -90,6 +94,58 @@ func TestReadPluginFile(t *testing.T) {
...
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotReceipt, err := ReadReceiptFromFile(tt.args.receiptFilePath)
+			if err != nil {
+				t.Fatalf("ReadReceiptFromFile() error: %v", err)
+			}
+			if diff := cmp.Diff(gotReceipt.Status, tt.wantStatus); diff != "" {
```
ahmetb: I think cmp.Diff is frequently used as cmp.Diff(want, got) –you can remember from git diff, like: (old,new). This will work regardless, but just a nit.

internal/installation/receipt/receipt_test.go
```diff
@@ -59,7 +60,11 @@ func TestLoad(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	testPluginReceipt := index.Receipt{Plugin: testPlugin}
+	testPluginReceipt := testutil.NewReceipt().WithPlugin(testPlugin).WithStatus(index.ReceiptStatus{
```
ahmetb: (Couldn't comment there but) on line :33 in `TestStore()`, shouldn't we be using the new `testutil.NewReceipt()` as well?

chriskim06: Ya I missed that test. I'll update it :+1:

internal/index/indexscanner/scanner_test.go
```diff
@@ -90,6 +94,58 @@ func TestReadPluginFile(t *testing.T) {
 	}
 }
 
+func TestReadReceiptFromFile(t *testing.T) {
+	type args struct {
+		receiptFilePath string
+	}
+	tests := []struct {
+		name       string
+		args       args
```
corneliusweig: I haven't read the previous review comments, but is there a reason why `receiptFilePath` is wrapped in this args struct? There could as well just be a single `receiptFilePath` field.

internal/index/indexscanner/scanner_test.go
```diff
@@ -90,6 +94,58 @@ func TestReadPluginFile(t *testing.T) {
...
+			name: "read receipt of plugin from default index with status",
+			args: args{
+				receiptFilePath: filepath.Join(testdataPath(t), "testindex", "receipts", "foo-explicit.yaml"),
+			},
+			wantStatus: index.ReceiptStatus{
+				Source: index.SourceIndex{
+					Name: constants.DefaultIndexName,
+				},
+			},
+		},
```
corneliusweig: I think this and the next test case are pretty much identical. I wouldn't use the `default` index as an explicitly given status value, because it is somewhat special. Just use something that is unlikely to be used by accident, like `test-custom` and everything is fine.

Then you can also get rid of one of the two additional receipt yamls.

internal/index/indexscanner/scanner_test.go
```diff
@@ -90,6 +94,58 @@ func TestReadPluginFile(t *testing.T) {
...
+	}{
+		{
+			name: "read receipt of plugin without status",
+			args: args{
+				receiptFilePath: filepath.Join(testdataPath(t), "testindex", "receipts", "foo.yaml"),
+			},
+			wantStatus: index.ReceiptStatus{},
+		},
+		{
+			name: "read receipt of plugin from default index with status",
```
corneliusweig:
```suggestion
			name: "read receipt with explicit source index",
```

internal/installation/receipt/receipt_test.go
```diff
@@ -22,26 +22,26 @@ import (
...
 	testPlugin := testutil.NewPlugin().WithName("some-plugin").WithPlatforms(testutil.NewPlatform().V()).V()
+	testReceipt := testutil.NewReceipt().WithPlugin(testPlugin).V()
 	dest := tmpDir.Path("some-plugin.yaml")
 
 	if err := Store(testPlugin, dest); err != nil {
 		t.Fatal(err)
 	}
 
-	actual, err := indexscanner.LoadPluginByName(tmpDir.Root(), "some-plugin")
+	actual, err := indexscanner.ReadReceiptFromFile(tmpDir.Path("some-plugin.yaml"))
```
corneliusweig:
```suggestion
	actual, err := indexscanner.ReadReceiptFromFile(dest)
```

internal/index/indexscanner/scanner_test.go
```diff
@@ -90,6 +94,58 @@ func TestReadPluginFile(t *testing.T) {
...
+	tests := []struct {
+		name       string
+		args       args
+		wantStatus index.ReceiptStatus
+	}{
+		{
+			name: "read receipt of plugin without status",
+			args: args{
+				receiptFilePath: filepath.Join(testdataPath(t), "testindex", "receipts", "foo.yaml"),
+			},
```
corneliusweig: You could also use your testdata builder `NewReceipt().WithXX().V()` and store that in a temporary directory with `tmpDir.WriteYAML(...)`.
Why did you opt for test fixtures instead?

ahmetb:
/lgtm
@corneliusweig for a quick review

internal/index/indexscanner/scanner_test.go
chriskim06: Its left over from a previous version. I had some other things there before that I moved out, I'll make the change.

internal/index/indexscanner/scanner_test.go
chriskim06: I wanted to test 3 types of receipts in this: one from a custom index, one from the default index before migration, and one from the default index after migration (with/without status object). If that seems excessive I can remove some of them.

internal/index/indexscanner/scanner_test.go
chriskim06: Ya I can change it to this. I think I did it this way because this is how it was outlined in the previous receipts PR and I just had that on my mind while I was working on this initially.

internal/index/indexscanner/scanner_test.go
ahmetb: Honestly, `name: custom` and `name: default` are equivalent things from what they're testing. So one of them is redundant.

Similarly, not having `status:` field is _expected_ to yield the zero-value of that field (that's just what YAML/JSON libraries do). So we aren't testing anything valuable there either.

All I'm interested in testing is "round-tripping the YAML parsing of field names", which can be done in 1 file with:

```
[...]
status:
  source:
    name: foo
```

ahmetb:
It's also currently unclear to me where we should be doing the "defaulting" of the missing `status.index.name`. Potentially parsing function could do that.

WDYT @corneliusweig?

internal/index/indexscanner/scanner_test.go
chriskim06: Sounds good, I'll simplify the tests.

corneliusweig:
Doing the defaulting while reading leads to the least duplication, so let's do it that way. Good idea!

internal/index/indexscanner/scanner_test.go
corneliusweig: Yeah, we talked about both options. Back then we didn't have the receipt builder, so the static file was easier. But now, I think the dynamic approach is far better.

internal/index/indexscanner/scanner.go
```diff
@@ -108,6 +108,9 @@ func ReadPlugin(f io.ReadCloser) (index.Plugin, error) {
 func ReadReceiptFromFile(path string) (index.Receipt, error) {
 	var receipt index.Receipt
 	err := readFromFile(path, &receipt)
+	if receipt.Status.Source.Name == "" {
+		receipt.Status.Source.Name = constants.DefaultIndexName
```
corneliusweig: Aren't you missing a test now? I think there should be one where there's no status in the original manifest, but when you read the receipt, you should check that the default index was filled in.

pkg/index/types.go
```diff
@@ -62,4 +62,17 @@ type FileOperation struct {
 // Receipt describes a plugin receipt file.
 type Receipt struct {
 	Plugin `json:",inline" yaml:",inline"`
+
+	Status ReceiptStatus `json:"status"`
```
corneliusweig: When the json tag does not enable the `omitempty` option, are files without that field valid during unmarshalling? I'm not 100% sure about this. This would affect all existing receipts which do not have the status field.
cc @ahmetb

ahmetb: omitempty is NOOP for unmarshal.

This specifically would ensure `status:` field is always saved explicitly in marshal path.

ahmetb:
/approve
we're close1!!!1

pkg/index/types.go
chriskim06: Ya its still valid. I'm pretty sure after working on this, that those struct tags only affect it when the object is marshalled. I saw that setting the key name `json:"Status"` vs `json:"status"` didn't actually affect how it was unmarshalled (more info [here](https://github.com/kubernetes-sigs/krew/pull/526#discussion_r386545371)).

internal/index/indexscanner/scanner.go
chriskim06: Ah you're right, my bad. I've been changing this a lot and missed that. I can add it back in.

internal/index/indexscanner/scanner_test.go
```diff
@@ -96,24 +96,58 @@ func TestReadPluginFile(t *testing.T) {
...
 
 func TestReadReceiptFromFile(t *testing.T) {
-	tmpDir, cleanup := testutil.NewTempDir(t)
-	defer cleanup()
-
-	name := "foo"
-	wantStatus := &index.ReceiptStatus{
-		Source: index.SourceIndex{
-			Name: name,
+	type args struct {
```
ahmetb: for this time it's ok but in the future I recommend avoiding single-field `type args`

chriskim06: Crap sorry about that, you mentioned that before. I'll remember next time and change this in a follow up PR most likely.

corneliusweig:
:confetti_ball: One more step! So nice to see this going forward!

PR #530, opened 2020-03-03
site: make footer show last updated

ahmetb:
Little JS snippet to actually show something like:

    · Page last updated [3 days ago](link to commit).

Instead of:

    · Build from commit [abcdef](link to commit).

Written in ES6 syntax (https://caniuse.com/#search=es6).

### DIFF
--- a/site/layouts/partials/footer.html
+++ b/site/layouts/partials/footer.html
@@ -6,15 +6,14 @@
                 <a href="https://github.com/kubernetes/community/blob/master/sig-cli/README.md#cli-special-interest-group" target="_blank" class="text-info">
                     SIG CLI</a> project.
                 <p>
-                    {{ with .File }}
+                    {{- with .File -}}
                         <a href="http://sigs.k8s.io/krew/site/content/{{- .File.Path -}}" target="_blank" class="text-info">
                         Edit Page</a>
-                    {{ end }}
-
-                    {{ if .GitInfo }}
+                    {{- end -}}
+                    {{- if .GitInfo }}
                         &middot;
                         <span id="lastmod"></span>
-                    {{ end}}
+                    {{- end}}
                 </p>
             </footer>
         </div>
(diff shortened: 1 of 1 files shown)


site/layouts/partials/footer.html
```diff
@@ -12,16 +12,40 @@
...
-        </footer>
+            </footer>
         </div>
     </div><!-- .container -->
+
+    {{ if .GitInfo }}
+    <script>
+        var lastMod = {{- .Lastmod.Unix -}};
+        var commitHash = '{{- .GitInfo.Hash -}}';
+        function writeLastMod(){
```
corneliusweig: I find that `window.lastMod` rather ugly. After all, this sets up global variables. What about passing those two values to the function, like so:
```js
function writeLastMod(lastMod, commitHash) { ... }
```
Then you could also move the function into a static resource script which is included in the header. WDYT?

ahmetb: I think we can keep the script here. It doesn’t make a lot of difference. If anything we don’t worry about old versions being cached (and hence incompatible with the new html) this way. 

But you are right. I can add the template var during invocation.

corneliusweig: Yeah, extracting the script is not critical. Either way is fine.

site/layouts/partials/footer.html
```diff
@@ -6,22 +6,43 @@
...
+                text = `${Math.trunc(days/30)} months ago`
+            } else {
+                text = `${Math.trunc(days/365)} years ago`
+            }
+            document.getElementById('lastmod').innerHTML = `Page last updated ` +
+                `<a class="text-info" target="_blank"` +
+                `href="https://github.com/kubernetes-sigs/krew/commit/${commitHash}">` +
+                `${text}</a>.`;
+        };
+    writeLastMod({{- .Lastmod.Unix -}}, '{{- .GitInfo.Hash -}}');
```
corneliusweig: Shouldn't this be indented one more level?

corneliusweig:
Works like a charm!
/lgtm
/approve
/hold

site/layouts/partials/footer.html
ahmetb: Done.

corneliusweig:
/lgtm
/approve
Sorry it took so long 🙏

PR #537, opened 2020-03-06
Add index list command

chriskim06:
Initial implementation of `kubectl krew index list`. This prints the indexes in a table in the same format as `kubectl krew list`:
```
$ KREW_ROOT="/home/kimc/playground" X_KREW_ENABLE_MULTI_INDEX=1 ./bin/krew-linux_amd64 index list
INDEX    URL
default  https://github.com/kubernetes-sigs/krew-index.git
```

I slightly refactored the `exec` function in gitutil to return the output of the command. The existing uses of it don't need it but I'm using the output from `git config --get remote.origin.url` to get the index's remote url. Let me know if there's a better way to get the url from git, that was the only way I knew how.

Related issue: #483

### DIFF
--- a/cmd/krew/cmd/index.go
+++ b/cmd/krew/cmd/index.go
@@ -27,8 +27,8 @@ import (
 // indexCmd represents the index command
 var indexCmd = &cobra.Command{
 	Use:    "index",
-	Short:  "Add, remove, and list indexes",
-	Long:   "Perform krew index commands such as adding, removing, and listing indexes.",
+	Short:  "Manage custom plugin indexes",
+	Long:   "Manage which repositories are used to discover and install plugins from.",
 	Args:   cobra.NoArgs,
 	Hidden: true, // TODO(chriskim06) remove this once multi-index is enabled
 }
--- a/cmd/krew/cmd/index.go
+++ b/cmd/krew/cmd/index.go
@@ -29,7 +29,7 @@ import (
 var indexCmd = &cobra.Command{
 	Use:    "index",
 	Short:  "Perform krew index commands",
-	Long:   "Show a list of installed kubectl plugins and their versions.",
+	Long:   "Perform krew index commands such as adding and removing indexes.",
 	Args:   cobra.NoArgs,
 	Hidden: true, // remove this once multi-index is enabled
 }
@@ -40,7 +40,7 @@ var indexListCmd = &cobra.Command{
 	Long: `Print a list of configured indexes.
 
 This command prints a list of indexes. It shows the name and the remote URL for
-each configured index.`,
+each configured index in table format.`,
 	Args: cobra.NoArgs,
 	RunE: func(_ *cobra.Command, _ []string) error {
 		dirs, err := ioutil.ReadDir(paths.IndexBase())
--- a/cmd/krew/cmd/index.go
+++ b/cmd/krew/cmd/index.go
@@ -1,0 +1,63 @@
+// Copyright 2019 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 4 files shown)


ahmetb:
Needs integration tests. :)

cmd/krew/cmd/index.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/spf13/cobra"
+
+	"sigs.k8s.io/krew/internal/gitutil"
+	"sigs.k8s.io/krew/pkg/constants"
+)
+
+// indexCmd represents the index command
+var indexCmd = &cobra.Command{
+	Use:    "index",
+	Short:  "Perform krew index commands",
```
ahmetb: this doesn't look good in the help message:


```
Available Commands:

  index       Perform krew index commands
``` 

let's say something more meaningful.

cmd/krew/cmd/index.go
```diff
@@ -0,0 +1,68 @@
...
+		if err != nil {
+			return errors.Wrapf(err, "failed to read directory %s", paths.IndexBase())
+		}
+		var rows [][]string
+		for _, dir := range dirs {
+			indexName := dir.Name()
+			remote, err := gitutil.GetRemoteURL(paths.IndexPath(indexName))
+			if err != nil {
+				return errors.Wrapf(err, "failed to list the remote URL for index %s", indexName)
+			}
```
ahmetb: Soon you'll find out you need a package where this lives and is tested.

I propose a package that does

```
ListIndexes(path string) ([]Index, error)
RemoveIndex(path, name string) error
AddIndex...
```

We can start small, but let's add tests.

Right now the code I just highlighted here has 0 tests.

cmd/krew/cmd/index.go
```diff
@@ -0,0 +1,68 @@
...
+	"sigs.k8s.io/krew/pkg/constants"
+)
+
+// indexCmd represents the index command
+var indexCmd = &cobra.Command{
+	Use:    "index",
+	Short:  "Perform krew index commands",
+	Long:   "Perform krew index commands such as adding and removing indexes.",
+	Args:   cobra.NoArgs,
+	Hidden: true, // remove this once multi-index is enabled
```
ahmetb: use `// TODO` syntax

internal/gitutil/git.go
```diff
@@ -72,10 +72,11 @@ func EnsureUpdated(uri, destinationPath string) error {
 
 // GetRemoteURL returns the url of the remote origin
 func GetRemoteURL(dir string) (string, error) {
-	return exec(dir, "config", "--get", "remote.origin.url")
+	remote, err := Exec(dir, "config", "--get", "remote.origin.url")
+	return strings.TrimRight(remote, "\n\r"), err
```
ahmetb: You can use strings.TrimFunc(unicode.IsSpace)

Furthermore, maybe just move this to Exec, as other cmds are likely going to need it.

cmd/krew/cmd/index.go
```diff
@@ -15,23 +15,22 @@
...
+	"sigs.k8s.io/krew/internal/index/indexoperations"
 	"sigs.k8s.io/krew/pkg/constants"
 )
 
 // indexCmd represents the index command
 var indexCmd = &cobra.Command{
 	Use:    "index",
-	Short:  "Perform krew index commands",
-	Long:   "Perform krew index commands such as adding and removing indexes.",
+	Short:  "Add, remove, and list indexes",
```
ahmetb: Let’s just say:

> Manage custom plugin indexes

& long

> Manage which repositories are used to discover and install plugins from.

ahmetb:
/lgtm
/approve
/hold
only small comments.

internal/gitutil/git.go
chriskim06: Ended up using `strings.TrimSpace`, the linter complained about `strings.TrimFunc('...', unicode.IsSpace)`

ahmetb:
/lgtm
/hold cancel
Thank you. 🙏

PR #540, opened 2020-03-08
site: Alternative layout

ahmetb:
Instead of a shadowed box in the middle of the screen, sticking navbar to the
top and floating content in the middle. This is a bit more consistent with
current design patterns and more mobile-friendly.

If we like the preview, we can consider incorporating.

### DIFF
--- a/site/layouts/partials/footer.html
+++ b/site/layouts/partials/footer.html
@@ -5,7 +5,7 @@
             Krew is a Kubernetes
                 <a href="https://github.com/kubernetes/community/blob/master/sig-cli/README.md#cli-special-interest-group" target="_blank" class="text-info">
                     SIG CLI</a> project.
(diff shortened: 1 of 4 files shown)


corneliusweig:
In Chrome 80, the top nav bar doesn't stick. It scrolls away with the remaining content. Is that the intended behavior?

Other than that, the new page layout looks nice. It's a good idea to adopt a more modern design.

ahmetb:
> In Chrome 80, the top nav bar doesn't stick. It scrolls away with the remaining content. Is that the intended behavior?

Intended, but can be easily corrected. I didn't mean "stick" in that way but we can make it sticky/fixed.

corneliusweig:
/lgtm
/approve
Let's decide on the stickiness in a follow-up.

PR #542, opened 2020-03-09
Closes #541: isSupportedAPIVersion's check is now relaxed

### LINKED ISSUE #541: Plan for updating apiVersion in manifests
So far, Krew has successfully maintained backwards compatibility for plugin
manifests. However, we at some point we'll need to move forward.

Currently our manifest validation code does a strict check of apiVersion
equality to a particular string (`krew.googlecontainertools.github.com/v1alpha2`).

Since eventually we will change this value, we must slowly start changing the
strict == check behaviour by relaxing it. Ideally we would have a list of supported
versions. That way we can maintain some backwards-compatibility for N-1 or N-2
apiVersions (by doing in-memory manifest conversion, if necessary).

As an exercise, we can move to krew.sigs.k8s.io/v1beta1 apiGroupVersion and
still have the v1alpha2 manifests working.

Sooner we get to this, the better, as we'll always have users on older versions.

/help
/good-first-issue
/priority important-soon

nirex0:
isSupportedAPIVersion function now checks to see if the input (apiVersion) exists within a list defined in pkg\constants\constants.go, if so it returns true, otherwise false is returned.

The function GetSupportedAPIVersions in constants.go is how the supported versions list is accessed. (It is hardcoded however, can be modified to use an external file instead)

Fixes #541
Related issue: #541

### DIFF
--- a/internal/index/validation/validate.go
+++ b/internal/index/validation/validate.go
@@ -55,7 +55,13 @@ func IsSafePluginName(name string) bool {
 }
 
 func isSupportedAPIVersion(apiVersion string) bool {
-	return apiVersion == constants.CurrentAPIVersion
+	versions := constants.GetSupportedAPIVersions()
+	for version := range versions {
+		if apiVersion == versions[version] {
+			return true
+		}
+	}
+	return false
 }
 
 func isValidSHA256(s string) bool { return validSHA256.MatchString(s) }
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -28,3 +28,8 @@ const (
 	// the features around multiple indexes (this will be removed later on).
 	EnableMultiIndexSwitch = "X_KREW_ENABLE_MULTI_INDEX"
 )
+
+// GetSupportedAPIVersions returns a list containing all the supported versions of the API in from of strings.
+func GetSupportedAPIVersions() [1]string {
+	return [1]string{"krew.googlecontainertools.github.com/v1alpha2"}
+}


nirex0:
I signed it

nirex0:
I signed it

nirex0:
"I signed it"

nirex0:
I signed it

corneliusweig:
Hi @nirex0, thanks for working on this.
It looks like there was some misunderstanding, though. Your patch prepares to relax the api version check, so that it accepts several api versions. However, the purpose of #541 is to implement a new api version and do an in-memory conversion from the old to the new api version on the fly. Does that make sense?

nirex0:
Right, so the system needs to be implemented in a way that does a version comparison and determines the result. 

Please explain thoroughly how the ("**v1alpha2**") part of the API version will change, since the pattern is needed if what is required, is a fully automated system.

ahmetb:
It's not fully automated. This system would let us convert "supported" older apiVersions to the newer ones. But right now we have only 1. So this patch doesn't really solve the problem. Here's why:

1. User 1 is on krew 0.N that supports ONLY the old apiVersion
2. User 2 is on latest krew v0.N+1, that supports both apiVersions.
3. We update `krew-index` with the new apiVersion (v1beta1).
4. User 1 runs `krew update` &rarr; their krew will fail, saying "apiVersion is invalid" so half of the krew commands will print warnings and/or fail.

The only way I see forward here is that **we should disable apiVersion check** for a while (1-2 releases), then re-enable again to check one of supported apiVersions.

Without doing that, we are currently unable to make any apiVersion changes for 3-4 months (a sane period where people would upgrade their krew versions, hopefully).

PR #545, opened 2020-03-10
Add index add command

chriskim06:
Relatively small change. The majority of it is the test I added. I wasn't sure how to test the case where an index is added since it relies on git to clone the repo, so the only test cases I have are negative ones where the index with a name has already been configured and an invalid url is provided for the clone operation. I was thinking of using `constants.IndexURI` in a test case but didn't like that the unit test would need to be able to reach github so I didn't add it. 

Related issue: #483

### DIFF
--- a/cmd/krew/cmd/index.go
+++ b/cmd/krew/cmd/index.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"k8s.io/klog"
 
 	"sigs.k8s.io/krew/internal/index/indexoperations"
 	"sigs.k8s.io/krew/pkg/constants"
@@ -54,9 +55,27 @@ each configured index in table format.`,
 	},
 }
 
+var indexAddCmd = &cobra.Command{
+	Use:   "add",
+	Short: "Add a new index",
+	Long: `Configure a new index to install plugins from.
+
+Example:
+  kubectl krew index add index-name index-url`,
+	RunE: func(_ *cobra.Command, args []string) error {
+		err := indexoperations.AddIndex(paths.IndexBase(), args[0], args[1])
+		if err != nil {
+			return errors.Wrapf(err, "failed to add index %s: %s", args[0], args[1])
+		}
+		klog.Infoln("Successfully added index")
+		return nil
+	},
+}
+
 func init() {
 	if _, ok := os.LookupEnv(constants.EnableMultiIndexSwitch); ok {
 		indexCmd.AddCommand(indexListCmd)
+		indexCmd.AddCommand(indexAddCmd)
 		rootCmd.AddCommand(indexCmd)
 	}
 }
--- a/cmd/krew/cmd/index.go
+++ b/cmd/krew/cmd/index.go
@@ -19,7 +19,6 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
-	"k8s.io/klog"
 
 	"sigs.k8s.io/krew/internal/index/indexoperations"
 	"sigs.k8s.io/krew/pkg/constants"
@@ -46,11 +45,6 @@ each configured index in table format.`,
 		indexes, err := indexoperations.ListIndexes(paths.IndexBase())
 		if err != nil {
 			return errors.Wrap(err, "failed to list indexes")
-		}
-
-		if len(indexes) == 0 {
-			klog.Info(indexoperations.NoIndexMessage)
-			return nil
 		}
 
 		var rows [][]string
--- a/cmd/krew/cmd/index.go
+++ b/cmd/krew/cmd/index.go
@@ -62,7 +62,11 @@ var indexAddCmd = &cobra.Command{
 
 Example:
   kubectl krew index add index-name index-url`,
-	RunE: func(_ *cobra.Command, args []string) error {
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) != 2 {
+			return cmd.Help()
+		}
+
 		err := indexoperations.AddIndex(paths.IndexBase(), args[0], args[1])
 		if err != nil {
 			return errors.Wrapf(err, "failed to add index %s: %s", args[0], args[1])
--- a/cmd/krew/cmd/index.go
+++ b/cmd/krew/cmd/index.go
@@ -56,20 +56,15 @@ each configured index in table format.`,
 }
 
 var indexAddCmd = &cobra.Command{
-	Use:   "add",
-	Short: "Add a new index",
-	Long: `Configure a new index to install plugins from.
-
-Example:
-  kubectl krew index add index-name index-url`,
-	RunE: func(cmd *cobra.Command, args []string) error {
-		if len(args) != 2 {
-			return cmd.Help()
-		}
-
+	Use:     "add",
+	Short:   "Add a new index",
+	Long:    "Configure a new index to install plugins from.",
+	Example: "kubectl krew index add default " + constants.IndexURI,
+	Args:    cobra.ExactArgs(2),
+	RunE: func(_ *cobra.Command, args []string) error {
 		err := indexoperations.AddIndex(paths.IndexBase(), args[0], args[1])
 		if err != nil {
-			return errors.Wrapf(err, "failed to add index %s: %s", args[0], args[1])
+			return errors.Wrap(err, "failed to add index")
 		}
 		klog.Infoln("Successfully added index")
 		return nil
--- a/cmd/krew/cmd/index.go
+++ b/cmd/krew/cmd/index.go
@@ -47,6 +47,12 @@ each configured index in table format.`,
 		if err != nil {
 			return errors.Wrap(err, "failed to list indexes")
 		}
+
+		if len(indexes) == 0 {
+			klog.Info(indexoperations.NoIndexMessage)
+			return nil
+		}
+
 		var rows [][]string
 		for _, index := range indexes {
 			rows = append(rows, []string{index.Name, index.URL})
@@ -62,12 +68,7 @@ var indexAddCmd = &cobra.Command{
 	Example: "kubectl krew index add default " + constants.IndexURI,
 	Args:    cobra.ExactArgs(2),
 	RunE: func(_ *cobra.Command, args []string) error {
-		err := indexoperations.AddIndex(paths.IndexBase(), args[0], args[1])
-		if err != nil {
-			return errors.Wrap(err, "failed to add index")
-		}
-		klog.Infoln("Successfully added index")
-		return nil
+		return indexoperations.AddIndex(paths.IndexBase(), args[0], args[1])
 	},
 }
 
--- a/integration_test/index_test.go
+++ b/integration_test/index_test.go
@@ -17,31 +17,65 @@ package integrationtest
 import (
 	"bytes"
 	"os"
-	"path/filepath"
 	"testing"
 
 	"sigs.k8s.io/krew/pkg/constants"
 )
 
-func TestKrewIndex(t *testing.T) {
+func TestKrewIndexAdd(t *testing.T) {
 	skipShort(t)
 
 	test, cleanup := NewTest(t)
 	defer cleanup()
 
-	initialList := test.WithMigratedIndex().WithEnv(constants.EnableMultiIndexSwitch, 1).Krew("index", "list").RunOrFailOutput()
-	if !bytes.Contains(initialList, []byte(constants.IndexURI)) {
-		t.Fatalf("expected krew-index in output:\n%s", string(initialList))
+	test.WithEnv(constants.EnableMultiIndexSwitch, 1).WithIndex()
+	if err := test.Krew("index", "add").Run(); err == nil {
+		t.Fatal("expected index add with no args to fail")
+	}
+	if err := test.Krew("index", "add", "foo", "https://invalid").Run(); err == nil {
+		t.Fatal("expected index add with invalid URL to fail")
+	}
+	if err := test.Krew("index", "add", "../../usr/bin", constants.IndexURI); err == nil {
+		t.Fatal("expected index add with path characters to fail")
+	}
+	if err := test.Krew("index", "add", "foo", constants.IndexURI).Run(); err != nil {
+		t.Fatalf("error adding new index: %v", err)
+	}
+}
+
+func TestKrewIndexList(t *testing.T) {
+	skipShort(t)
+
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	test.WithEnv(constants.EnableMultiIndexSwitch, 1).WithIndex()
+	out := test.Krew("index", "list").RunOrFailOutput()
+	if !bytes.Contains(out, []byte(constants.DefaultIndexName)) {
+		t.Fatalf("expected index 'default' in output:\n%s", string(out))
 	}
 
-	indexName := "foo"
-	test.Krew("index", "add", indexName, constants.IndexURI).RunOrFail()
-	if _, err := os.Stat(filepath.Join(test.Root(), "index", indexName, ".git")); err != nil {
-		t.Fatalf("error adding index: %s", err)
+	test.Krew("index", "add", "foo", constants.IndexURI).RunOrFail()
+	out = test.Krew("index", "list").RunOrFailOutput()
+	if !bytes.Contains(out, []byte("foo")) {
+		t.Fatalf("expected index 'foo' in output:\n%s", string(out))
+	}
+}
+
+func TestKrewIndexList_NoIndexes(t *testing.T) {
+	skipShort(t)
+
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	test.WithEnv(constants.EnableMultiIndexSwitch, 1).WithIndex()
+	index := test.TempDir().Path("index/" + constants.DefaultIndexName)
+	if err := os.RemoveAll(index); err != nil {
+		t.Fatalf("error removing default index: %v", err)
 	}
 
-	eventualList := test.Krew("index", "list").RunOrFailOutput()
-	if !bytes.Contains(eventualList, []byte(indexName)) {
-		t.Fatalf("expected index 'foo' in output:\n%s", string(eventualList))
+	out := test.Krew("index", "list").RunOrFailOutput()
+	if !bytes.Equal(out, []byte("INDEX  URL\n")) {
+		t.Fatalf("expected index list to be empty:\n%s", string(out))
 	}
 }
--- a/integration_test/index_test.go
+++ b/integration_test/index_test.go
@@ -1,0 +1,47 @@
+// Copyright 2019 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package integrationtest
+
+import (
+	"bytes"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"sigs.k8s.io/krew/pkg/constants"
+)
+
+func TestKrewIndex(t *testing.T) {
+	skipShort(t)
+
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	initialList := test.WithMigratedIndex().WithEnv(constants.EnableMultiIndexSwitch, 1).Krew("index", "list").RunOrFailOutput()
+	if !bytes.Contains(initialList, []byte(constants.IndexURI)) {
+		t.Fatalf("expected krew-index in output:\n%s", string(initialList))
+	}
+
+	indexName := "foo"
+	test.Krew("index", "add", indexName, constants.IndexURI).RunOrFail()
+	if _, err := os.Stat(filepath.Join(test.Root(), "index", indexName, ".git")); err != nil {
+		t.Fatalf("error adding index: %s", err)
+	}
+
+	eventualList := test.Krew("index", "list").RunOrFailOutput()
+	if !bytes.Contains(eventualList, []byte(indexName)) {
+		t.Fatalf("expected index 'foo' in output:\n%s", string(eventualList))
+	}
+}
--- a/integration_test/index_test.go
+++ b/integration_test/index_test.go
@@ -53,7 +53,7 @@ func TestKrewIndexList(t *testing.T) {
 	defer cleanup()
 
 	test.WithEnv(constants.EnableMultiIndexSwitch, 1).WithIndex()
-	out := test.Krew("index", "list").RunOrFailOutput()
+	out := string(test.Krew("index", "list").RunOrFailOutput())
 	if !bytes.Contains(out, []byte(constants.DefaultIndexName)) {
 		t.Fatalf("expected index 'default' in output:\n%s", string(out))
 	}
--- a/integration_test/index_test.go
+++ b/integration_test/index_test.go
@@ -15,7 +15,6 @@
 package integrationtest
 
 import (
-	"bytes"
 	"os"
 	"testing"
 
@@ -53,15 +52,17 @@ func TestKrewIndexList(t *testing.T) {
 	defer cleanup()
 
 	test.WithEnv(constants.EnableMultiIndexSwitch, 1).WithIndex()
-	out := string(test.Krew("index", "list").RunOrFailOutput())
-	if !bytes.Contains(out, []byte(constants.DefaultIndexName)) {
-		t.Fatalf("expected index 'default' in output:\n%s", string(out))
+	out := test.Krew("index", "list").RunOrFailOutput()
+	if indexes := lines(out); len(indexes) < 2 {
+		// the first line is the header
+		t.Fatal("expected at least 1 index in output")
 	}
 
 	test.Krew("index", "add", "foo", test.TempDir().Path("index/"+constants.DefaultIndexName)).RunOrFail()
 	out = test.Krew("index", "list").RunOrFailOutput()
-	if !bytes.Contains(out, []byte("foo")) {
-		t.Fatalf("expected index 'foo' in output:\n%s", string(out))
+	if indexes := lines(out); len(indexes) < 3 {
+		// the first line is the header
+		t.Fatal("expected 2 indexes in output")
 	}
 }
 
@@ -78,7 +79,8 @@ func TestKrewIndexList_NoIndexes(t *testing.T) {
 	}
 
 	out := test.Krew("index", "list").RunOrFailOutput()
-	if !bytes.Equal(out, []byte("INDEX  URL\n")) {
+	if indexes := lines(out); len(indexes) > 1 {
+		// the first line is the header
 		t.Fatalf("expected index list to be empty:\n%s", string(out))
 	}
 }
--- a/integration_test/index_test.go
+++ b/integration_test/index_test.go
@@ -15,10 +15,11 @@
 package integrationtest
 
 import (
-	"bytes"
 	"os"
+	"strings"
 	"testing"
 
+	"sigs.k8s.io/krew/internal/index/indexoperations"
 	"sigs.k8s.io/krew/pkg/constants"
 )
 
@@ -38,7 +39,7 @@ func TestKrewIndexAdd(t *testing.T) {
 	if err := test.Krew("index", "add", "../../usr/bin", constants.IndexURI); err == nil {
 		t.Fatal("expected index add with path characters to fail")
 	}
-	if err := test.Krew("index", "add", "foo", constants.IndexURI).Run(); err != nil {
+	if err := test.Krew("index", "add", "foo", test.TempDir().Path("index/"+constants.DefaultIndexName)).Run(); err != nil {
 		t.Fatalf("error adding new index: %v", err)
 	}
 }
@@ -51,13 +52,13 @@ func TestKrewIndexList(t *testing.T) {
 
 	test.WithEnv(constants.EnableMultiIndexSwitch, 1).WithIndex()
 	out := test.Krew("index", "list").RunOrFailOutput()
-	if !bytes.Contains(out, []byte(constants.DefaultIndexName)) {
+	if !strings.Contains(string(out), constants.DefaultIndexName) {
 		t.Fatalf("expected index 'default' in output:\n%s", string(out))
 	}
 
-	test.Krew("index", "add", "foo", constants.IndexURI).RunOrFail()
+	test.Krew("index", "add", "foo", test.TempDir().Path("index/"+constants.DefaultIndexName)).RunOrFail()
 	out = test.Krew("index", "list").RunOrFailOutput()
-	if !bytes.Contains(out, []byte("foo")) {
+	if !strings.Contains(string(out), "foo") {
 		t.Fatalf("expected index 'foo' in output:\n%s", string(out))
 	}
 }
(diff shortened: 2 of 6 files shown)


cmd/krew/cmd/index.go
```diff
@@ -54,9 +55,31 @@ each configured index in table format.`,
...
 
+var indexAddCmd = &cobra.Command{
+	Use:   "add",
+	Short: "Add a new index",
+	Long: `Configure a new index to install plugins from.
+
+Example:
+  kubectl krew index add index-name index-url`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) != 2 {
```
ahmetb: Cobra supports this natively. You can set minargs 2

cmd/krew/cmd/index.go
```diff
@@ -54,9 +55,31 @@ each configured index in table format.`,
...
+			return cmd.Help()
+		}
+
+		err := indexoperations.AddIndex(paths.IndexBase(), args[0], args[1])
+		if err != nil {
+			return errors.Wrapf(err, "failed to add index %s: %s", args[0], args[1])
```
ahmetb: No need to print args back again

cmd/krew/cmd/index.go
```diff
@@ -54,9 +55,31 @@ each configured index in table format.`,
 	},
 }
 
+var indexAddCmd = &cobra.Command{
+	Use:   "add",
+	Short: "Add a new index",
+	Long: `Configure a new index to install plugins from.
+
+Example:
+  kubectl krew index add index-name index-url`,
```
ahmetb: This isnt really an example, but a placeholder. 

Lets use default index as example

cmd/krew/cmd/index.go
```diff
@@ -54,9 +55,31 @@ each configured index in table format.`,
 	},
 }
 
+var indexAddCmd = &cobra.Command{
+	Use:   "add",
+	Short: "Add a new index",
+	Long: `Configure a new index to install plugins from.
+
+Example:
```
ahmetb: Theres an example field on cobra.

cmd/krew/cmd/index.go
```diff
@@ -54,9 +55,31 @@ each configured index in table format.`,
 	},
 }
 
+var indexAddCmd = &cobra.Command{
+	Use:   "add",
+	Short: "Add a new index",
+	Long: `Configure a new index to install plugins from.
```
ahmetb: See if you can name the positional args. 
Like NAME GIT_URL. 
Cobra probably has facilities to do that.

internal/index/indexoperations/index.go
```diff
@@ -52,3 +53,14 @@ func ListIndexes(path string) ([]Index, error) {
 	}
 	return indexes, nil
 }
+
+// AddIndex initializes a new index to install plugins from.
+func AddIndex(path, name, url string) error {
+	dir := filepath.Join(path, name)
```
ahmetb: @corneliusweig wdyt of a preliminary check to see if url is an actual url? Simple url.Parse can help.

internal/index/indexoperations/index.go
```diff
@@ -52,3 +53,14 @@ func ListIndexes(path string) ([]Index, error) {
...
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
+		return gitutil.EnsureCloned(url, dir)
+	} else if err != nil {
+		return errors.Wrapf(err, "failed to describe %s", dir)
+	}
+	return errors.New("index already exists")
```
ahmetb: Handle negative case first. That way you can unindent thr huge block above

internal/index/indexoperations/index_test.go
```diff
@@ -64,3 +54,65 @@ func TestListIndexes(t *testing.T) {
...
+			}
+
+			gotErr := AddIndex(tmpDir.Path("index"), tt.args.index.Name, tt.args.index.URL)
+			if (gotErr != nil) != tt.wantErr {
+				t.Errorf("AddIndex(), error = %v, wantErr = %v", gotErr, tt.wantErr)
+			}
+		})
+	}
+}
+
```
ahmetb: Mark as t.Helper()

internal/index/indexoperations/index_test.go
```diff
@@ -64,3 +54,65 @@ func TestListIndexes(t *testing.T) {
...
+func initEmptyGitRepo(t *testing.T, path, url string) {
```
ahmetb: Mark as t.Helper()

internal/index/indexoperations/index_test.go
```diff
@@ -64,3 +54,65 @@ func TestListIndexes(t *testing.T) {
...
+	if err := os.MkdirAll(path, os.ModePerm); err != nil {
+		t.Fatalf("cannot create directory %q: %s", filepath.Dir(path), err)
+	}
```
ahmetb: Merge err declaration and if, ditto below

internal/index/indexoperations/index_test.go
```diff
@@ -64,3 +54,65 @@ func TestListIndexes(t *testing.T) {
 		t.Errorf("output does not match: %s", diff)
 	}
 }
+
+func TestAddIndex(t *testing.T) {
+	type args struct {
```
ahmetb: Unwrap

internal/index/indexoperations/index_test.go
```diff
@@ -64,3 +54,65 @@ func TestListIndexes(t *testing.T) {
...
+			name: "preexisting index",
+			args: args{
+				index: Index{
+					Name: "foo",
+					URL:  "https://github.com/valid/index.git",
+				},
+			},
+			wantErr: true,
+		},
+	}
```
ahmetb: Add success case

ahmetb:
Please add CLI integration tests to ./integration_test.

internal/index/indexoperations/index_test.go
chriskim06: Is it ok for it to clone a repo during the unit tests? I had a success case initially using krew-index but didn't like that it required a network connection.

ahmetb: You can clone local dirs as repos IIRC

ahmetb: https://stackoverflow.com/questions/21045061/git-clone-from-another-directory

ahmetb: Maybe you can make success test another unit test method. Tabular tests are often overkill for 1-2 cases.

internal/index/indexoperations/index.go
corneliusweig: I thought this is not desirable, because it might break magic corp URLs with non-standard protocol prefixes. Also, somebody pointed out that `url.Parse` is not foolproof, although I don't remember exactly why.

IMHO, `EnsureCloned` should surface a readable error message? Maybe we also want to delete the index directory if cloning failed. Otherwise users will have to manually delete the index directory.

internal/index/indexoperations/index_test.go
```diff
@@ -64,3 +54,65 @@ func TestListIndexes(t *testing.T) {
...
+	tests := []struct {
+		name    string
+		args    args
+		wantErr bool
+	}{
+		{
+			name: "invalid git repository",
+			args: args{
+				index: Index{
+					Name: "foo",
```
corneliusweig: Is that even a valid case? The cobra command already ensures, that both index name and index URL are given. So we should not bother about calling this function with an empty URL, right?

internal/index/indexoperations/index.go
```diff
@@ -52,3 +53,14 @@ func ListIndexes(path string) ([]Index, error) {
 	}
 	return indexes, nil
 }
+
+// AddIndex initializes a new index to install plugins from.
+func AddIndex(path, name, url string) error {
+	dir := filepath.Join(path, name)
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
+		return gitutil.EnsureCloned(url, dir)
```
corneliusweig: If cloning fails, you should definitely clean up the directory. Otherwise then index registry is left in an inconsistent state and the user has to clean up manually.

internal/index/indexoperations/index_test.go
chriskim06: You're right this is kind of confusing, I wanted to add a case where the provided url is incorrect. In this case the test doesn't initialize an empty repo since the url is blank. I'll change this.

internal/index/indexoperations/index.go
chriskim06: Actually, wouldn't the directory not exist if cloning fails?

corneliusweig: You're right. Git is already taking care of this. My bad.

internal/index/indexoperations/index_test.go
corneliusweig: How do you tell whether a URL is correct / incorrect?

internal/index/indexoperations/index.go
corneliusweig: If bad URLs are handled by `git`, maybe we need to be more careful that the user actually sees an error. Can you make sure of that?

internal/index/indexoperations/index_test.go
chriskim06: I'm using the `localRepo` variable to initialize an empty git repo that gets used as the url in clone. In the success case it initializes a git repo at `local/foo` and then clones that to `index/foo`, one of the failure cases initializes the repo at `index/foo` already so that it tests the case where an index with a name already exists.

I'm pretty much using the existence of the local git repo as a correct url.

corneliusweig: I see, good idea. So then I recommend to separate the two test cases into two test functions. That makes the tests so much more readable!

internal/index/indexoperations/index.go
chriskim06: The git clone error does ultimately get wrapped and surfaced to the user. Is that enough or did you mean to add some other context around the git error?

corneliusweig: Ok, that is what we need. This should be good enough.

integration_test/index_test.go
```diff
@@ -0,0 +1,47 @@
...
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	initialList := test.WithMigratedIndex().WithEnv(constants.EnableMultiIndexSwitch, 1).Krew("index", "list").RunOrFailOutput()
+	if !bytes.Contains(initialList, []byte(constants.IndexURI)) {
+		t.Fatalf("expected krew-index in output:\n%s", string(initialList))
+	}
+
+	indexName := "foo"
+	test.Krew("index", "add", indexName, constants.IndexURI).RunOrFail()
```
ahmetb: Add/list should have separate tests ideally. 

But for this you can merge them but make sure test name reflects that. Refer to other integration tests for an idea on how to develop. We test cli scenario and behavior here. 

For example you don’t test 0 args fails case here :)

integration_test/index_test.go
```diff
@@ -0,0 +1,47 @@
...
+	if _, err := os.Stat(filepath.Join(test.Root(), "index", indexName, ".git")); err != nil {
+		t.Fatalf("error adding index: %s", err)
+	}
+
```
ahmetb: Need test case: what happens if index list is empty.

integration_test/testutil_test.go
```diff
@@ -171,6 +173,13 @@ func (it *ITest) WithIndex() *ITest {
 	return it
 }
 
+// WithMigratedIndex initializes the index and performs the index migration.
```
ahmetb: Add TODO for deletion, ditto migrateIndex.

Why don’t we use the env variable for behavior change in initializeIndex instead? Simple if can solve this but you added 2 methods for it. :)

chriskim06: That's a good idea, I'll change it

ahmetb:
@chriskim06 does the code handle this case?

    krew index add ../../../usr/bin [REPO]

we need sanity checks around names to ensure they don't serve path escapes.

chriskim06:
I didn't think about that, I'll add those checks.

cmd/krew/cmd/index.go
```diff
@@ -54,9 +55,26 @@ each configured index in table format.`,
...
+	Long:    "Configure a new index to install plugins from.",
+	Example: "kubectl krew index add default " + constants.IndexURI,
+	Args:    cobra.ExactArgs(2),
+	RunE: func(_ *cobra.Command, args []string) error {
+		err := indexoperations.AddIndex(paths.IndexBase(), args[0], args[1])
+		if err != nil {
+			return errors.Wrap(err, "failed to add index")
+		}
+		klog.Infoln("Successfully added index")
+		return nil
```
ahmetb: honestly this entire func can be 1 line :)
we'd know the success from the command execution anyway, so no need to log that.

integration_test/index_test.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func TestKrewIndexList(t *testing.T) {
+	skipShort(t)
+
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	test.WithEnv(constants.EnableMultiIndexSwitch, 1).WithIndex()
+	out := test.Krew("index", "list").RunOrFailOutput()
```
ahmetb: if you add string() around this output your life might get easier below :)

integration_test/index_test.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func TestKrewIndexList_NoIndexes(t *testing.T) {
+	skipShort(t)
+
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	test.WithEnv(constants.EnableMultiIndexSwitch, 1).WithIndex()
+	index := test.TempDir().Path("index/" + constants.DefaultIndexName)
```
ahmetb: you can just remove `index` completely. that should still return empty list, right?

internal/index/indexoperations/index.go
```diff
@@ -52,3 +54,18 @@ func ListIndexes(path string) ([]Index, error) {
 	}
 	return indexes, nil
 }
+
+// AddIndex initializes a new index to install plugins from.
+func AddIndex(path, name, url string) error {
+	if strings.ContainsAny(name, "./") {
```
ahmetb: as you might imagine this won't work on Windows

please use os.PathSeparator

ahmetb: consider extracting this to `IsValidIndexName` method and check for further things, like spaces, `\` `/`, `.` ... + add unit tests + reuse everywhere (add, delete, ...)

internal/index/indexoperations/index.go
```diff
@@ -52,3 +54,18 @@ func ListIndexes(path string) ([]Index, error) {
...
+		return errors.New("index name cannot contain path characters")
+	}
+
+	dir := filepath.Join(path, name)
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
+		return gitutil.EnsureCloned(url, dir)
+	} else if err != nil {
+		return errors.Wrapf(err, "failed to describe %s", dir)
```
ahmetb: - `describe` &rarr; `read directory`

- %s is not necessary, as file errors always contain the path anyway.

cmd/krew/cmd/index.go
chriskim06: I thought that it seemed a little nicer with some explicit validation (similar to index/receipt migration) but am fine either way. I'll remove it.

integration_test/index_test.go
chriskim06: List uses `paths.IndexBase()` right now so it returns an error when it tries to `ReadDir()` it to see what indexes are there.

ahmetb: Hmm I'm not sure it should fail in case of os.IsNotExist...

What if user has never run krew (so ~/.krew doesn't even exist). IMO `krew index list` should not fail. @corneliusweig right?

integration_test/index_test.go
```diff
@@ -0,0 +1,81 @@
...
+	test.WithEnv(constants.EnableMultiIndexSwitch, 1).WithIndex()
+	if err := test.Krew("index", "add").Run(); err == nil {
+		t.Fatal("expected index add with no args to fail")
+	}
+	if err := test.Krew("index", "add", "foo", "https://invalid").Run(); err == nil {
+		t.Fatal("expected index add with invalid URL to fail")
+	}
+	if err := test.Krew("index", "add", "../../usr/bin", constants.IndexURI); err == nil {
+		t.Fatal("expected index add with path characters to fail")
+	}
```
corneliusweig: This case is also the case of an invalid  URL. File paths are perfectly valid repo locations for git and it will clone happily if the source exists (it's dangerous though, because both repos will share their blob store).

So I'd recommend to remove this case.

integration_test/index_test.go
```diff
@@ -0,0 +1,81 @@
...
+	if err := test.Krew("index", "add", "foo", constants.IndexURI).Run(); err != nil {
+		t.Fatalf("error adding new index: %v", err)
+	}
```
corneliusweig: You can avoid cloning from remote which makes the test a bit slow. Instead, you can clone from the default index by path.

Since several tests do that, this would really make sense.

If you feel that you need a clone via network, then I suggest to just leave one test for that.

integration_test/index_test.go
corneliusweig: Yeah, this is a subtle bug... The problem is that the index is only initialized for _some_ commands (install, update, upgrade), see `ensureIndexUpdated`. So if the index is removed, our index migration check thinks it is already migrated and does nothing. This is not some construed case, because it is prefectly reasonable to run `krew index list` right after install. So we need to fix this.

And when it's done, it should be enough to remove `index` as suggested.

integration_test/index_test.go
```diff
@@ -0,0 +1,81 @@
...
+	defer cleanup()
+
+	test.WithEnv(constants.EnableMultiIndexSwitch, 1).WithIndex()
+	index := test.TempDir().Path("index/" + constants.DefaultIndexName)
+	if err := os.RemoveAll(index); err != nil {
+		t.Fatalf("error removing default index: %v", err)
+	}
+
+	out := test.Krew("index", "list").RunOrFailOutput()
+	if !bytes.Equal(out, []byte("INDEX  URL\n")) {
```
corneliusweig: It would be nicer to get a message saying `No index configured, run krew index add default https://github.com/kubernetes-sigs/krew-index to get started`. Or something similar.

integration_test/testutil_test.go
```diff
@@ -283,6 +285,14 @@ func (it *ITest) initializeIndex() {
 	if err := cmd.Run(); err != nil {
 		it.t.Fatalf("cannot restore index from cache: %s", err)
 	}
+
+	for _, e := range it.env {
+		if strings.Contains(e, constants.EnableMultiIndexSwitch) {
```
corneliusweig: Can you leave a todo comment as a reminder to simplify this setup when krew has fully migrated to the new layout?

internal/index/indexoperations/index.go
corneliusweig: Even easier, we could simply restrict index names to k8s resource names (see https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names). Hopefully, there is a k8s tool that we can just use?

internal/index/indexoperations/index.go
corneliusweig: I assume that the go error already has a sufficient human-readable description of the error. So maybe wrapping is just unnecessary here.

internal/index/indexoperations/index_test.go
```diff
@@ -64,3 +54,49 @@ func TestListIndexes(t *testing.T) {
...
+	tmpDir, cleanup := testutil.NewTempDir(t)
+	defer cleanup()
+
+	indexName := "foo"
+	localRepo := tmpDir.Path("local/" + indexName)
+	initEmptyGitRepo(t, localRepo, "")
+
+	if err := AddIndex(tmpDir.Path("index"), indexName, localRepo); err != nil {
+		t.Errorf("error adding index: %v", err)
+	}
```
corneliusweig: IMHO, you should also test that `AddIndex` has actually done something.

integration_test/index_test.go
chriskim06: I added this one for testing if the name of the index contains those path characters. It fails at a different point than when the provided URL is invalid. The error message is probably not worded super well since its pretty similar to the one with the invalid URL, I can update that to mention that the path characters are invalid in the name.

corneliusweig: Oh.. you're right, my bad. So what about extracting the name sanitizer logic into  its own function and test that instead? In the best case, we can reuse a k8s util anyway and we don't need to test anything.

chriskim06: Ahmet commented the same and I have some changes which I have yet to push that moves the logic to its own function. I want to address all the comments before pushing the changes, so I'll look into any existing k8s name validation util. I'll also remove this test since we'll have that logic elsewhere.

integration_test/index_test.go
chriskim06: I wasn't sure if this would be ok or not. Since these tests were for testing cli usage I figured it would be good to add an index via network, but if you guys feel its unnecessary I will remove it :+1:

internal/index/indexoperations/index.go
chriskim06: Like `IsValidPathSegmentName` https://github.com/kubernetes/apimachinery/blob/master/pkg/api/validation/path/name.go?

ahmetb: Let's start restrictive, we can relax as we need.
Honestly, something like `[A-Za-z0-9_-]` is good enough and safest for now. Let's make it a method regardless as I'm intending to reuse in other places.

integration_test/index_test.go
ahmetb: We should still test an invalid path case in integration_test, since it's CLI behavior. I'm not sure if `indexoperations` methods should be aware of security checks.

integration_test/index_test.go
chriskim06: So if a user has 0.3.4 and has never run krew they won't have access to the index command until they update krew. If they update through krew they'll have old index directory and next command will prompt to migrate (or automatically migrate if/when we change to that).

Fresh install of newer version means the user is already technically "migrated" even without the index directory being initialized right? The default krew-index should be initialized in the new structure, or a different "default" can be added through `krew index add`.

Do those sound right to you guys or is there a scenario I'm missing? As to `index list` not failing, I think its just a matter of not treating `os.IsNotExist` as a failure like Ahmet mentioned.

ahmetb: > Fresh install of newer version means the user is already technically "migrated" even without the index directory being initialized right? The default krew-index should be initialized in the new structure, or a different "default" can be added through `krew index add`.

I am hoping we solve this in a different PR.

For this PR: when Krew runs, it will always go ahead and create some dirs in ~/.krew. https://github.com/kubernetes-sigs/krew/blob/5c18854cb38e5a258bb73015391eed49a6c99932/cmd/krew/cmd/root.go#L110-L114 Right now it doesn't have `index/` dir, so let's go ahead and add it there.

That way, any time krew runs (even for --help or a failure) the `index/` will exist. That way, we can detect no index exists (ls index/ == 0 items) and initialize index.

internal/index/indexoperations/index_test.go
ahmetb: To do that, you can have a combined Add+List scenario.

ahmetb: That way you can also move away from doing manual mkdir/clone in the test code.

internal/index/indexoperations/index.go
corneliusweig: Ah, finally found it: https://github.com/kubernetes/kubernetes/blob/fdb2cb4c8832da1499069bda918c014762d8ac05/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L203

If we can, let's use it. Then we get all the goodness with names starting with alpha characters.

ahmetb: Are you 100% sure that string supports all repo names on github. If so, let's use it. If not, let's stick to something more relevant.

cmd/krew/cmd/index.go
```diff
@@ -47,6 +47,12 @@ each configured index in table format.`,
 		if err != nil {
 			return errors.Wrap(err, "failed to list indexes")
 		}
+
+		if len(indexes) == 0 {
+			klog.Info(indexoperations.NoIndexMessage)
+			return nil
+		}
```
ahmetb: let's not do this here.

We can do that when user tries to query a plugin (`krew install`, `krew update`, `krew upgrade` etc) but there's no plugins installed.

Furthermore klog isn't good for this. we probably should use fmt.Fprintf or something like that with color pkg.

integration_test/index_test.go
corneliusweig: I guess, creating `index/` also fixes the bug entirely.

integration_test/index_test.go
```diff
@@ -38,7 +39,7 @@ func TestKrewIndexAdd(t *testing.T) {
 	if err := test.Krew("index", "add", "../../usr/bin", constants.IndexURI); err == nil {
 		t.Fatal("expected index add with path characters to fail")
 	}
-	if err := test.Krew("index", "add", "foo", constants.IndexURI).Run(); err != nil {
+	if err := test.Krew("index", "add", "foo", test.TempDir().Path("index/"+constants.DefaultIndexName)).Run(); err != nil {
 		t.Fatalf("error adding new index: %v", err)
```
ahmetb: do you mind explaining why is this a failure?

integration_test/index_test.go
```diff
@@ -38,7 +39,7 @@ func TestKrewIndexAdd(t *testing.T) {
...
 	}
```
ahmetb: can you add repeat this command but test "re-adding existing index must fail"

cmd/krew/cmd/index.go
ahmetb: Also I think this command should still print an "empty table".

We need to print the output properly to stderr so it doesn't get mixed up with other stuff.

integration_test/index_test.go
```diff
@@ -69,13 +70,13 @@ func TestKrewIndexList_NoIndexes(t *testing.T) {
...
+	index := test.TempDir().Path("index")
 	if err := os.RemoveAll(index); err != nil {
 		t.Fatalf("error removing default index: %v", err)
 	}
 
 	out := test.Krew("index", "list").RunOrFailOutput()
-	if !bytes.Equal(out, []byte("INDEX  URL\n")) {
+	if !strings.Contains(string(out), indexoperations.NoIndexMessage) {
 		t.Fatalf("expected index list to be empty:\n%s", string(out))
 	}
```
ahmetb: testing with messages isn't good practice, let's not do that.
the previous test was good.

internal/index/indexoperations/index.go
```diff
@@ -18,11 +18,17 @@ import (
...
+	"regexp"
 
 	"github.com/pkg/errors"
 
 	"sigs.k8s.io/krew/internal/gitutil"
+	"sigs.k8s.io/krew/pkg/constants"
+)
+
+var (
+	NoIndexMessage   = "No index configured, run kubectl krew index add default " + constants.IndexURI + " to get started"
```
ahmetb: Let's not do this here. We should ideally implement this in cmd/krew pkg as a `func PrintDefaultIndexSuggestion()`.

Index commands are for pro-users, so most users should be realizing they miss an index while using main user cmds (install, search, update etc). "index" cmds are mgmt cmds, so they ideally shouldn't print these.

integration_test/index_test.go
corneliusweig:
```suggestion
	out := string(test.Krew("index", "list").RunOrFailOutput())
```

internal/index/indexoperations/index_test.go
```diff
@@ -64,3 +54,74 @@ func TestListIndexes(t *testing.T) {
...
+	if IsValidIndexName("foo\\bar") {
+		t.Error("name cannot contain back slashes")
+	}
+	if IsValidIndexName("foo.bar") {
+		t.Error("name cannot contain forward periods")
+	}
+	if !IsValidIndexName("foo") {
+		t.Error("expected valid index name: foo")
+	}
```
ahmetb: this actually would be the perfect use case for tabular tests :)

internal/index/indexoperations/index.go
```diff
@@ -33,7 +41,9 @@ type Index struct {
 // the base path of the index.
 func ListIndexes(path string) ([]Index, error) {
 	dirs, err := ioutil.ReadDir(path)
-	if err != nil {
+	if os.IsNotExist(err) {
+		return nil, nil
```
corneliusweig: nit: now that we ensure that `index/` is present on startup, this error case no longer needs to be treated in a special way.

internal/index/indexoperations/index_test.go
```diff
@@ -64,3 +54,74 @@ func TestListIndexes(t *testing.T) {
...
+
+	localRepo := tmpDir.Path("local/" + indexName)
+	initEmptyGitRepo(t, tmpDir.Path("index/"+indexName), "")
+	initEmptyGitRepo(t, localRepo, "")
+
+	if err := AddIndex(indexRoot, indexName, localRepo); err == nil {
+		t.Error("expected error when adding an index that already exists")
+	}
+
+	if err := AddIndex(indexRoot, "foo/bar", ""); err == nil {
```
corneliusweig: This invalid index name case was already above. You should co-locate the two cases, so that one has better overview over what's tested.

integration_test/index_test.go
chriskim06: Ah that's a good idea.

integration_test/index_test.go
chriskim06: Ok I'll change it back

cmd/krew/cmd/index.go
chriskim06: Do you mean it should still print:
```
INDEX  URL
```

internal/index/indexoperations/index.go
chriskim06: I'll just remove this for now then since I'm not covering the main user cmds in this PR.

cmd/krew/cmd/index.go
ahmetb: Yeah. That’s expected. Like krew list.

internal/index/indexoperations/index.go
ahmetb: Yeah sgtm. We’ll handle the scenario later.

integration_test/index_test.go
ahmetb: I think I also mentioned this. Better to add before we merge.

integration_test/index_test.go
ahmetb: Lets do this elsewhere (in main user cmds) or later.

integration_test/index_test.go
chriskim06: I had this before and changed the comparisons to `strings.Contains` but you mentioned that testing with messages wasn't good practice and that the previous tests were good. I probably missed something, could you elaborate a little more?

ahmetb: matching output format (regexp, format, line count) is one thing, matching warning msg is another. See TestKrewList, it’s not pretty but we measure line counts etc.

internal/index/indexoperations/index.go
corneliusweig: Looks like GitHub also allows dots in repo names, i.e. `[A-Za-z0-9_-.]+`. And it does not care whether a repo name starts with an alpha character or `_-.`. So if we want to support everything that GitHub does, the current pattern is better.

internal/index/indexoperations/index_test.go
chriskim06: Sure I can change it to use that.

internal/index/indexoperations/index.go
```diff
@@ -52,3 +56,23 @@ func ListIndexes(path string) ([]Index, error) {
 	}
 	return indexes, nil
 }
+
+// AddIndex initializes a new index to install plugins from.
+func AddIndex(path, name, url string) error {
+	if !IsValidIndexName(name) {
+		return errors.New("index name cannot contain path characters")
```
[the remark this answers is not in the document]
chriskim06: What about something like "index name can only contain alphanumeric characters, -, and _"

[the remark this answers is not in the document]
ahmetb: This should just say “invalid index name”. 
In a separate PR you can return an error from the validation func and wrap it here :)

ahmetb:
/lgtm
/approve

Thanks @chriskim06 for pushing it through.
Thanks @corneliusweig for thorough reviews!

I'll be making several followup PRs to clean up a few things! 
I also have a patch ready for `krew index delete`.

PR #547, opened 2020-03-11
Add custom usage template for cobra command so that usage will display 'kubectl krew' instead of just 'krew'

### LINKED ISSUE #521: Inconsistent program name in --help message
See the end, it doesn't say `kubectl krew`

```
$ kubectl krew
[...]
Flags:
  -h, --help      help for krew
  -v, --v Level   number for the log level verbosity

Use "krew [command] --help" for more information about a command.
```

brianpursley:
Fixes #521 

Custom usage template lets us add `kubectl` in front of the usages

---
**Example output of default usage message with this PR:**
```
out/bin/krew-linux_amd64 
```
```
krew is the kubectl plugin manager.
You can invoke krew through kubectl: "kubectl krew [command]..."

Usage:
  kubectl krew [command]

Available Commands:
  help        Help about any command
  info        Show information about a kubectl plugin
  install     Install kubectl plugins
  list        List installed kubectl plugins
  search      Discover kubectl plugins
  uninstall   Uninstall plugins
  update      Update the local copy of the plugin index
  upgrade     Upgrade installed plugins to newer versions
  version     Show krew version and diagnostics

Flags:
  -h, --help      help for krew
  -v, --v Level   number for the log level verbosity

Use "kubectl krew [command] --help" for more information about a command.
```

---
**Example output of a command help with this PR in place:**
```
out/bin/krew-linux_amd64 list --help
```
```
Show a list of installed kubectl plugins and their versions.

Remarks:
  Redirecting the output of this command to a program or file will only print
  the names of the plugins installed. This output can be piped back to the
  "install" command.

Usage:
  kubectl krew list [flags]

Flags:
  -h, --help   help for list

Global Flags:
  -v, --v Level   number for the log level verbosity
```


brianpursley:
By the way, I waffled on whether to just bring in the entire template from cobra and maintain it in krew or to use string.Replace().  I ended up with string.Replace() because it was less code and seemed a little more targeted.  Let me know if you have a preference.

Here is their template:
```
Usage:{{if .Runnable}}
  {{.UseLine}}{{end}}{{if .HasAvailableSubCommands}}
  {{.CommandPath}} [command]{{end}}{{if gt (len .Aliases) 0}}

Aliases:
  {{.NameAndAliases}}{{end}}{{if .HasExample}}

Examples:
{{.Example}}{{end}}{{if .HasAvailableSubCommands}}

Available Commands:{{range .Commands}}{{if (or .IsAvailableCommand (eq .Name "help"))}}
  {{rpad .Name .NamePadding }} {{.Short}}{{end}}{{end}}{{end}}{{if .HasAvailableLocalFlags}}

Flags:
{{.LocalFlags.FlagUsages | trimTrailingWhitespaces}}{{end}}{{if .HasAvailableInheritedFlags}}

Global Flags:
{{.InheritedFlags.FlagUsages | trimTrailingWhitespaces}}{{end}}{{if .HasHelpSubCommands}}

Additional help topics:{{range .Commands}}{{if .IsAdditionalHelpTopicCommand}}
  {{rpad .CommandPath .CommandPathPadding}} {{.Short}}{{end}}{{end}}{{end}}{{if .HasAvailableSubCommands}}

Use "{{.CommandPath}} [command] --help" for more information about a command.{{end}}
```

ahmetb:
> I waffled on whether to just bring in the entire template from cobra and maintain it in krew or to use string.Replace()

Honestly, my only concern is that this value isn't a const in cobra, and I'm not sure if they treat the template as "API" that they shouldn't be breaking.

But this works for me.

Worth adding a note "why" we're doing this in the code (with links to relevant issues/PRs).

brianpursley:
Added comment with link to issue

brianpursley:
Can I get ok-to-test?

ahmetb:
/ok-to-test
/retest

That test failure looks like a flake. We should look into it if reproes. Feel free to open an issue with the stage’s output.

ahmetb:
Can we add a note to the `	Use:   "krew",` saying this is prefixed by kubectl in the template?

ahmetb:
Not sure why the tests's failing, do you mind quickly `git fetch origin/master && git rebase origin/master && git push -f`.

cmd/krew/cmd/root.go
```diff
@@ -98,6 +99,14 @@ func init() {
 	}
 
 	paths = environment.MustGetKrewPaths()
+
+	// Cobra doesn't have a way to specify a two word command (ie. "kubectl krew"), so set a custom usage template
+	// with kubectl in it. Cobra will use this template for the root and all child commands.
+	// See issue #521 for more details on the reason for this: https://github.com/kubernetes-sigs/krew/issues/521
```
corneliusweig: This last comment line is more apt for the commit message, IMHO. Can you move it there?

cmd/krew/cmd/root.go
```diff
@@ -98,6 +99,14 @@ func init() {
...
+	customUsageTemplate := rootCmd.UsageTemplate()
+	customUsageTemplate = strings.ReplaceAll(customUsageTemplate, "{{.UseLine}}", "kubectl {{.UseLine}}")
+	customUsageTemplate = strings.ReplaceAll(customUsageTemplate, "{{.CommandPath}}", "kubectl {{.CommandPath}}")
```
corneliusweig: This looks like a use-case for `strings.NewReplacer(args ...string)`.

ahmetb:
> One more idea: if it breaks, we won't know unless somebody happens to notice it. So could you modify one of our integration tests to look out for `kubectl krew` in the command output?

Argh not sure if that's gonna be a clean test. :| :| `krew --help` would have at least 3 different occurrences of `kubectl krew` in different places.

corneliusweig:
> would have at least 3 different occurrences of kubectl krew in different places.

Isn't it good enough to see if the substring `kubectl krew` is there at all?

cmd/krew/cmd/root.go
brianpursley: Hmm. Good call.  I'll take a look at that

brianpursley:
I will see if I can add some kind of test like you said.  I think it is a good idea.

cmd/krew/cmd/root.go
brianpursley: Oh yeah.  I thought you meant to add it in the code comment, I misunderstood.  Definitely can move it to the commit message instead

brianpursley:
@ahmetb @corneliusweig I made the proposed changes. Can you take another look.

(Also, it looks like the test that was failing before has passed, so maybe the rebase fixed it)

ahmetb:
> Isn't it good enough to see if the substring `kubectl krew` is there at all?

nope. There's already a `kubectl krew` that comes from the USAGE string.

cmd/krew/cmd/root_test.go
```diff
@@ -0,0 +1,43 @@
...
+		if !strings.Contains(usageTemplate, expectedString) {
+			t.Errorf("expected usage template to contain '%s' but it did not:\n%v", expectedString, usageTemplate)
+		}
+	}
+}
+
+func TestUsageStringContainsKubectlKrew(t *testing.T) {
+	usage := rootCmd.UsageString()
+
+	if !strings.Contains(usage, "kubectl krew") {
```
ahmetb: this isn't good. Example already has `kubectl krew`. You need to at least count all `krew` usages have `kubectl` next to them or something, and even that's not great either.

let's just scrap this test. it wasn't agreed upon fully.

brianpursley: Ok.  It's probably enough to check the template anyway.  We don't need to test cobras ability to render it :)

corneliusweig:
/lgtm
/approve
Thanks for fixing this!

.gitignore
```diff
@@ -11,6 +11,10 @@
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
 
+# Files generated by JetBrains IDEs, e.g. IntelliJ IDEA
+.idea/
+*.iml
```
ahmetb, line 16: @brianpursley I just noticed this.

Ideally we don't wanna have editor-specific stuff here.

You should add these to your ~/.gitignore_global file and set git global config `core.excludesfile=~/.gitignore_global`

brianpursley: Sorry about that.  I will make a PR to clean it up :)

PR #548, opened 2020-03-12
ListInstalledPlugins use Receipts for parsing

ahmetb:
While parsing into a Plugin works here, what we really parse is Receipts,
so use that.

/assign @corneliusweig
/cc @chriskim06


corneliusweig:
/lgtm
/approve
/hold
@chriskim06 you can remove the hold unless you have comments.

ahmetb:
/hold cancel
chris might not be able to un-hold but I sent as FYI.

PR #552, opened 2020-03-17
cmd: Add 'index remove' command

### LINKED ISSUE #550: indexoperations methods should take environment.Paths
While trying to refactor `indexoperations` in #549 and adding a `DeleteIndex` method I realized these methods should take `environment.Paths` value and use its method to construct paths to index dirs etc.

That's precisely why we created `Paths` and its methods: **To avoid spread and duplication of** FS layout of things and repetition of filepath.Joins all over the codebase. :) 

/assign

I'll get to this once #549 is merged.
cc: @chriskim06

ahmetb:
Fixes #550
Related issue: #545, #483

### DIFF
--- a/cmd/krew/cmd/index.go
+++ b/cmd/krew/cmd/index.go
@@ -98,7 +98,7 @@ func indexDelete(_ *cobra.Command, args []string) error {
 			names = append(names, v.Name)
 		}
 
-		internal.PrintWarning(os.Stderr, `%s Plugins [%s] are still installed from index %q.
+		internal.PrintWarning(os.Stderr, `Plugins [%s] are still installed from index %q.
 			"Removing indexes while there are plugins installed from them may result in
 			"unsupported behavior (use --force to ignore this check).`, strings.Join(names, ", "), name)
 		return errors.Errorf("there are still plugins installed from this index")
--- a/cmd/krew/cmd/index.go
+++ b/cmd/krew/cmd/index.go
@@ -104,7 +104,7 @@ func indexDelete(_ *cobra.Command, args []string) error {
 
 		internal.PrintWarning(os.Stderr, `Plugins [%s] are still installed from index %q!
 Removing indexes while there are plugins installed from is not recommended
-(you can use --force to ignore this check).`, strings.Join(names, ", "), name)
+(you can use --force to ignore this check).`+"\n", strings.Join(names, ", "), name)
 		return errors.Errorf("there are still plugins installed from this index")
 	}
 
--- a/cmd/krew/cmd/index.go
+++ b/cmd/krew/cmd/index.go
@@ -23,7 +23,6 @@ import (
 	"github.com/spf13/cobra"
 	"k8s.io/klog"
 
-	"sigs.k8s.io/krew/internal/environment"
 	"sigs.k8s.io/krew/internal/index/indexoperations"
 	"sigs.k8s.io/krew/internal/installation"
 	"sigs.k8s.io/krew/pkg/constants"
@@ -87,7 +86,7 @@ installed from them is not supported behavior.`,
 func indexDelete(_ *cobra.Command, args []string) error {
 	name := args[0]
 
-	pl, err := installation.InstalledPluginsFromIndex(environment.MustGetKrewPaths().InstallReceiptsPath(), name)
+	pl, err := installation.InstalledPluginsFromIndex(paths.InstallReceiptsPath(), name)
 	if err != nil {
 		return errors.Wrap(err, "failed querying plugins installed from the index")
 	}
@@ -105,7 +104,7 @@ func indexDelete(_ *cobra.Command, args []string) error {
 		return errors.Errorf("refusing to remove due to installed plugins from this index")
 	}
 
-	err = indexoperations.DeleteIndex(environment.MustGetKrewPaths(), name)
+	err = indexoperations.DeleteIndex(paths, name)
 	if os.IsNotExist(err) {
 		if *forceIndexDelete {
 			klog.V(4).Infof("Index not found, but --force is used, so not returning an error")
--- a/cmd/krew/cmd/index.go
+++ b/cmd/krew/cmd/index.go
@@ -16,11 +16,16 @@ package cmd
 
 import (
 	"os"
+	"strings"
 
+	"github.com/fatih/color"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"k8s.io/klog"
 
+	"sigs.k8s.io/krew/internal/environment"
 	"sigs.k8s.io/krew/internal/index/indexoperations"
+	"sigs.k8s.io/krew/internal/installation"
 	"sigs.k8s.io/krew/pkg/constants"
 )
 
@@ -42,7 +47,7 @@ This command prints a list of indexes. It shows the name and the remote URL for
 each configured index in table format.`,
 	Args: cobra.NoArgs,
 	RunE: func(_ *cobra.Command, _ []string) error {
-		indexes, err := indexoperations.ListIndexes(paths.IndexBase())
+		indexes, err := indexoperations.ListIndexes(paths)
 		if err != nil {
 			return errors.Wrap(err, "failed to list indexes")
 		}
@@ -62,14 +67,67 @@ var indexAddCmd = &cobra.Command{
 	Example: "kubectl krew index add default " + constants.IndexURI,
 	Args:    cobra.ExactArgs(2),
 	RunE: func(_ *cobra.Command, args []string) error {
-		return indexoperations.AddIndex(paths.IndexBase(), args[0], args[1])
+		return indexoperations.AddIndex(paths, args[0], args[1])
 	},
 }
 
+var indexDeleteCmd = &cobra.Command{
+	Use:   "remove",
+	Short: "Remove a configured index",
+	Long: `Remove a configured index repository which is used to discover and
+install plugins from.
+
+If there are plugins installed from the specified index, the index cannot be
+removed (unless --force) is used. Removing indexes while there are plugins
+installed from them is not supported behavior.`,
+	Args: cobra.ExactArgs(1),
+	RunE: indexDelete,
+}
+
+func indexDelete(_ *cobra.Command, args []string) error {
+	name := args[0]
+
+	pl, err := installation.InstalledPluginsFromIndex(environment.MustGetKrewPaths().InstallReceiptsPath(), name)
+	if err != nil {
+		return errors.Wrap(err, "failed querying plugins installed from the index")
+	}
+	klog.V(4).Infof("Found %d plugins from index", len(pl))
+
+	if len(pl) > 0 && !*forceIndexDelete {
+		var names []string
+		for _, v := range pl {
+			names = append(names, v.Name)
+		}
+
+		warning := color.New(color.FgRed, color.Bold).Sprint("WARNING:")
+		klog.Warningf("%s Plugins [%s] are still installed from index %q.", warning, strings.Join(names, ", "), name)
+		klog.Warning("Uninstall them first before removing this index (or, use --force, which may result in unsupported behavior).")
+		return errors.Errorf("refusing to remove due to installed plugins from this index")
+	}
+
+	err = indexoperations.DeleteIndex(environment.MustGetKrewPaths(), name)
+	if os.IsNotExist(err) {
+		if *forceIndexDelete {
+			klog.V(4).Infof("Index not found, but --force is used, so not returning an error")
+			return nil // success if --force specified and index does not exist.
+		}
+		return errors.Errorf("index %q does not exist", name)
+	}
+	return errors.Wrap(err, "error while removing the plugin index")
+}
+
+var (
+	forceIndexDelete *bool
+)
+
 func init() {
+	forceIndexDelete = indexDeleteCmd.Flags().Bool("force", false, "Remove index even if it has plugins installed (may yield in unsupported behavior)")
+
+	indexCmd.AddCommand(indexAddCmd)
+	indexCmd.AddCommand(indexListCmd)
+	indexCmd.AddCommand(indexDeleteCmd)
+
 	if _, ok := os.LookupEnv(constants.EnableMultiIndexSwitch); ok {
-		indexCmd.AddCommand(indexListCmd)
-		indexCmd.AddCommand(indexAddCmd)
 		rootCmd.AddCommand(indexCmd)
 	}
 }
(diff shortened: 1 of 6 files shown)


ahmetb:
/assign @corneliusweig 
cc: @chriskim06 to review as well.

ahmetb:
Getting a rather odd test failure I can't repro locally currently. Will continue to investigate, but suspecting something’s off in the test execution environment.

> image

cmd/krew/cmd/index.go
```diff
@@ -62,14 +67,67 @@ var indexAddCmd = &cobra.Command{
...
+			names = append(names, v.Name)
+		}
+
+		warning := color.New(color.FgRed, color.Bold).Sprint("WARNING:")
+		klog.Warningf("%s Plugins [%s] are still installed from index %q.", warning, strings.Join(names, ", "), name)
+		klog.Warning("Uninstall them first before removing this index (or, use --force, which may result in unsupported behavior).")
+		return errors.Errorf("refusing to remove due to installed plugins from this index")
+	}
+
+	err = indexoperations.DeleteIndex(environment.MustGetKrewPaths(), name)
```
chriskim06: Same question about `paths` var

cmd/krew/cmd/index.go
```diff
@@ -62,11 +67,65 @@ var indexAddCmd = &cobra.Command{
...
+removed (unless --force) is used. Removing indexes while there are plugins
+installed from them is not supported behavior.`,
+	Args: cobra.ExactArgs(1),
+	RunE: indexDelete,
+}
+
+func indexDelete(_ *cobra.Command, args []string) error {
+	name := args[0]
+
+	pl, err := installation.InstalledPluginsFromIndex(environment.MustGetKrewPaths().InstallReceiptsPath(), name)
```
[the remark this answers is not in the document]
ahmetb: I wasnt aware. 😄

cmd/krew/cmd/index.go
```diff
@@ -62,11 +67,65 @@ var indexAddCmd = &cobra.Command{
...
 
+var indexDeleteCmd = &cobra.Command{
+	Use:   "remove",
+	Short: "Remove a configured index",
+	Long: `Remove a configured index repository which is used to discover and
+install plugins from.
+
+If there are plugins installed from the specified index, the index cannot be
+removed (unless --force) is used. Removing indexes while there are plugins
+installed from them is not supported behavior.`,
```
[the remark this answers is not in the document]
ahmetb: 👍🏼

cmd/krew/cmd/index.go
```diff
@@ -62,14 +67,67 @@ var indexAddCmd = &cobra.Command{
...
+	}
+	klog.V(4).Infof("Found %d plugins from index", len(pl))
+
+	if len(pl) > 0 && !*forceIndexDelete {
+		var names []string
+		for _, v := range pl {
+			names = append(names, v.Name)
+		}
+
+		warning := color.New(color.FgRed, color.Bold).Sprint("WARNING:")
```
chriskim06: Would this be worthwhile putting somewhere? There's a couple places where something like this exists `boldRed := color.New(color.FgRed, color.Bold).SprintFunc()` and is called `boldRed("WARNING")`. I'm not even sure where it would go and this is pretty minor, just thought I'd bring it up.

cmd/krew/cmd/index.go
```diff
@@ -62,14 +67,67 @@ var indexAddCmd = &cobra.Command{
...
+var (
+	forceIndexDelete *bool
+)
+
 func init() {
+	forceIndexDelete = indexDeleteCmd.Flags().Bool("force", false, "Remove index even if it has plugins installed (may yield in unsupported behavior)")
+
+	indexCmd.AddCommand(indexAddCmd)
+	indexCmd.AddCommand(indexListCmd)
+	indexCmd.AddCommand(indexDeleteCmd)
```
chriskim06: What's the reason for moving these out of the env var gate? If `indexCmd` isn't being made available can't we just skip these?

ahmetb: Practically makes no difference. So yeah maybe I shouldn't have changed this but it was prior to your patch so I kept it, it's harmless.

cmd/krew/cmd/index.go
ahmetb: +1

cmd/krew/cmd/index.go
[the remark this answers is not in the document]
ahmetb: Done.

cmd/krew/cmd/index.go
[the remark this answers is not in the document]
ahmetb: Done.

cmd/krew/cmd/index.go
ahmetb: Done.

cmd/krew/cmd/index.go
ahmetb: Done. `internal.PrintWarning` method created and existing uses migrated.

cmd/krew/cmd/index.go
```diff
@@ -62,11 +67,65 @@ var indexAddCmd = &cobra.Command{
...
+	}
+	return errors.Wrap(err, "error while removing the plugin index")
+}
+
+var (
+	forceIndexDelete *bool
+)
+
 func init() {
+	forceIndexDelete = indexDeleteCmd.Flags().Bool("force", false, "Remove index even if it has plugins installed (may yield in unsupported behavior)")
```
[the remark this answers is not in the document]
ahmetb: Done.

cmd/krew/cmd/index.go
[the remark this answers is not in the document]
chriskim06: I meant more about the "unsupported behavior" part. I feel like since there's a `--force` option it is supported, but it might result in unsupported behavior.

cmd/krew/cmd/index.go
```diff
@@ -62,14 +66,68 @@ var indexAddCmd = &cobra.Command{
...
+
+	if len(pl) > 0 && !*forceIndexDelete {
+		var names []string
+		for _, v := range pl {
+			names = append(names, v.Name)
+		}
+
+		internal.PrintWarning(os.Stderr, `Plugins [%s] are still installed from index %q.
+			"Removing indexes while there are plugins installed from them may result in
+			"unsupported behavior (use --force to ignore this check).`, strings.Join(names, ", "), name)
```
corneliusweig: This raw string looks a bit peculiar: Rows 2+3 are oddly indented.

Again, I'd also talk about "...not recommended."

cmd/krew/cmd/index.go
```diff
@@ -62,14 +66,68 @@ var indexAddCmd = &cobra.Command{
...
+			return nil // success if --force specified and index does not exist.
+		}
+		return errors.Errorf("index %q does not exist", name)
+	}
+	return errors.Wrap(err, "error while removing the plugin index")
+}
+
+var (
+	forceIndexDelete *bool
+)
```
corneliusweig: I would look for this var declaration right after the imports. Not sure if we have a recommended pattern for this, though..

internal/index/indexoperations/index.go
```diff
@@ -35,16 +35,16 @@ type Index struct {
 
 // ListIndexes returns a slice of Index objects. The path argument is used as
 // the base path of the index.
-func ListIndexes(path string) ([]Index, error) {
-	dirs, err := ioutil.ReadDir(path)
+func ListIndexes(paths environment.Paths) ([]Index, error) {
+	dirs, err := ioutil.ReadDir(paths.IndexBase())
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to read directory %s", path)
+		return nil, errors.New("failed to list directory")
```
corneliusweig: I would expect the path to be part of the error. AFAICT it no longer is.

internal/index/indexoperations/index.go
```diff
@@ -72,6 +72,19 @@ func AddIndex(path, name, url string) error {
...
+func DeleteIndex(paths environment.Paths, name string) error {
+	if !IsValidIndexName(name) {
+		return errors.New("invalid index name")
+	}
+
+	dir := paths.IndexPath(name)
+	if _, err := os.Stat(dir); err != nil {
+		return err
+	}
+	return os.RemoveAll(dir)
```
corneliusweig: Should we be extra careful here and check that `dir` is not a symlink? E.g. `ln ~ ~/.krew/index/bad-bad-bad; krew index rm bad-bad-bad`

integration_test/index_test.go
```diff
@@ -84,3 +84,50 @@ func TestKrewIndexList_NoIndexes(t *testing.T) {
...
+	}
+}
+
+func TestKrewIndexRemove_ok(t *testing.T) {
+	skipShort(t)
+	test, cleanup := NewTest(t)
+	test = test.WithEnv(constants.EnableMultiIndexSwitch, 1)
+	defer cleanup()
+
+	test.Krew("index", "add", "foo", constants.IndexURI).RunOrFail()
```
corneliusweig: That does another clone of the remote repo. Can we avoid that by using a local clone?

integration_test/index_test.go
```diff
@@ -84,3 +84,50 @@ func TestKrewIndexList_NoIndexes(t *testing.T) {
...
+}
+
+func TestKrewIndexRemoveFailsWhenPluginsInstalled(t *testing.T) {
+	skipShort(t)
+	test, cleanup := NewTest(t)
+	test = test.WithEnv(constants.EnableMultiIndexSwitch, 1)
+	defer cleanup()
+
+	test.Krew("install", validPlugin).RunOrFailOutput()
+	if err := test.Krew("remove", "default").Run(); err == nil {
```
corneliusweig:
```suggestion
	if err := test.Krew("index", "remove", "default").Run(); err == nil {
```

cmd/krew/cmd/index.go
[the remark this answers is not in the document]
ahmetb: Please give exact copy and I'll incorporate.

ahmetb:
> Its very existence is proof that this is supported behavior. If anything, I would prefer talking about "not recommended". Anyway, we should think hard if we really want that. My opinion is -- either support it -- or don't.

First of all we expect >95% users won't use `krew index` cmds. So anything we do here is "advanced" land. Furthermore `--force` says "not recommended" –which means this is even more advanced.

Here's my **primary rationale** for having `--force`: Assume you just added Jim’s index –but Jim messed up one of his manifest files, its parsing fails. So, naturally most our primary cmds (install, update, search) start failing. But you have plugins from Jim, so you can just go and "nuke" that index with the consequences.

`--force` puts you in an "unsupported" land because you might install plugin `foo`  from indexA , then `remove --force indexA`, install indexB which also has a `foo`.

integration_test/index_test.go
ahmetb: Will do.

internal/index/indexoperations/index.go
ahmetb: Given we manage what's cloned directly under ~/.krew/index , this means user has done something malicious to themselves by messing w/ krew internals. Also, `git clone` produces these dirs –which won't be symlinks.

I prefer not to protect the user from their own mistakes. 

Furthermore `ln -s a b; rm -rf b;` doesn't actually remove `a/*` based on my experimentation. Correct?

I don't see any need for action here + I'd prefer to keep things simple (+not have to write tests for that scenario).

integration_test/index_test.go
ahmetb: /facepalm

cmd/krew/cmd/index.go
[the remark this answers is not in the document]
chriskim06: My vote is for:
```
Remove a configured index repository which is used to discover and
install plugins from.

If there are plugins installed from the specified index, the index cannot be
removed (unless --force is used). Removing indexes while there are plugins
installed from them may result in unsupported behavior.
```

I like mentioning "unsupported behavior" because it conveys the not recommended part while alluding to potential consequences of performing this action.

cmd/krew/cmd/index.go
```diff
@@ -62,14 +66,68 @@ var indexAddCmd = &cobra.Command{
...
+removed (unless --force is used). Removing indexes while there are plugins
+installed from them is not supported behavior.`,
+	Args: cobra.ExactArgs(1),
+	RunE: indexDelete,
+}
+
+func indexDelete(_ *cobra.Command, args []string) error {
+	name := args[0]
+
+	pl, err := installation.InstalledPluginsFromIndex(paths.InstallReceiptsPath(), name)
```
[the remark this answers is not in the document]
ahmetb: Done.

cmd/krew/cmd/index.go
ahmetb: Done.

internal/index/indexoperations/index.go
ahmetb: Done.

integration_test/index_test.go
ahmetb: Done.

integration_test/index_test.go
ahmetb: Done.

cmd/krew/cmd/index.go
ahmetb: Done.

internal/index/indexoperations/index.go
corneliusweig: On the command line, yes. `RemoveAll()` seems to behave identically, so yeah I was a bit to paranoid. Maybe I'm losing my sanity over corona :wink:

corneliusweig:
Yeah, that all makes sense. I'm also ok with having `--force`, but even if we don't recommend using it, we should make sure that the core workflow continues to work without hiccups (`install`, `upgrade`, `update`, `list`, `uninstall`). I'm mostly concerned about `krew upgrade` which bails out when finding an unknown plugin at the moment. 

So in a way we _do_ support the core workflow after `rm --force` even though we claim to not support it. Or is that just dumb?

cmd/krew/cmd/index.go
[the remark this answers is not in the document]
corneliusweig: I suggest the following:

> Removes a plugin index.
>
> It is only safe to remove indexes without installed plugins. Removing an index while there are plugins installed will result in an error unless the `--force` option is used. This is not recommended.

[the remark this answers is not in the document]
ahmetb: Done.

ahmetb:
> I'm mostly concerned about `krew upgrade` which bails out when finding an unknown plugin at the moment.

I think there's no reason for us to support all those weird corner cases at the moment. We'll make `krew update` and `krew upgrade` tolerant to broken/unreachable indexes –but anything beyond that to make this bulletproof adds SO much complexity and very little in return.

PR #553, opened 2020-03-17
remove codecov.io integration

ahmetb:
After switching fully to GitHub Workflows we no longer use Codecov.io. We
also didn't really get much out of it.

/assign @corneliusweig

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,4 +7,3 @@
 /SECURITY_CONTACTS
 /.travis.yml
 /code-of-conduct.md
-/coverage.txt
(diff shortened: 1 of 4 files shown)


corneliusweig:
Can you add instructions to the readme how to get visual feedback for coverage? Like
```
go test -coverprofile=coverage.out
go tool cover -html=coverage.out
```

ahmetb:
Aren't they the same for all go projects? Why need that?

corneliusweig:
They are the same, yes. But we could still put it in our developer docs, just in case sbdy is not familiar with go and wants to contribute.

ahmetb:
Honestly, not sure why would it occur people to do that. I suspect most contributors don't read dev docs in full –so I'm inclined to leave it out. We review for missing tests anyway.

corneliusweig:
I sort of have a deja vu that we talked about once before. Let's leave as is until somebody complains :)
/lgtm
/approve

PR #555, opened 2020-03-21
Populate receipt status on install/upgrade

chriskim06:
Related issue: #483

### DIFF
--- a/cmd/krew/cmd/install.go
+++ b/cmd/krew/cmd/install.go
@@ -95,7 +95,7 @@ Remarks:
 			var install []index.Receipt
 			for _, name := range pluginNames {
 				indexName, pluginName := pathutil.CanonicalPluginName(name)
-				plugin, err := indexscanner.LoadPluginByName(paths.IndexPluginsPath(constants.DefaultIndexName), pluginName)
+				plugin, err := indexscanner.LoadPluginByName(paths.IndexPluginsPath(indexName), pluginName)
 				if err != nil {
 					if os.IsNotExist(err) {
 						return errors.Errorf("plugin %q does not exist in the plugin index", name)
--- a/cmd/krew/cmd/install.go
+++ b/cmd/krew/cmd/install.go
@@ -123,15 +123,16 @@ Remarks:
 				return cmd.Help()
 			}
 
-			for _, plugin := range install {
-				klog.V(2).Infof("Will install plugin: %s\n", plugin.Name)
+			for _, pluginReceipt := range install {
+				klog.V(2).Infof("Will install plugin: %s\n", pluginReceipt.Name)
 			}
 
 			var failed []string
 			var returnErr error
-			for _, plugin := range install {
+			for _, pluginReceipt := range install {
+				plugin := pluginReceipt.Plugin
 				fmt.Fprintf(os.Stderr, "Installing plugin: %s\n", plugin.Name)
-				err := installation.Install(paths, plugin, installation.InstallOpts{
+				err := installation.Install(paths, pluginReceipt, installation.InstallOpts{
 					ArchiveFileOverride: *archiveFileOverride,
 				})
 				if err == installation.ErrIsAlreadyInstalled {
--- a/cmd/krew/cmd/install.go
+++ b/cmd/krew/cmd/install.go
@@ -137,7 +137,7 @@ Remarks:
 			}
 
 			for _, pluginEntry := range install {
-				klog.V(2).Infof("Will install plugin: %s\n", pluginEntry.p.Name)
+				klog.V(2).Infof("Will install plugin: %s/%s\n", pluginEntry.indexName, pluginEntry.p.Name)
 			}
 
 			var failed []string
--- a/cmd/krew/cmd/install.go
+++ b/cmd/krew/cmd/install.go
@@ -27,11 +27,15 @@ import (
 	"sigs.k8s.io/krew/cmd/krew/cmd/internal"
 	"sigs.k8s.io/krew/internal/index/indexscanner"
 	"sigs.k8s.io/krew/internal/installation"
-	"sigs.k8s.io/krew/internal/installation/receipt"
 	"sigs.k8s.io/krew/internal/pathutil"
 	"sigs.k8s.io/krew/pkg/constants"
 	"sigs.k8s.io/krew/pkg/index"
 )
+
+type pluginEntry struct {
+	p         index.Plugin
+	indexName string
+}
 
 func init() {
 	var (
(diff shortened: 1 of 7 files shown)


cmd/krew/cmd/install.go
```diff
@@ -90,45 +92,47 @@ Remarks:
...
+				install = append(install, receipt.New(plugin, indexName))
 			}
 
 			if *manifest != "" {
 				plugin, err := indexscanner.ReadPluginFromFile(*manifest)
 				if err != nil {
 					return errors.Wrap(err, "failed to load plugin manifest from file")
 				}
-				install = append(install, plugin)
+				install = append(install, receipt.New(plugin, constants.DefaultIndexName))
```
ahmetb: I think we should not convert to `Receipt` until we actually installed, and we're about to write the `Receipt` object to disk. Thoughts @corneliusweig?

This complicates the code a bit, because nothing's installed yet, but scan index dir + convert everything to Receipt in-memory already. Looks odd in the code to me.

That's why I was imagining a small internal type, specific to this pkg that contains Plugin+indexname only.

ahmetb: I have something like this in mind:

```go
type pluginEntry {
   p plugin
   indexName string
}
```

internal/installation/install.go
```diff
@@ -54,7 +54,8 @@ var (
 
 // Install will download and install a plugin. The operation tries
 // to not get the plugin dir in a bad state if it fails during the process.
-func Install(p environment.Paths, plugin index.Plugin, opts InstallOpts) error {
+func Install(p environment.Paths, pluginReceipt index.Receipt, opts InstallOpts) error {
```
ahmetb: this should not receive a Receipt, conceptually sounds wrong.

internal/installation/receipt/receipt.go
```diff
@@ -21,12 +21,13 @@ import (
...
 
 	"sigs.k8s.io/krew/internal/index/indexscanner"
+	"sigs.k8s.io/krew/pkg/constants"
 	"sigs.k8s.io/krew/pkg/index"
 )
 
 // Store saves the given plugin receipt at the destination.
 // The caller has to ensure that the destination directory exists.
-func Store(plugin index.Plugin, dest string) error {
+func Store(plugin index.Receipt, dest string) error {
```
ahmetb: you might wanna rename variable name to `receipt` as well + update godoc. :)

internal/installation/receipt/receipt.go
```diff
@@ -41,3 +42,19 @@ func Store(plugin index.Plugin, dest string) error {
 func Load(path string) (index.Receipt, error) {
 	return indexscanner.ReadReceiptFromFile(path)
 }
+
+// New returns a new receipt with the given plugin and index name.
+func New(plugin index.Plugin, indexName string) index.Receipt {
+	name := indexName
+	if name == "" {
+		name = constants.DefaultIndexName
+	}
```
ahmetb: under what condition we specify "" ?

cmd/krew/cmd/install.go
chriskim06: Ya I was going back and forth on that a lot when I was working on this. It felt weird to add a new type that was essentially the same as `Receipt` but I agree that it is weird/conceptually wrong to create the receipt before the plugin is actually installed.

internal/installation/receipt/receipt.go
chriskim06: You're right, I forgot the `CanonicalPluginName` function returns the default index already. I can remove this.

ahmetb: Even without that, this should not happen. We'll always have an index name in pkg/installation.

`CanonicalPluginName` should be used only for turning user input (e.g. install args) to full plugin names.

cmd/krew/cmd/install.go
corneliusweig: Yeah, this is definitely cleaner. Just because `Receipt` happens to have the field that we want to store in addition does not mean it is the right choice. Consider what happens when we add more fields to `Receipt`. What would happen most likely is that we would not update this instance and then we have a in invalid `Receipt` in our hands. So a dedicated struct is what we need here.

internal/installation/receipt/receipt.go
corneliusweig: It sounds like you are talking about the same thing. Anyway, everybody agrees that the check `name == ""` is unnecessary.

cmd/krew/cmd/install.go
```diff
@@ -90,45 +96,56 @@ Remarks:
...
 
 			if *manifest != "" {
 				plugin, err := indexscanner.ReadPluginFromFile(*manifest)
 				if err != nil {
 					return errors.Wrap(err, "failed to load plugin manifest from file")
 				}
-				install = append(install, plugin)
+				install = append(install, pluginEntry{
+					p:         plugin,
+					indexName: constants.DefaultIndexName,
```
ahmetb: this is a case we did not think of.
I guess we can keep shoving them into default. thoughts @corneliusweig ?

ahmetb:
Right now we lack any integration tests for the changes.

Note that you're changing criticial path in this PR. So we need to make sure newly added behavior works. We can merge, but the very next PR we need would be the integration tests.

Some basic cases I'm thinking easily:

- add index + install with its prefix
- install plugins with `default/` prefix

chriskim06:
I can add the integration tests here or in a separate one if that'll make it easier. Let me know which one you'd prefer.

cmd/krew/cmd/install.go
```diff
@@ -90,45 +96,56 @@ Remarks:
...
 			}
 
 			if len(install) == 0 {
 				return cmd.Help()
 			}
 
-			for _, plugin := range install {
-				klog.V(2).Infof("Will install plugin: %s\n", plugin.Name)
+			for _, pluginEntry := range install {
+				klog.V(2).Infof("Will install plugin: %s\n", pluginEntry.p.Name)
```
ahmetb: this will omit the index name.

ahmetb:
> I can add the integration tests here or in a separate one if that'll make it easier. Let me know which one you'd prefer.

I'm curious how you tested this currently. I'm trying to get it to work locally and tried a fake local index at a directory, and wasn't able to install. So the integration tests can actually verify it works.

chriskim06:
That's strange. I pretty much did what you mentioned and what's in the integration test: add index then install and it worked.

ahmetb:
/lgtm
/approve

nice and small!

ahmetb:
While testing locally I noticed that if I create a fake index dir, do `git init .` and locally and add that, the repo is cloned but it is empty (since it has no commits).

So we need to document that while testing with local index repos, they need to commit the changes, so that the `update` or `index add` mechanism can clone the copy of the index from a commit.

cmd/krew/cmd/install.go
corneliusweig: This is what happens at the moment. But we can use the opportunity to separate this better from the `default` index. So instead, what about setting the indexName to `detached` or something similar?

ahmetb: ++ @chriskim06 do you mind making a small PR?

PR #558, opened 2020-03-27
Refactor ListInstalledPlugins

chriskim06:
Small refactor that I think will be necessary for updating search/list/upgrade to work with custom indexes.

Related issue: #483

### DIFF
--- a/cmd/krew/cmd/search.go
+++ b/cmd/krew/cmd/search.go
@@ -54,7 +54,7 @@ Examples:
 			pluginMap[p.Name] = p
 		}
 
-		receipts, err := installation.ListInstalledPluginReceipts(paths.InstallReceiptsPath())
+		receipts, err := installation.GetInstalledPluginReceipts(paths.InstallReceiptsPath())
 		if err != nil {
 			return errors.Wrap(err, "failed to load installed plugins")
 		}
--- a/cmd/krew/cmd/search.go
+++ b/cmd/krew/cmd/search.go
@@ -54,9 +54,13 @@ Examples:
 			pluginMap[p.Name] = p
 		}
 
-		installed, err := installation.ListInstalledPlugins(paths.InstallReceiptsPath())
+		receipts, err := installation.ListInstalledPlugins(paths.InstallReceiptsPath())
 		if err != nil {
 			return errors.Wrap(err, "failed to load installed plugins")
+		}
+		installed := make(map[string]string)
+		for _, receipt := range receipts {
+			installed[receipt.Name] = receipt.Spec.Version
 		}
 
 		var matchNames []string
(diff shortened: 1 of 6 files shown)


cmd/krew/cmd/search.go
```diff
@@ -54,10 +54,14 @@ Examples:
...
 		}
 
-		installed, err := installation.ListInstalledPlugins(paths.InstallReceiptsPath())
+		receipts, err := installation.GetInstalledPluginReceipts(paths.InstallReceiptsPath())
 		if err != nil {
 			return errors.Wrap(err, "failed to load installed plugins")
 		}
+		installed := make(map[string]string)
+		for _, receipt := range receipts {
+			installed[receipt.Name] = receipt.Spec.Version
```
ahmetb: this sounds wrong. what if two plugins with identical names from different indexes exist? it'll show both as installed.

chriskim06: This is here to preserve the existing functionality. I'm not trying to update search to work with custom indexes in this PR.

cmd/krew/cmd/update.go
```diff
@@ -120,10 +120,14 @@ func ensureIndexUpdated(_ *cobra.Command, _ []string) error {
...
+	receipts, err := installation.GetInstalledPluginReceipts(paths.InstallReceiptsPath())
 	if err != nil {
 		return errors.Wrap(err, "failed to load installed plugins list after update")
 	}
+	installedPlugins := make(map[string]string)
+	for _, receipt := range receipts {
+		installedPlugins[receipt.Name] = receipt.Spec.Version
+	}
 
 	showUpdatedPlugins(os.Stderr, preUpdateIndex, posUpdateIndex, installedPlugins)
```
ahmetb: this part need more comprehensive and careful refactoring. 
I think we need to think heavier refactoring of this part because of

1. plugin name collisions between indexes
2. we ought to show index/ prefix to updated plugins as well.

Maybe let's comment out this feature (showing updated stuff) and let's add a TODO to reactivate in upcoming PRs. Not sure. @corneliusweig thoughts?

cmd/krew/cmd/search.go
ahmetb: Yeah makes sense. That said do you mind adding a TODO so we don't forget these things? The same works below, too.

cmd/krew/cmd/update.go
```diff
@@ -129,6 +129,7 @@ func ensureIndexUpdated(_ *cobra.Command, _ []string) error {
 		installedPlugins[receipt.Name] = receipt.Spec.Version
 	}
 
+	// TODO(chriskim06) consider commenting this out when refactoring for custom indexes
```
ahmetb: not quite commenting out, but I think we need to pass a similar Pair<indexName,plugin> here as well :) :)

chriskim06: Ya `pluginEntry` would be available here. I was just making a note of possibly commenting out the show updated plugins feature like you mentioned in your other comment

PR #560, opened 2020-03-28
cmd: refactor `list` to work with custom indexes

chriskim06:
Pretty small change. This leverages the new `GetInstalledPluginReceipts` function to prepend the index name to installed plugins for the list command. I added a function, `receipt.CanonicalName`, to get the "canonical" name of the plugin from the receipt (let me know if I'm using that incorrectly and the function name should be changed). I think this will also be useful down the road when search and upgrade are refactored.

Related issue: #483

### DIFF
--- a/internal/installation/receipt/receipt.go
+++ b/internal/installation/receipt/receipt.go
@@ -43,6 +43,8 @@ func Load(path string) (index.Receipt, error) {
 	return indexscanner.ReadReceiptFromFile(path)
 }
 
+// CanonicalName returns the index and plugin name from a receipt.
+// The index name is omitted if it is the default index.
 func CanonicalName(receipt index.Receipt) string {
 	name := receipt.Name
 	if receipt.Status.Source.Name != constants.DefaultIndexName {
--- a/internal/installation/receipt/receipt.go
+++ b/internal/installation/receipt/receipt.go
@@ -21,6 +21,7 @@ import (
 	"sigs.k8s.io/yaml"
 
 	"sigs.k8s.io/krew/internal/index/indexscanner"
+	"sigs.k8s.io/krew/pkg/constants"
 	"sigs.k8s.io/krew/pkg/index"
 )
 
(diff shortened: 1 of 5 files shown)


chriskim06:
Not sure if this will work for me but want to try it out

/area multi-index

internal/installation/receipt/receipt.go
```diff
@@ -42,6 +43,16 @@ func Load(path string) (index.Receipt, error) {
 	return indexscanner.ReadReceiptFromFile(path)
 }
 
+// CanonicalName returns the index and plugin name from a receipt.
+// The index name is omitted if it is the default index.
+func CanonicalName(receipt index.Receipt) string {
```
ahmetb: something defined on the type could be more idiomatic.

`func (r index.Receipt) CanonicalName() string`

ahmetb: this is actually not a `CanonicalName`.
Canonical name of a `foo` plugin is `default/foo`.

If anything, this is a `DisplayName` method.

Therefore, I recommend we define this in `cmd/` package.
Also keep in mind this will come up in many other use cases e.g.:

- `krew search`, similarly listing plugins
- `krew install`/`upgrade` calling `krew update` implicitly which lists new/updated plugins in the index(es).

internal/installation/receipt/receipt_test.go
```diff
@@ -73,6 +74,38 @@ func TestLoad_preservesNonExistsError(t *testing.T) {
...
+		name     string
+		receipt  index.Receipt
+		expected string
+	}{
+		{
+			name:     "from default index",
+			receipt:  testutil.NewReceipt().WithPlugin(testutil.NewPlugin().WithName("foo").V()).V(),
+			expected: "foo",
+		},
+		{
```
ahmetb: the missing test case here would be that there's an explicit status, it's from index `default`.

internal/installation/receipt/receipt.go
ahmetb: discussed offline, we've decided to move this to cmd/.

- displaying is a CLI concern and all the functionality in there would use this
- we decided not to add it on Receipt type, to keep the pkg/index (one of our two public pkgs) small and type-only

ahmetb:
thanks!
/lgtm
/approve

PR #563, opened 2020-03-28
cmd/info: Support for multiple indexes

### LINKED ISSUE #559: Proposal: make `krew info` work with multiple indexes
Since we had 1 index up until now, `krew index` had been fairly easy to implement. It did not distinguish between `Receipt`s or `Plugin`s (ref: `LoadManifestFromReceiptOrIndex`).

In other words, an "installed" plugin could only come from the Plugin manifest with the same name –as there were no other index.

So now, by allowing multiple indexes, `krew info` now has to understand if:

    krew info foo

means `foo` the installed plugin (could be from any index), or the `default/foo` uninstalled plugin.

-----

## Option 1

I recommend we change krew info’s behavior. `krew info foo` means:

- show info of `foo` installed plugin (could be from any index)
- show info of `foo` uninstalled plugin (from `default` index)

To disambiguate an "uninstalled" plugin, the user can run `krew info my-index/foo`. However, if `my-index/foo` is installed, could the user have meant show the "installed" plugin?

Keep in mind that `krew info` works on plugins installed via `--manifest` (i.e. has no index, but recorded as `default` with #555).

I'm losing my train of thought here.
If you see a straightforward solution, let me know.
/cc @corneliusweig 
cc: @chriskim06 
/kind proposal
/area multi-index

ahmetb:
Somewhat breaking change. `krew info` now looks up ONLY available plugins
(no more looking at installed plugin receipts). The supported syntaxes are:

    kubectl krew info NAME # assumes 'default'
    kubectl krew info INDEX/NAME

With this change, `info` cmd will no longer work with plugins installed via
`--manifest` (as we won't be looking at installation receipts anymore).

/area multi-index
/assign @corneliusweig
cc: @chriskim06
Related issue: #483
fixes #559


cmd/krew/cmd/info.go
```diff
@@ -25,38 +25,41 @@ import (
...
-	Short: "Show information about a kubectl plugin",
-	Long: `Show information about a kubectl plugin.
+	Short: "Show information about an available plugin",
+	Long: `Show detailed information about an available plugin.
 
-This command can be used to print information such as its download URL, last
-available version, platform availability and the caveats.
+Examples:
+  kubectl krew info PLUGIN
+  kubectl krew info INDEX/PLUGIN`,
```
corneliusweig: cobra.Command has an `Examples` section. This should go in there.

corneliusweig: So far, the multi-index features are behind a feature gate. This will expose the new syntax without it. That's a bit confusing.

cmd/krew/cmd/info.go
```diff
@@ -25,38 +25,41 @@ import (
...
 	Use:   "info",
-	Short: "Show information about a kubectl plugin",
-	Long: `Show information about a kubectl plugin.
+	Short: "Show information about an available plugin",
+	Long: `Show detailed information about an available plugin.
 
-This command can be used to print information such as its download URL, last
-available version, platform availability and the caveats.
+Examples:
+  kubectl krew info PLUGIN
```
corneliusweig: We should say that this looks in the default index. Or is there a better place to explain the index resolution logic?

cmd/krew/cmd/info.go
```diff
@@ -25,38 +25,41 @@ import (
...
-		printPluginInfo(os.Stdout, plugin)
+		printPluginInfo(os.Stdout, p, index)
 		return nil
 	},
 	PreRunE: checkIndex,
 	Args:    cobra.ExactArgs(1),
 }
 
-func printPluginInfo(out io.Writer, plugin index.Plugin) {
+func printPluginInfo(out io.Writer, plugin index.Plugin, indexName string) {
```
corneliusweig: Can we swap the order of arguments? Usually, indexName comes first.
```suggestion
func printPluginInfo(out io.Writer, indexName string, plugin index.Plugin) {
```

integration_test/info_test.go
```diff
@@ -37,3 +46,22 @@ func TestKrewInfoInvalidPlugin(t *testing.T) {
...
+func TestKrewInfoCustomIndex(t *testing.T) {
+	skipShort(t)
+
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	test = test.WithEnv(constants.EnableMultiIndexSwitch, 1).WithIndex()
+	defaultIndexRoot := test.TempDir().Path("index/" + constants.DefaultIndexName)
+
+	test.Krew("index", "add", "foo", defaultIndexRoot).RunOrFail()
```
corneliusweig: That trick has been used quite often and it's not straightforward what's happening. Mind to extract this into a helper?

cmd/krew/cmd/info.go
ahmetb: IIRC it’s a single Example field. So we need to figure out what indentation does the template use so we can have multiple lines.

ahmetb: ++ we should put it behind the gate

cmd/krew/cmd/info.go
ahmetb: IMO not needed since it’s implied from “search” and “list” outputs. 

Similarly brew commands use a x/taps/y syntax but they just call it FORMULA. They don’t really make it too obvious to everyone that taps concept exists as majority doesn’t need to know.

integration_test/info_test.go
ahmetb: I will open an issue to do this later. This multi-index work has caused few other cleanup items that are increasing visible.

cmd/krew/cmd/info.go
corneliusweig: Ah, just saw that we don't use it in krew. So nevermind..

cmd/krew/cmd/info.go
corneliusweig: Ok, makes sense.

cmd/krew/cmd/info.go
ahmetb: We use it in _certain_ cmds that have a single example. But `install`, `search` have rather complex examples, so they're in LongDescription.

integration_test/info_test.go
ahmetb: opened #565

PR #567, opened 2020-03-29
remove unused type installation.PluginEntry

ahmetb:
currently it appears that the pluginEntry type can just be in the
cmd/krew pkg. 

cc: @chriskim06
/assign @corneliusweig
/area multi-index


chriskim06:
Whoops, I must have accidentally left that in one of my previous PRs /facepalm

PR #568, opened 2020-03-29
Use special "detached" index for manifest plugins

chriskim06:
Let me know if you want me to move this out to a package level variable, figured it would be ok like this since this situation is only encountered in this file.

Related issue: #483

### DIFF
--- a/integration_test/install_test.go
+++ b/integration_test/install_test.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"testing"
 
+	"sigs.k8s.io/krew/internal/installation/receipt"
 	"sigs.k8s.io/krew/pkg/constants"
 )
 
@@ -132,6 +133,7 @@ func TestKrewInstall_Manifest(t *testing.T) {
 		"--manifest", filepath.Join("testdata", validPlugin+constants.ManifestExtension)).
 		RunOrFail()
 	test.AssertExecutableInPATH("kubectl-" + validPlugin)
+	assertManifestPluginIndex(t, test, validPlugin)
 }
 
 func TestKrewInstall_ManifestURL(t *testing.T) {
@@ -146,6 +148,7 @@ func TestKrewInstall_ManifestURL(t *testing.T) {
 		"--manifest-url", srv+"/"+validPlugin+constants.ManifestExtension).
 		RunOrFail()
 	test.AssertExecutableInPATH("kubectl-" + validPlugin)
+	assertManifestPluginIndex(t, test, validPlugin)
 }
 
 func TestKrewInstall_ManifestAndArchive(t *testing.T) {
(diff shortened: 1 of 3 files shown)


ahmetb:
> Let me know if you want me to move this out to a package level variable, figured it would be ok like this since this situation is only encountered in this file.

I think it’s fine.

ahmetb:
Since we introduce a behavior, we should test.

Maybe add a test to integ tests of krewinstallmanifest ones loading receipt and checking source==detached.

ahmetb:
> So let's extract it to `constants.DetachedIndexName` or similar.

But technically we’re offering these libraries as part of our API. Would anyone build tooling/automation around this const? If not, better not expose?

chriskim06:
The integration test failed for some reason, I think I vaguely remember seeing this in other PRs. I'll investigate further later tonight.

ahmetb:
> The integration test failed for some reason (worked locally), I think I vaguely remember seeing this in other PRs. I'll investigate further later tonight.

Seems like it was a flake (somehow), re-running the workflow worked. It happened at least once before. I'm suspecting GitHub's endpoint sending us an error response somehow. Need to look at download code to see if it's checking for non-200 codes.

integration_test/install_test.go
```diff
@@ -216,3 +220,14 @@ func localTestServer() (string, func()) {
 	s := httptest.NewServer(http.FileServer(http.Dir("testdata")))
 	return s.URL, s.Close
 }
+
+func assertManifestPluginIndex(t *testing.T, test *ITest, plugin string) {
```
ahmetb: assertPluginFromIndex is a better name potentially?
also why is this not defined on `test` (like assertExecutableInPath)? that way you don't have to pass it `t`.

integration_test/install_test.go
```diff
@@ -216,3 +220,14 @@ func localTestServer() (string, func()) {
...
+	receiptPath := test.TempDir().Path("receipts/" + plugin + constants.ManifestExtension)
```
ahmetb: don't guess filesystem layout; use environment.Paths (unless you're actually testing user-visible behavior, breaking behavior between versions, or Paths type).

integration_test/install_test.go
chriskim06: I can add it there. I added this to just check "detached" and thought that would be a little too specific to `install --manifest` to add to `testutil_test.go`.

chriskim06:
@corneliusweig @ahmetb let me know what you guys think about "detached". There's now a few usages of it in the integration test so would a good compromise be to make it a public const in the cmd package so that its not exposed but still available within the code? Although I think that might be a little strange seeing `cmd.DetachedIndexName`.

ahmetb:
@chriskim06 nothing wrong with actually duplicating those in integration_test.

In fact, tests should rely on consts less and less. Because we want to test "behavior". If someone changes the value of the const, that should break the tests to notify us of behavior change, so please refrain from using the const in the tests.

ahmetb:
/lgtm
/approve

Thanks!

PR #572, opened 2020-03-30
integration_test: Use new helper method

ahmetb:
* refactor AssertPluginFromIndex
  * no need to pass t
  * mark as helper
  * swap parameter order to match method name + more intuitive

* refactor existing "krew install" integ tests to use this

Related: #568
/assign @chriskim06
/area multi-index


chriskim06:
Nice!

/lgtm

Issue #566, opened 2020-03-29
Implement custom plugin indexes


chriskim06:
Here's a list of some of the remaining work regarding support for multiple indexes:
- [x] update `krew list` to work with multi-index (#560)
- [x] update `krew info` to work with multi-index (#563)
- [x] update `krew search` to work with multi-index (#574)
- [x] update `krew upgrade` to work with multi-index (#594)
- [x] update `krew update` to work with multi-index (#588)
- [x] ~~update `krew install` to work with multi-index~~ (no work needed?)
- [x] update `krew remove` to work with multi-index (#579)
- [x] implement `krew index add|remove|list` commands (#537, #545, #552)
- [x] refactor the code that shows updated/new plugins during update to also show index (#593)
- [x] cleanup issues (#562, #565, #625)
- [x] revisit `--help` messages to ensure examples reflect index/ syntax where applicable (#597)
- [x] discuss a strategy for index directory layout migration (#600)
- [x] rethink how the "local plugin index is not initialized" message should read, or how we suggest users to add krew-index if there are no indexes? (#595)
- [x] write integration tests (like `receiptsmigration`) to ensure index migration happens v0.3.4 -> x (#621)
- [x] revisit commands to ensure they can cope with an unreachable index (or an index w/ malformed yaml) (#614)
- [x] remove multi-index feature flag (#624)
- [x] update user guide to add documentation on custom plugin indexes (#618)
- [x] update developer guide to add documentation on hosting custom plugin indexes + what to pay attention to (#618)
- [x] define trust model for custom indexes for users (#576, #615)
- [x] make multi-index docs "not draft"
- [x] tag a new release
- [x] find solution to krew’s own first-time installation being "detached"

Proposal issue: #483
Proposal doc: [link](https://docs.google.com/document/d/1Im5gmBA6yuKRWgJWbaBjTTFNvkH0me05V3wEeNPU7fE/edit)

/area multi-index

ahmetb:
I'm scratching off these two for now as I see no work necessary there. 

>  ~update `krew install` to work with multi-index~ (no work needed?)
>  ~update `krew remove` to work with multi-index~ (no work needed?)

"install" already works (correct me if I'm wrong), and "remove" has no notion of indexes (as it does not need to).

chriskim06:
[...]

PR #574, opened 2020-03-30
cmd: make krew search work with multiple indexes

ahmetb:
Mostly just refactoring. Added test to make sure search output is
alphabetically sorted (even in case of multiple indexes).

Related issue: #566, #483
/assign @chriskim06 
/assign @corneliusweig 
/area multi-index

### DIFF
--- a/cmd/krew/cmd/search.go
+++ b/cmd/krew/cmd/search.go
@@ -22,11 +22,12 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sahilm/fuzzy"
 	"github.com/spf13/cobra"
+	"k8s.io/klog"
 
+	"sigs.k8s.io/krew/internal/index/indexoperations"
 	"sigs.k8s.io/krew/internal/index/indexscanner"
 	"sigs.k8s.io/krew/internal/installation"
 	"sigs.k8s.io/krew/pkg/constants"
-	"sigs.k8s.io/krew/pkg/index"
 )
 
 // searchCmd represents the search command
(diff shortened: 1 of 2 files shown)


cmd/krew/cmd/search.go
```diff
@@ -43,58 +44,91 @@ Examples:
...
+				return errors.Wrapf(err, "failed to list plugin indexes available")
+			}
+			indexes = out
 		}
 
+		klog.V(3).Infof("found %d indexes", len(indexes))
+
+		var plugins []pluginEntry
+		for _, idx := range indexes {
+			ps, err := indexscanner.LoadPluginListFromFS(paths.IndexPluginsPath(constants.DefaultIndexName))
```
chriskim06: Should this be `paths.IndexPluginsPath(idx.Name)`? As it is now it seems like it will only load plugins from the default index even if `X_KREW_ENABLE_MULTI_INDEX` is set.

ahmetb: Nice catch! Why the hell are tests not catching this I’ll take a look.

chriskim06: It's because we copy the default index into the custom index

ahmetb: Ah, right. I guess it cannot be helped for now :)

ahmetb:
/hold
for cornelius review

cmd/krew/cmd/search.go
```diff
@@ -43,58 +44,91 @@ Examples:
...
+			}
+		}
+
+		keys := func(v map[string]pluginEntry) []string {
+			out := make([]string, 0, len(v))
+			for k := range v {
+				out = append(out, k)
+			}
+			return out
+		}
```
corneliusweig: There's only a single use of this lambda. Have you considered inlining it?

A proper function would also make sense because this does not capture any variables.

cmd/krew/cmd/search.go
```diff
@@ -43,58 +44,91 @@ Examples:
...
 
 		var rows [][]string
 		cols := []string{"NAME", "DESCRIPTION", "INSTALLED"}
-		for _, name := range matchNames {
-			plugin := pluginMap[name]
+		for _, name := range searchResults {
+			v := pluginMap[name]
 			var status string
-			if _, ok := installed[name]; ok {
+			if index := installed[v.p.Name]; index == v.indexName {
```
corneliusweig: This is a slight abuse of `map[string]string` for a map of a pair of properties. Iow, instead of `{"plugin": "index"}` we should have `{"index/plugin": true}`.
You already do this for `pluginMap`, why not for `installed` too?

cmd/krew/cmd/search.go
```diff
@@ -43,58 +44,91 @@ Examples:
...
-		if len(matchNames) == 0 {
+		if len(searchResults) == 0 {
 			return nil
 		}
 
 		var rows [][]string
 		cols := []string{"NAME", "DESCRIPTION", "INSTALLED"}
-		for _, name := range matchNames {
-			plugin := pluginMap[name]
+		for _, name := range searchResults {
```
corneliusweig:
```suggestion
		for _, canonicalName := range searchResults {
```
would make it clearer that this name includes the index.

integration_test/search_test.go
```diff
@@ -46,3 +50,52 @@ func TestKrewSearchOne(t *testing.T) {
 		t.Errorf("The first match should be krew")
 	}
 }
+
+func TestKrewSearchMultiIndex(t *testing.T) {
```
corneliusweig: Nice tests!

cmd/krew/cmd/search.go
ahmetb: Yeah, that'd work to let me refactor.

cmd/krew/cmd/search.go
ahmetb: will do

cmd/krew/cmd/search.go
ahmetb: Done.

cmd/krew/cmd/search.go
ahmetb: Done.

cmd/krew/cmd/search.go
ahmetb: Done.

corneliusweig:
Nice work!
/lgtm
/approve
/hold cancel

PR #575, opened 2020-03-30
integration_test: change validPlugin2

ahmetb:
mtail just recently became a larger Go based plugin, so change.

/assign @corneliusweig


corneliusweig:
/lgtm
/approve
Good catch!

PR #582, opened 2020-04-01
integration_test: Refine cmd run utilities

### LINKED ISSUE #561: integration_test: RunOrFailOutput + RunOrFail is redundant
I noticed that integration tests with `.RunOrFail()` actually fail not so useful error messages like `krew [index add]: exit error 255`.

Further exploration revealed that there's actually no value add of `.RunOrFail()`.

image

/good-first-issue
/help

ahmetb:
- refactored ITest.Run() to return ([]byte,error)
- refactored all Run*** methods to use ITest.Run()

This also will help us test commands that are expected to fail but were
previously not making their stdout/stderr available to the test case for
string tests.

Fixes #561
/assign @corneliusweig

### DIFF
--- a/integration_test/testutil_test.go
+++ b/integration_test/testutil_test.go
@@ -205,33 +205,9 @@ func (it *ITest) WithStdin(r io.Reader) *ITest {
 	return it
 }
 
-// RunOrFail runs the krew command and fails the test if the command returns an error.
-func (it *ITest) RunOrFail() {
-	it.t.Helper()
-	if err := it.Run(); err != nil {
-		it.t.Fatal(err)
-	}
-}
-
-// Run runs the krew command.
-func (it *ITest) Run() error {
-	it.t.Helper()
-
-	cmd := it.cmd(context.Background())
-	it.t.Log(cmd.Args)
-
-	start := time.Now()
-	if err := cmd.Run(); err != nil {
-		return errors.Wrapf(err, "krew %v", it.args)
-	}
-
-	it.t.Log("Ran in", time.Since(start))
-	return nil
-}
-
-// RunOrFailOutput runs the krew command and fails the test if the command
-// returns an error. It only returns the standard output.
-func (it *ITest) RunOrFailOutput() []byte {
+// Run runs the krew command, and returns its combined output even when
+// it fails.
+func (it *ITest) Run() ([]byte, error) {
 	it.t.Helper()
 
 	cmd := it.cmd(context.Background())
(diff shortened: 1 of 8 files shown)


integration_test/testutil_test.go
```diff
@@ -205,48 +205,46 @@ func (it *ITest) WithStdin(r io.Reader) *ITest {
...
 	start := time.Now()
 	if err := cmd.Run(); err != nil {
-		return errors.Wrapf(err, "krew %v", it.args)
+		out := b.Bytes()
+		return out, errors.Wrapf(err, "krew %v: %v, %s", it.args, err, string(out))
 	}
-
 	it.t.Log("Ran in", time.Since(start))
-	return nil
+	return b.Bytes(), nil
```
corneliusweig, line 229: This can be simpler:
```go
err := cmd.Run()
it.t.Log("Ran in", time.Since(start))
return b.Bytes(), errors.Wrapf(err, "krew %v: %v, %s", it.args, err, string(out))
```
WDYT?

integration_test/testutil_test.go
```diff
@@ -205,48 +205,46 @@ func (it *ITest) WithStdin(r io.Reader) *ITest {
...
+}
+
+// RunOrFail runs the krew command and fails the test if the command returns an error.
+func (it *ITest) RunOrFail() {
```
corneliusweig, line 233: Is that still used? IMO we can also get rid of it and make the `t.Fatal(..)` more visible in the calling code. But this doesn't have to be here.

integration_test/testutil_test.go
```diff
@@ -205,48 +205,46 @@ func (it *ITest) WithStdin(r io.Reader) *ITest {
...
-	if err := it.Run(); err != nil {
-		it.t.Fatal(err)
-	}
-}
-
-// Run runs the krew command.
-func (it *ITest) Run() error {
+// Run runs the krew command, and returns its combined output even when
+// it fails.
+func (it *ITest) Run() ([]byte, error) {
```
corneliusweig, line 210:
```suggestion
func (it *ITest) Run() (string, error) {
```
I could be mistaken, but I think we literally _everywhere_ convert the returned byte array to a string. So we could just move the conversion into this helper.

integration_test/testutil_test.go
ahmetb: yeah 👍  seems like it except we need to call `b.Bytes()` only once

integration_test/testutil_test.go
ahmetb: yeah this has 33 use cases, so it helps us save 33 `if err := ... ; err != nil { t.fatalf }` blocks :)

integration_test/testutil_test.go
ahmetb: `RunOrFailOutput` is `[]byte` so I kept it consistent w/ that.
We can convert both to `string` in another PR, wanted to keep this patch short.

corneliusweig:
/lgtm
/approve
Thanks for honing our test utils. This is worth so much!

PR #583, opened 2020-04-01
Check index or plugin name safety in cmd

### LINKED ISSUE #581: Some Krew commands susceptible to path manipulation
This is low-risk security issue so I'm opening it here.

While using `KREW_ROOT=~/.krew2`, I can still manipulate stuff in `~/.krew/` through the CLI by hijacking `PLUGIN` arguments with relative paths:

```
$ krew remove ../../.krew/receipts/whoami
Uninstalled plugin ../../.krew/receipts/whoami
```

This works even if `whoami` plugin is not installed in `~/.krew2` because we point to a valid receipt, which then gets deleted.

This happens because:
- `remove` command is not actually using `validation.IsSafePluginName` (like `LoadPluginByName` does) because it's operating on `Receipt`s
  - We actually added this check so that people don't use `name:` field in the plugin manifest to get random fs access.
- `environment.Paths` is vulnerable to manipulation because it doesn't do checks or return errors. I think this is by design

`krew index` cmds are not susceptible 
The methods who use LoadPluginByName fail correctly:

```
$ out/bin/krew-darwin_amd64 info ../../.krew/receipts/warp
F0401 11:02:03.677641   50768 root.go:79] failed to load plugin manifest: plugin name
"../.krew/receipts/warp" not allowed
```
```
out/bin/krew-darwin_amd64 upgrade ../../.krew/receipts/warp
F0401 11:04:52.552720   50879 root.go:79] failed to load the plugin manifest for plugin ../../.krew/receipts/warp: plugin name
"../.krew/receipts/warp" not allowed
```
```
out/bin/krew-darwin_amd64 install ../../.krew/receipts/warp
F0401 11:02:27.087157   50788 root.go:79] failed to load plugin "../../.krew/receipts/warp" from the index: plugin name
"../.krew/receipts/warp" not allowed
```

`krew index [add|remove]` are not susceptible to this as they use `IsValidIndexName` (although they do this check in pkg/indexoperations; not cmd/)

I propose we scan all cmds and validate plugin/index names with respective methods. The part that I'm not sure: should we do this in `cmd/` or `pkg/*`? Technically, cmd/ is a better place as it reduces test burden in lower layers.

ahmetb:
tl;dr: Some krew cmds are currently susceptible to path manipulation through positional arguments. Low-level machinery isn't responsible for this, so move them to cmd + add tests.

Moving plugin name or index name safety checks to cmd/ where they are given by
the user to the program. This way we shift the need for validation from being
unit tests of low-level machinery to the integration_test which tests
user-facing concerns.

> Note: Checks will fail until #582 is merged and this patch is rebased on it.

Fixes #581
/assign @corneliusweig
/assign @chriskim06

### DIFF
--- a/cmd/krew/cmd/install.go
+++ b/cmd/krew/cmd/install.go
@@ -98,11 +98,11 @@ Remarks:
 
 			var install []pluginEntry
 			for _, name := range pluginNames {
-				if !validation.IsSafePluginName(name) {
-					return unsafePluginNameErr(name)
+				indexName, pluginName := pathutil.CanonicalPluginName(name)
+				if !validation.IsSafePluginName(pluginName) {
+					return unsafePluginNameErr(pluginName)
 				}
 
-				indexName, pluginName := pathutil.CanonicalPluginName(name)
 				plugin, err := indexscanner.LoadPluginByName(paths.IndexPluginsPath(indexName), pluginName)
 				if err != nil {
 					if os.IsNotExist(err) {
--- a/cmd/krew/cmd/install.go
+++ b/cmd/krew/cmd/install.go
@@ -26,6 +26,7 @@ import (
 
 	"sigs.k8s.io/krew/cmd/krew/cmd/internal"
 	"sigs.k8s.io/krew/internal/index/indexscanner"
+	"sigs.k8s.io/krew/internal/index/validation"
 	"sigs.k8s.io/krew/internal/installation"
 	"sigs.k8s.io/krew/internal/pathutil"
 	"sigs.k8s.io/krew/pkg/index"
@@ -97,6 +98,10 @@ Remarks:
 
 			var install []pluginEntry
 			for _, name := range pluginNames {
+				if !validation.IsSafePluginName(name) {
+					return unsafePluginNameErr(name)
+				}
+
 				indexName, pluginName := pathutil.CanonicalPluginName(name)
 				plugin, err := indexscanner.LoadPluginByName(paths.IndexPluginsPath(indexName), pluginName)
 				if err != nil {
--- a/integration_test/index_test.go
+++ b/integration_test/index_test.go
@@ -138,7 +138,7 @@ func TestKrewIndexRemove_unsafe(t *testing.T) {
 	expected := "invalid index name"
 	cases := []string{"a/b", `a\b`, "../a", `..\a`}
 	for _, c := range cases {
-		b, err := test.Krew("index", "add", c, constants.IndexURI).Run()
+		b, err := test.Krew("index", "remove", c, constants.IndexURI).Run()
 		if err == nil {
 			t.Fatalf("%q: expected error", c)
 		} else if !strings.Contains(string(b), expected) {
--- a/integration_test/index_test.go
+++ b/integration_test/index_test.go
@@ -16,6 +16,7 @@ package integrationtest
 
 import (
 	"os"
+	"strings"
 	"testing"
 
 	"sigs.k8s.io/krew/pkg/constants"
@@ -42,6 +43,25 @@ func TestKrewIndexAdd(t *testing.T) {
 	}
 	if _, err := test.Krew("index", "add", "foo", test.TempDir().Path("index/"+constants.DefaultIndexName)).Run(); err == nil {
 		t.Fatal("expected adding same index to fail")
+	}
+}
+
+func TestKrewIndexAddUnsafe(t *testing.T) {
+	skipShort(t)
+	test, cleanup := NewTest(t)
+	defer cleanup()
+	test = test.WithEnv(constants.EnableMultiIndexSwitch, 1).WithIndex()
+
+	cases := []string{"a/b", `a\b`, "../a", `..\a`}
+	expected := "invalid index name"
+
+	for _, c := range cases {
+		b, err := test.Krew("index", "add", c, constants.IndexURI).Run()
+		if err == nil {
+			t.Fatalf("%q: expected error", c)
+		} else if !strings.Contains(string(b), expected) {
+			t.Fatalf("%q: output doesn't contain %q: %q", c, expected, string(b))
+		}
 	}
 }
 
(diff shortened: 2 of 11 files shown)


cmd/krew/cmd/install.go
```diff
@@ -97,6 +98,10 @@ Remarks:
 
 			var install []pluginEntry
 			for _, name := range pluginNames {
+				if !validation.IsSafePluginName(name) {
+					return unsafePluginNameErr(name)
+				}
```
corneliusweig, line 103: I'd feel better if we validate plugin names in batch upfront. One way we could achieve this is by changing the API to
```go
func IsSafePluginName(names ...string) bool { .. }
```

(same in several other places)

Also, it seems that `unsafePluginNameErr()` could be returned from this function (making it obsolete), so that we have instead:
```go
func CheckSafePluginName(names ...string) error { .. }
```

integration_test/uninstall_test.go
```diff
@@ -75,3 +76,30 @@ func TestKrewRemove_ManifestRemovedFromIndex(t *testing.T) {
...
+	test.WithEnv(constants.EnableMultiIndexSwitch, 1).WithIndex()
+	test.Krew("install", validPlugin).RunOrFailOutput()
+
+	cases := []string{
+		`../index/` + validPlugin,
+		`..\index\` + validPlugin,
+		`../default/` + validPlugin,
+		`..\default\` + validPlugin,
+		`../receipts/` + validPlugin,
+		`..\receipts\` + validPlugin,
```
corneliusweig, line 93: How is `/receipt/` different from `/default/`? Why is this needed?
Otherwise, let's remove it.

integration_test/upgrade_test.go
```diff
@@ -50,6 +50,31 @@ func TestKrewUpgrade(t *testing.T) {
...
+	defer cleanup()
+	test.WithEnv(constants.EnableMultiIndexSwitch, 1).WithIndex()
+
+	cases := []string{
+		`../index/` + validPlugin,
+		`..\index\` + validPlugin,
+		`../default/` + validPlugin,
+		`..\default\` + validPlugin,
+		`index-name/sub-directory/plugin-name`,
+	}
```
corneliusweig, line 65: The list of test cases is repeated 3 times (if my question about `receipt` applies). Should we extract this to a central place?

integration_test/index_test.go
```diff
@@ -45,6 +46,25 @@ func TestKrewIndexAdd(t *testing.T) {
...
+
+	cases := []string{"a/b", `a\b`, "../a", `..\a`}
+	expected := "invalid index name"
+
+	for _, c := range cases {
+		b, err := test.Krew("index", "add", c, constants.IndexURI).Run()
+		if err == nil {
+			t.Fatalf("%q: expected error", c)
+		} else if !strings.Contains(string(b), expected) {
+			t.Fatalf("%q: output doesn't contain %q: %q", c, expected, string(b))
```
chriskim06, line 63: Small nit: `"%q: output doesn't contain %q: %q"` -> `"%q: output doesn't contain '%q': %q"`. This way it doesn't read as `a/b: output doesn't contain invalid index name...` but as `a/b: output doesn't contain 'invalid index name'...`

corneliusweig:
Test failure seems legit:
```
You used fmt.Errorf; use pkg/errors.Errorf instead to preserve stack traces:
cmd/krew/cmd/uninstall.go:58:func unsafePluginNameErr(n string) error { return fmt.Errorf("plugin name %q not allowed", n) }
```

integration_test/index_test.go
```diff
@@ -109,6 +129,24 @@ func TestKrewIndexRemove_ok(t *testing.T) {
...
+func TestKrewIndexRemove_unsafe(t *testing.T) {
+	skipShort(t)
+	test, cleanup := NewTest(t)
+	test = test.WithEnv(constants.EnableMultiIndexSwitch, 1).WithIndex()
+	defer cleanup()
+
+	expected := "invalid index name"
+	cases := []string{"a/b", `a\b`, "../a", `..\a`}
+	for _, c := range cases {
+		b, err := test.Krew("index", "add", c, constants.IndexURI).Run()
```
chriskim06, line 141: I think this should be "remove"

integration_test/upgrade_test.go
chriskim06: Also the test unsafe functions are all relatively similar (some of the error messaging is slightly different). The cases, expected error string, and arguments to `test.Krew` are the only differences. I think this could be useful as a `testutil` function like `test.AssertFailureWithUnsafeName` (not really sure what a good name would be)

integration_test/uninstall_test.go
ahmetb: receipt/ is needed because we need to point this to a receipt file. and at the moment all these paths parseable as a Receipt

integration_test/upgrade_test.go
ahmetb: I think duplication in tests aren't as bad, and each test actually has a slightly different things it's looking at, so I'm not entirely sure if the de-duplication helps.

integration_test/index_test.go
ahmetb: %q always adds `"` around the string, so this is not an issue.

integration_test/index_test.go
ahmetb: yep good catch.

cmd/krew/cmd/install.go
ahmetb: yeah although there's another method in `validation.go` that's geared towards plugin authors (and therefore it prints a regexp) as it runs during manifest validation tests.

    return errors.Errorf("the plugin name %q is not allowed, must match %q", name, safePluginRegexp.String())

that's why I kept it like this.

integration_test/index_test.go
```diff
@@ -109,6 +129,24 @@ func TestKrewIndexRemove_ok(t *testing.T) {
...
+func TestKrewIndexRemove_unsafe(t *testing.T) {
+	skipShort(t)
+	test, cleanup := NewTest(t)
+	test = test.WithEnv(constants.EnableMultiIndexSwitch, 1).WithIndex()
+	defer cleanup()
+
+	expected := "invalid index name"
+	cases := []string{"a/b", `a\b`, "../a", `..\a`}
+	for _, c := range cases {
+		b, err := test.Krew("index", "remove", c, constants.IndexURI).Run()
```
chriskim06, line 141: This last parameter isn't used in the index remove command

integration_test/index_test.go
chriskim06: TIL :smile:

ahmetb:
Another round of reviews please. 🙏

integration_test/upgrade_test.go
```diff
@@ -50,6 +50,31 @@ func TestKrewUpgrade(t *testing.T) {
...
+
+	expectedErr := `not allowed`
+	for _, c := range cases {
+		b, err := test.Krew("upgrade", c).Run()
```
chriskim06, line 69: Looks like you'd need https://github.com/kubernetes-sigs/krew/pull/582 before you can do this.

ahmetb: yep see the PR description :)

chriskim06: Whoops didn't see that haha! LGTM, I'll hold off for @corneliusweig to review.

integration_test/upgrade_test.go
corneliusweig: It would be nice to have the unsafe names in a variable, so that it is easily re-used without the need to copy. But let's merge as-is and iterate from there.

cmd/krew/cmd/install.go
corneliusweig: I understand that you want to keep the API in `validation` consistent.

Maybe this helper can be put in `cmd/internal` instead? Just an idea. It's fine as it stands, but validating upfront would be even better IMO.

corneliusweig:
@ahmetb Can you take a final look and rebase? Then I'll lgtm.

ahmetb:
Rebased + test run https://github.com/ahmetb/krew/runs/559418218 actually has captured a legitimate bug. Posted a fix.

corneliusweig:
/lgtm
/approve
Thanks for finding and fixing this loophole!

PR #584, opened 2020-04-02
constants: change the variable name

ahmetb:
This change indicates the possibility of non-centralized plugin index. However,
this breaks the API (e.g. exported pkg/constants), so I'm not sure we should do
this.

/assign @corneliusweig


ahmetb:
ah yeah that version.go field should not have been changed.

Issue #536, opened 2020-03-05
Proposal: cut git dependency


ahmetb:
Through a pure Go implementation like https://github.com/src-d/go-git
we should be able to phase out shelling to git and therefore having external
dependencies.

Of course git will always be beteter and we don't rely on a lot of detailed
or exotic features (the only interesting one that comes to mind is
"git reset --hard @{upstream}" usage we have).

I might try and prototype this.

/kind proposal

corneliusweig:
This would make it easier to use `krew` in minimal environments, such as an alpine container. I would still be careful here. Let's avoid to tie our hands in the future.

@ferhatelmas Do you have some experience with this library? I saw that you contributed to it.

ahmetb:
> Let's avoid to tie our hands in the future.

Worst case, the old implementation would still be there (possibly behind a feature flag), and our git.go does a decent job of serving as a wrapper.

ahmetb:
I think one of the main concerns would be supporting custom credential helpers (for custom indexes).

For example, at Google, `sso://user/ahmetb/foo.git`  is a valid git repository, authenticated via `git-credential-helper-sso` auth plugin that's in PATH. I'm not seeing support for that in go-git. https://godoc.org/gopkg.in/src-d/go-git.v4/plumbing/transport/client

corneliusweig:
> Worst case, the old implementation would still be there (possibly behind a feature flag), and our git.go does a decent job of serving as a wrapper.

Yes, I would prefer that. Maybe just fallback to the library, if no `git` is found in `PATH`? Then Google-magic would work out of the box without a flag.

ahmetb:
Some more on go-git project health:
https://github.com/go-git/go-git/wiki/go-git-has-a-new-home!-v5-and-some-explanations

corneliusweig:
Ok.. so let's see if the project stays as active as before if the backing company doesn't exist anymore.

ferhatelmas:
@ahmetb @corneliusweig sorry I was late. I take a stub on this. 

Project is active but moved to https://github.com/go-git/go-git due to copyright.

/assign @ferhatelmas

PR #587, opened 2020-04-06
Add tests for gitutil

### LINKED ISSUE #538: Add unit tests for internal/gitutil pkg
This package is currently lacking tests. It should be fairly easy to implement some tests that reach 100% coverage in this package. This would also help #536.

/priority backlog
/help
/good-first-issue

naxhh:
Fixes #538 
Related issue: #536 

Provide tests for gitutil as discussed in #538 
Because of the nature of `EnsureCloned` and `EnsureUpdated` this tests are quite slow compared to the rest of the suite (around 25 seconds) maybe would be good to mark most of them to be avoided in short runs, or mock the Exec calls to avoid doing fetches.

Regarding `Exec` I see it only used in a test to do a `git init && git remote add origin {url}` maybe is better to expose a function that does that instead of exposing `Exec` i didn't perform tests on it since is used everywhere in the rest of the tested code.

Finally I noticed `EnsureUpdate` does the same as `EnsureClone` + the update part.  Should we keep both exposed?

### DIFF
--- a/internal/gitutil/git_test.go
+++ b/internal/gitutil/git_test.go
@@ -1,0 +1,213 @@
+// Copyright 2019 The Kubernetes Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package gitutil
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"sigs.k8s.io/krew/internal/testutil"
+)
+
+func assertRepoIsCloned(t *testing.T, repoPath string) {
+	t.Helper()
+
+	ok, err := IsGitCloned(repoPath)
+
+	if err != nil {
+		t.Errorf("expected to finish correctly: %s", err)
+	}
+
+	if ok != true {
+		t.Errorf("expected to return true after cloning the repo")
+	}
+}
+
+func TestEnsureClonedClonesARepository(t *testing.T) {
+	tempDir, cleanup := testutil.NewTempDir(t)
+	defer cleanup()
+
+	httpClonePath := tempDir.Path("krew-from-https")
+	localClonePath := tempDir.Path("krew-from-local-path")
+
+	if err := EnsureCloned("https://github.com/kubernetes-sigs/krew.git", httpClonePath); err != nil {
+		t.Errorf("http clone expected to finish correctly: %s", err)
+	}
+
+	if err := EnsureCloned(httpClonePath, localClonePath); err != nil {
+		t.Errorf("folder clone expected to finish correctly: %s", err)
+	}
+
+	assertRepoIsCloned(t, httpClonePath)
+	assertRepoIsCloned(t, localClonePath)
+}
+
+func TestEnsureClonedDoesntCloneAnExistingRepo(t *testing.T) {
+	tempDir, cleanup := testutil.NewTempDir(t)
+	defer cleanup()
+
+	if err := EnsureCloned("https://github.com/kubernetes-sigs/krew.git", tempDir.Root()); err != nil {
+		t.Errorf("clone expected to finish correctly: %s", err)
+	}
+
+	tempDir.Write("file", []byte("create a file to ensure is not overwriten in the next clone"))
+
+	if err := EnsureCloned("https://github.com/kubernetes-sigs/krew.git", tempDir.Root()); err != nil {
+		t.Errorf("second clone expected to finish correctly: %s", err)
+	}
+
+	_, err := os.Stat(tempDir.Path("file"))
+
+	if err != nil {
+		t.Error("expected to dont perform any operation in the second clone")
+	}
+}
+
+func TestIsGitClonedWhenPathDoesntExists(t *testing.T) {
+	tempDir, cleanup := testutil.NewTempDir(t)
+	defer cleanup()
+
+	ok, err := IsGitCloned(tempDir.Path("does-not-exist"))
+
+	if err != nil {
+		t.Errorf("expected to finish correctly: %s", err)
+	}
+
+	if ok != false {
+		t.Errorf("expected to return false on a non existing folder")
+	}
+}
+
+func TestIsGitClonedWhenIsFalse(t *testing.T) {
+	tempDir, cleanup := testutil.NewTempDir(t)
+	defer cleanup()
+
+	ok, err := IsGitCloned(tempDir.Root())
+
+	if err != nil {
+		t.Errorf("expected to finish correctly: %s", err)
+	}
+
+	if ok != false {
+		t.Errorf("expected to return false on a folder that's not a git repo")
+	}
+}
+
+func TestIsGitClonedWhenIsTrue(t *testing.T) {
+	tempDir, cleanup := testutil.NewTempDir(t)
+	defer cleanup()
+
+	if err := os.MkdirAll(tempDir.Path(".git"), os.ModePerm); err != nil {
+		t.Fatalf("cannot create directory %q: %s", filepath.Dir(tempDir.Path(".git")), err)
+	}
+
+	ok, err := IsGitCloned(tempDir.Root())
+
+	if err != nil {
+		t.Errorf("expected to finish correctly: %s", err)
+	}
+
+	if ok != true {
+		t.Errorf("expected to return true on a git repo")
+	}
+}
+
+func TestEnsureUpdatedClonesTheRepoIfDoesntExists(t *testing.T) {
+	tempDir, cleanup := testutil.NewTempDir(t)
+	defer cleanup()
+
+	if err := EnsureUpdated("https://github.com/kubernetes-sigs/krew.git", tempDir.Root()); err != nil {
+		t.Errorf("expected to finish correctly: %s", err)
+	}
+
+	assertRepoIsCloned(t, tempDir.Root())
+}
+
+func TestEnsureUpdatedUpdatesTheRepo(t *testing.T) {
+	tempDir, cleanup := testutil.NewTempDir(t)
+	defer cleanup()
+
+	if err := EnsureCloned("https://github.com/kubernetes-sigs/krew.git", tempDir.Root()); err != nil {
+		t.Errorf("first clone expected to finish correctly: %s", err)
+	}
+
+	lastCommitID, err := Exec(tempDir.Root(), "rev-parse", "HEAD")
+
+	if err != nil {
+		t.Errorf("get last commit expected to finish correctly: %s", err)
+	}
+
+	if _, err := Exec(tempDir.Root(), "reset", "--hard", "HEAD~1"); err != nil {
+		t.Errorf("commit reset expected to finish correctly: %s", err)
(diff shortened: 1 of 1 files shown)


ahmetb:
> Regarding `Exec` I see it only used in a test to do a `git init && git remote add origin {url}` maybe is better to expose a function that does that instead of exposing `Exec` i didn't perform tests on it since is used everywhere in the rest of the tested code.

Yeah don't worry about this.

internal/gitutil/git_test.go
```diff
@@ -0,0 +1,213 @@
...
+	"path/filepath"
+	"testing"
+
+	"sigs.k8s.io/krew/internal/testutil"
+)
+
+func assertRepoIsCloned(t *testing.T, repoPath string) {
+	t.Helper()
+
+	ok, err := IsGitCloned(repoPath)
```
ahmetb, line 28: ideally test code shouldn't depend on functionality that's being tested. 

in this case, IsGitCloned should have a test.

so please consider actually using ioutil.ReadDir and verify the repoPath has >2 files, one of them being `.git/`.

internal/gitutil/git_test.go
```diff
@@ -0,0 +1,213 @@
...
+		t.Errorf("expected to return true after cloning the repo")
+	}
+}
+
+func TestEnsureClonedClonesARepository(t *testing.T) {
+	tempDir, cleanup := testutil.NewTempDir(t)
+	defer cleanup()
+
+	httpClonePath := tempDir.Path("krew-from-https")
+	localClonePath := tempDir.Path("krew-from-local-path")
```
ahmetb, line 44: do you mind splitting into two tests?

internal/gitutil/git_test.go
```diff
@@ -0,0 +1,213 @@
...
+
+func TestEnsureClonedDoesntCloneAnExistingRepo(t *testing.T) {
+	tempDir, cleanup := testutil.NewTempDir(t)
+	defer cleanup()
+
+	if err := EnsureCloned("https://github.com/kubernetes-sigs/krew.git", tempDir.Root()); err != nil {
+		t.Errorf("clone expected to finish correctly: %s", err)
+	}
+
+	tempDir.Write("file", []byte("create a file to ensure is not overwriten in the next clone"))
```
ahmetb, line 66: need to handle err

internal/gitutil/git_test.go
```diff
@@ -0,0 +1,213 @@
...
+
+	if err := EnsureCloned("https://github.com/kubernetes-sigs/krew.git", tempDir.Root()); err != nil {
+		t.Errorf("second clone expected to finish correctly: %s", err)
+	}
+
+	_, err := os.Stat(tempDir.Path("file"))
+
```
ahmetb, line 73: do you mind removing unnecessary spaces, there are far too many :)

internal/gitutil/git_test.go
```diff
@@ -0,0 +1,213 @@
...
+	ok, err := IsGitCloned(tempDir.Root())
+
+	if err != nil {
+		t.Errorf("expected to finish correctly: %s", err)
+	}
+
+	if ok != true {
+		t.Errorf("expected to return true on a git repo")
+	}
+}
```
ahmetb, line 126: you can actually merge all these into 1 test case very easily. and merge `if-if` to `if-elseif`

internal/gitutil/git_test.go
```diff
@@ -0,0 +1,213 @@
...
+
+func TestEnsureUpdatedClonesTheRepoIfDoesntExists(t *testing.T) {
+	tempDir, cleanup := testutil.NewTempDir(t)
+	defer cleanup()
+
+	if err := EnsureUpdated("https://github.com/kubernetes-sigs/krew.git", tempDir.Root()); err != nil {
```
ahmetb, line 132: since this test is using the network, it's quite bad for a unit test (which is supposed to run fast).

if possible, we should be using a local directory as the remote repo. for that, we need to init a dir, maybe add a file and make 1 or 2 commits, then clone that dir.

ahmetb:
thanks for this work, it is SUPER appreciated.
coincidentally today I took a pass on removing `git` dependency from `gitutil` (excl. tests) using `go-git.v5` and the results look promising. 👍

ahmetb:
> Finally I noticed `EnsureUpdate` does the same as `EnsureClone` + the update part. Should we keep both exposed?

I think it depends on the usage outside the pkg. If it's not used outside the pkg, you can unexport.

internal/gitutil/git_test.go
naxhh: Sure. that sounds ok to me. I was doing it this way because i've specific tests to check if `IsGitCloned` worked

internal/gitutil/git_test.go
naxhh: Sure. I kept them together to don't make the process slower. but i'll split them

internal/gitutil/git_test.go
naxhh: the two first tests, sure I'll merge.

The last one has a custom `mkdirall` as part of the setup so I think merging them will make the code less clear? how would you merge this one with the others normally?

internal/gitutil/git_test.go
naxhh: That's why I was asking if I should mock the `Exec` call. to make them faster and don't depend on git and other stuff.

I can make all tests use a local git repo and only the one about https working to use the url if that sounds ok

naxhh:
> > Finally I noticed `EnsureUpdate` does the same as `EnsureClone` + the update part. Should we keep both exposed?
> 
> I think it depends on the usage outside the pkg. If it's not used outside the pkg, you can unexport.

It's being used so I'll keep them. Today or tomorrow I'll upload the fixes. thanks for your time.

naxhh:
Btw the ci failed on the linting part. but the error doesn't give me a hint if the problem is mine or is unrelated

> level=warning msg="[runner] Can't run linter goanalysis_metalinter: deadcode: analysis skipped: errors in package: [/home/runner/work/krew/krew/cmd/krew/main.go:20:2: could not import sigs.k8s.io/krew/cmd/krew/cmd (/home/runner/work/krew/krew/cmd/krew/cmd/search.go:50:21: IndexURI not declared by package constants)]"

ahmetb:
We changed the const to DefaultIndexUri. Not sure why the compilation hasn’t failed instead.

ahmetb:
@naxhh sorry it appears that a previous patch broke the code. I sent #590 to fix it. I still do not know why that PRs tests have not failed, this one’s currently mysterious.

ahmetb:
Please rebase on top of master now.

internal/gitutil/git_test.go
corneliusweig: Maybe you can also shell out to `git` to check if it is a repo. Then we don't even need to assume anything about the internals of git repositories.

Other than that, I think your approach has drawbacks as Ahmet pointed out, but it should be fine. After all, we can assume that `IsGitCloned` is pretty solid.

internal/gitutil/git_test.go
```diff
@@ -0,0 +1,213 @@
...
+	httpClonePath := tempDir.Path("krew-from-https")
+	localClonePath := tempDir.Path("krew-from-local-path")
+
+	if err := EnsureCloned("https://github.com/kubernetes-sigs/krew.git", httpClonePath); err != nil {
+		t.Errorf("http clone expected to finish correctly: %s", err)
+	}
+
+	if err := EnsureCloned(httpClonePath, localClonePath); err != nil {
+		t.Errorf("folder clone expected to finish correctly: %s", err)
+	}
```
corneliusweig, line 52: IMO one of the two cases is enough. Otherwise we are testing if git accepts HTTP URLs  or local file paths.

internal/gitutil/git_test.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+func TestEnsureClonedClonesARepository(t *testing.T) {
+	tempDir, cleanup := testutil.NewTempDir(t)
+	defer cleanup()
+
+	httpClonePath := tempDir.Path("krew-from-https")
+	localClonePath := tempDir.Path("krew-from-local-path")
+
+	if err := EnsureCloned("https://github.com/kubernetes-sigs/krew.git", httpClonePath); err != nil {
```
corneliusweig, line 46: Cloning krew is  not a good idea IMO. It is far too large to be cloned for an ordinary unit test.

We should really strive to not have to clone from remote too often. There are basically two ways to approach this:

1. Create a util similar to https://github.com/kubernetes-sigs/krew/blob/b2b8a40555208cf5ff2dcb017befca0edbb54480/integration_test/testutil_test.go#L265 (or extract it to a common place).
2. Set up a repo by shelling out to `git` and clone that repo like localClonePath.

It looks to me like 2) is more straightforward.

(same in other test cases)

internal/gitutil/git_test.go
```diff
@@ -0,0 +1,213 @@
...
+	tempDir, cleanup := testutil.NewTempDir(t)
+	defer cleanup()
+
+	if err := EnsureCloned("https://github.com/kubernetes-sigs/krew.git", tempDir.Root()); err != nil {
+		t.Errorf("clone expected to finish correctly: %s", err)
+	}
+
+	tempDir.Write("file", []byte("create a file to ensure is not overwriten in the next clone"))
+
+	if err := EnsureCloned("https://github.com/kubernetes-sigs/krew.git", tempDir.Root()); err != nil {
```
corneliusweig, line 68:
```suggestion
	if err := EnsureCloned("https://github.com/kubernetes-sigs/krew.git", tempDir.Root()); err == nil {
```
I would expect that the second invocation provokes an error. Are you sure this is correct?

internal/gitutil/git_test.go
corneliusweig, line 66: This is not needed to provoke an error in git. Consider removing it.

internal/gitutil/git_test.go
```diff
@@ -0,0 +1,213 @@
...
+
+	assertRepoIsCloned(t, httpClonePath)
+	assertRepoIsCloned(t, localClonePath)
+}
+
+func TestEnsureClonedDoesntCloneAnExistingRepo(t *testing.T) {
+	tempDir, cleanup := testutil.NewTempDir(t)
+	defer cleanup()
+
+	if err := EnsureCloned("https://github.com/kubernetes-sigs/krew.git", tempDir.Root()); err != nil {
```
corneliusweig, line 62: If you continue to use the same repo in all tests, please extract the URL to a package-local constant.

internal/gitutil/git_test.go
corneliusweig: Mocking `Exec` should not be necessary. I think it's fine to clone a local repository (see other comment above).
Also, I don't think that cloning from a remote repo is something that we should test in unit tests.

internal/gitutil/git_test.go
```diff
@@ -0,0 +1,213 @@
...
+		t.Errorf("first clone expected to finish correctly: %s", err)
+	}
+
+	lastCommitID, err := Exec(tempDir.Root(), "rev-parse", "HEAD")
+
+	if err != nil {
```
corneliusweig, line 149: Please remove the blank line. The error check really belongs to the previous line and should be grouped with it. (Same in other places)

internal/gitutil/git_test.go
naxhh: i'll double check but I think it just returned without an error.

ahmetb:
@naxhh Gently pinging, when you have time PTAL at the comments.

ahmetb:
@naxhh Ping again.

naxhh:
Sorry I've been busy changing jobs. I'll try to make some time and finish this

corneliusweig:
@naxhh Polite ping. Would be really great to have these tests.

PR #588, opened 2020-04-06
cmd: Update each index instead of just default

chriskim06:
Here's the first part to getting update to work with multiple indexes (second part to show updated/new plugins is in progress).

I extracted some of the new code that was added for search into the `namingutils.go` file. I'm not exactly sure it fits there, it was originally named `utils.go` where I think it could have made more sense. Let me know if this isn't the right place to put the functionality to get a list of index name/URLs. I also moved the helper function for creating an empty git repo to a method on `TempDir` to use in the new test.

Related issue: #483 

/area multi-index

### DIFF
--- a/cmd/krew/cmd/namingutils.go
+++ b/cmd/krew/cmd/namingutils.go
@@ -15,6 +15,12 @@
 package cmd
 
 import (
+	"os"
+
+	"github.com/pkg/errors"
+
+	"sigs.k8s.io/krew/internal/environment"
+	"sigs.k8s.io/krew/internal/index/indexoperations"
 	"sigs.k8s.io/krew/pkg/constants"
 	"sigs.k8s.io/krew/pkg/index"
 )
@@ -47,4 +53,24 @@ func canonicalName(p index.Plugin, indexName string) string {
 		indexName = constants.DefaultIndexName
 	}
 	return indexName + "/" + p.Name
+}
+
+// allIndexes returns a slice of index name and URL pairs
+func allIndexes(p environment.Paths) ([]indexoperations.Index, error) {
+	indexes := []indexoperations.Index{
+		{
+			Name: constants.DefaultIndexName,
+			URL:  constants.DefaultIndexURI,
+		},
+	}
+	if os.Getenv(constants.EnableMultiIndexSwitch) != "" {
+		out, err := indexoperations.ListIndexes(p)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to list plugin indexes available")
+		}
+		if len(out) != 0 {
+			indexes = out
+		}
+	}
+	return indexes, nil
 }
--- a/cmd/krew/cmd/namingutils.go
+++ b/cmd/krew/cmd/namingutils.go
@@ -15,12 +15,6 @@
 package cmd
 
 import (
-	"os"
-
-	"github.com/pkg/errors"
-
-	"sigs.k8s.io/krew/internal/environment"
-	"sigs.k8s.io/krew/internal/index/indexoperations"
 	"sigs.k8s.io/krew/pkg/constants"
 	"sigs.k8s.io/krew/pkg/index"
 )
@@ -53,24 +47,4 @@ func canonicalName(p index.Plugin, indexName string) string {
 		indexName = constants.DefaultIndexName
 	}
 	return indexName + "/" + p.Name
-}
-
-// allIndexes returns a slice of index name and URL pairs
-func allIndexes(p environment.Paths) ([]indexoperations.Index, error) {
-	indexes := []indexoperations.Index{
-		{
-			Name: constants.DefaultIndexName,
-			URL:  constants.DefaultIndexURI,
-		},
-	}
-	if os.Getenv(constants.EnableMultiIndexSwitch) != "" {
-		out, err := indexoperations.ListIndexes(p)
-		if err != nil {
-			return nil, errors.Wrapf(err, "failed to list plugin indexes available")
-		}
-		if len(out) != 0 {
-			indexes = out
-		}
-	}
-	return indexes, nil
 }
--- a/cmd/krew/cmd/update.go
+++ b/cmd/krew/cmd/update.go
@@ -113,12 +113,16 @@ func ensureIndexesUpdated(_ *cobra.Command, _ []string) error {
 	}
 
 	var failed []string
+	var returnErr error
 	for _, idx := range indexes {
 		indexPath := paths.IndexPath(idx.Name)
 		klog.V(1).Infof("Updating the local copy of plugin index (%s)", indexPath)
 		if err := gitutil.EnsureUpdated(idx.URL, indexPath); err != nil {
-			klog.V(1).Infof("error updating index %s: %s", idx.Name, err)
+			klog.Warningf("failed to update index %q: %v", idx.Name, err)
 			failed = append(failed, idx.Name)
+			if returnErr == nil {
+				returnErr = err
+			}
 		}
 	}
 
@@ -146,7 +150,7 @@ func ensureIndexesUpdated(_ *cobra.Command, _ []string) error {
 	showUpdatedPlugins(os.Stderr, preUpdateIndex, posUpdateIndex, installedPlugins)
 
 	if len(failed) != 0 {
-		return errors.Errorf("failed to update the following indexes: %s\n", strings.Join(failed, ", "))
+		return errors.Wrapf(returnErr, "failed to update the following indexes: %s\n", strings.Join(failed, ", "))
 	}
 	return nil
 }
--- a/cmd/krew/cmd/update.go
+++ b/cmd/krew/cmd/update.go
@@ -19,12 +19,15 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
+	"sync"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"k8s.io/klog"
 
 	"sigs.k8s.io/krew/internal/gitutil"
+	"sigs.k8s.io/krew/internal/index/indexoperations"
 	"sigs.k8s.io/krew/internal/index/indexscanner"
 	"sigs.k8s.io/krew/internal/installation"
 	"sigs.k8s.io/krew/pkg/constants"
@@ -43,7 +46,7 @@ plugin index from the internet.
 Remarks:
   You don't need to run this command: Running "krew update" or "krew upgrade"
   will silently run this command.`,
-	RunE: ensureIndexUpdated,
+	RunE: ensureIndexesUpdated,
 }
 
 func showFormattedPluginsInfo(out io.Writer, header string, plugins []string) {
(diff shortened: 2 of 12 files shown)


cmd/krew/cmd/namingutils.go
```diff
@@ -48,3 +54,23 @@ func canonicalName(p index.Plugin, indexName string) string {
...
+		out, err := indexoperations.ListIndexes(p)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to list plugin indexes available")
+		}
+		if len(out) != 0 {
+			indexes = out
+		}
+	}
+	return indexes, nil
+}
```
ahmetb, line 76: It might've been better if we extracted the feature gate into `indexoperations.ListIndexes` so we don't reimplement it here.

IIRC there's at least one more place we hardcode this:

```go
[]indexoperations.Index{
		{
			Name: constants.DefaultIndexName,
			URL:  constants.DefaultIndexURI,
		},
```

you should be able to find very easily.

ahmetb: yeah if you handle it in one place (for now, until we remove the feature gate saves the following +60 lines in `namingutils_test.go` for us, and reduces duplication)

internal/testutil/tempdir.go
```diff
@@ -83,3 +85,17 @@ func (td *TempDir) WriteYAML(file string, obj interface{}) *TempDir {
 	}
 	return td.Write(file, content)
 }
+
+func (td *TempDir) InitEmptyGitRepo(path, url string) {
```
ahmetb, line 89: if you search for `gitutil.Exec(... ,"init")` you're likely to find that we use this code in so many other places. would be good to consolidate.

I'm +1 for this method. But if we actually add a file + commit it, then the index can actually be `git clone`d. (But the method name needs to change obviously.)

ahmetb: I see you updated quite a few tests up there, maybe that's actually all of it.

cmd/krew/cmd/update.go
```diff
@@ -105,9 +105,16 @@ func showUpdatedPlugins(out io.Writer, preUpdate, posUpdate []index.Plugin, inst
 func ensureIndexUpdated(_ *cobra.Command, _ []string) error {
 	preUpdateIndex, _ := indexscanner.LoadPluginListFromFS(paths.IndexPluginsPath(constants.DefaultIndexName))
 
-	klog.V(1).Infof("Updating the local copy of plugin index (%s)", paths.IndexPath(constants.DefaultIndexName))
-	if err := gitutil.EnsureUpdated(constants.DefaultIndexURI, paths.IndexPath(constants.DefaultIndexName)); err != nil {
-		return errors.Wrap(err, "failed to update the local index")
+	indexes, err := allIndexes(paths)
+	if err != nil {
+		return err
```
ahmetb, line 110: wrap

cmd/krew/cmd/update.go
```diff
@@ -105,9 +105,16 @@ func showUpdatedPlugins(out io.Writer, preUpdate, posUpdate []index.Plugin, inst
 func ensureIndexUpdated(_ *cobra.Command, _ []string) error {
```
ahmetb, line 105: method needs rename? :)

cmd/krew/cmd/update.go
```diff
@@ -105,9 +105,16 @@ func showUpdatedPlugins(out io.Writer, preUpdate, posUpdate []index.Plugin, inst
...
-		return errors.Wrap(err, "failed to update the local index")
+	indexes, err := allIndexes(paths)
+	if err != nil {
+		return err
+	}
+	for _, idx := range indexes {
+		indexPath := paths.IndexPath(idx.Name)
+		klog.V(1).Infof("Updating the local copy of plugin index (%s)", indexPath)
+		if err := gitutil.EnsureUpdated(idx.URL, indexPath); err != nil {
+			return errors.Wrapf(err, "failed to update the local index %s", idx.Name)
```
ahmetb, line 116: potentially the most interesting part of the PR here:

Do we bail out on the first index failed to update ––OR–– do we go ahead and update all possible indexes, and collect the errors (or at least one of them) + report it at the end.

For example, today, `krew install ctx BAD_PLUGIN ns` will install ctx & ns, but the overall command will fail. This was intentionally designed for good UX.

With `krew install` and `krew upgrade` both calling `ensureIndexUpdated` every time before they do their job, I think a bad index (e.g. deleted repo) will render the majority of krew user surface unusable ––especially if that index has a name that's  alphabetically earlier than `"default"`.

Thoughts @corneliusweig ?

cmd/krew/cmd/namingutils.go
chriskim06: Ah good call, I didn't think about that.

chriskim06: So this was something that we talked about when going over the proposal:
> If no indexes are configured, a message is printed explaining how to add krew-index as default

I just want to make sure we're cool with that. The current functionality is that krew-index gets added by default, but once multi-index is out that means that users won't be able to `krew install ctx` right after install.

chriskim06: > You should hardcode that index to that method only when X_KREW_MULTI_INDEX is "not" set.

We decided before that `krew index list` after a fresh install should work and return an empty list. Do you want me to change that behavior and accompanying tests if we want it to return the hardcoded index when env var isn't set?

> How do we add the default=krew-index (once multi-index work is complete) is another discussion which I'm tracking at #566. (Right now we're auto-adding it if it's missing on install/update/upgrade) but this is a side-effect of old code, and it's not about this PR.

I brought it up because that was part of the proposal and it seemed like this would be the PR to take care of the help message when no indexes are configured. If that's not the case then what should the behavior be when the env var is set, but no indexes are configured?

ahmetb: > We decided before that krew index list after a fresh install should work and return an empty list. Do you want me to change that behavior and accompanying tests if we want it to return the hardcoded index when env var isn't set?

We should keep this behavior. I only wanted that we call indexoperations.ListIndexes always, but without the feature flag, it returns default index, with the feature flag it lists from filesystem.

I can make this change if you want me to.

> I brought it up because that was part of the proposal and it seemed like this would be the PR to take care of the help message when no indexes are configured

That can be a separate PR, let's keep this one small.

chriskim06: Ok I can take care of it here 👍

cmd/krew/cmd/update.go
corneliusweig: In the long run, we might want to update indexes in parallel. If doing this in parallel, it is no longer possible to fail on the first index update, because "first" is not a proper concept then. So to avoid behaviour change in the future, we should strive to update everything and collect errors.

Besides, it's just better UX IMO.

chriskim06:
I'm working on adding integration tests for this.

integration_test/update_test.go
```diff
@@ -49,6 +49,50 @@ func TestKrewUpdate(t *testing.T) {
...
+		t.Errorf("error removing plugins directory from index: %s", err)
+	}
+	test.Krew("update").RunOrFailOutput()
+	out, err := ioutil.ReadDir(paths.IndexPluginsPath("foo"))
+	if err != nil {
+		t.Errorf("error reading plugins directory: %s", err)
+	}
+	if len(out) == 0 {
+		t.Error("no plugins in index foo after update")
+	}
```
ahmetb, line 73: shouldn't you be also checking what's in `default` index now?
also you can use something like `krew search` here to find plugins in all indexes.

cmd/krew/cmd/update.go
```diff
@@ -102,13 +104,24 @@ func showUpdatedPlugins(out io.Writer, preUpdate, posUpdate []index.Plugin, inst
...
+	if err != nil {
+		return errors.Wrap(err, "failed to list indexes")
+	}
+
+	var failed []string
+	for _, idx := range indexes {
+		indexPath := paths.IndexPath(idx.Name)
+		klog.V(1).Infof("Updating the local copy of plugin index (%s)", indexPath)
+		if err := gitutil.EnsureUpdated(idx.URL, indexPath); err != nil {
+			klog.V(1).Infof("error updating index %s: %s", idx.Name, err)
```
ahmetb, line 120: you should probably do this as `klog.Warningf` otherwise the user will never know (unless they use -v) what the error is.

See how we handle multiple plugin install failures in `install.go`.

cmd/krew/cmd/update.go
```diff
@@ -132,6 +145,9 @@ func ensureIndexUpdated(_ *cobra.Command, _ []string) error {
 	// TODO(chriskim06) consider commenting this out when refactoring for custom indexes
 	showUpdatedPlugins(os.Stderr, preUpdateIndex, posUpdateIndex, installedPlugins)
 
+	if len(failed) != 0 {
+		return errors.Errorf("failed to update the following indexes: %s\n", strings.Join(failed, ", "))
```
ahmetb, line 149: Something else we do in install.go is that we collect the first error as `returnErr` and wrap+return it.

https://github.com/kubernetes-sigs/krew/blob/b2b8a40555208cf5ff2dcb017befca0edbb54480/cmd/krew/cmd/install.go#L161-L163

https://github.com/kubernetes-sigs/krew/blob/b2b8a40555208cf5ff2dcb017befca0edbb54480/cmd/krew/cmd/install.go#L178-L180

The reason for this is that when you wrap+return errors, when executed with `-v 1` or above, we the retain ability to print full stack trace (in this case, at least for 1 error which is good enough).

integration_test/update_test.go
chriskim06: Ya good call :+1:

corneliusweig: I think it should be enough to look for two plugin names in the output. For example, `ctx` and `foo/ctx`.

cmd/krew/cmd/update.go
```diff
@@ -132,6 +149,9 @@ func ensureIndexUpdated(_ *cobra.Command, _ []string) error {
 	// TODO(chriskim06) consider commenting this out when refactoring for custom indexes
 	showUpdatedPlugins(os.Stderr, preUpdateIndex, posUpdateIndex, installedPlugins)
```
ahmetb, line 151: I’ve realized this if (+counting) isnt necessary since errors.Wrap already handles nil err. Maybe remove?

integration_test/update_test.go
```diff
@@ -49,6 +49,51 @@ func TestKrewUpdate(t *testing.T) {
...
+	defer os.Unsetenv(constants.EnableMultiIndexSwitch)
+
+	paths := environment.NewPaths(test.Root())
+	test.TempDir().InitEmptyGitRepo(paths.IndexPath("foo"), "invalid-git")
+	out, err := test.Krew("update").Run()
+	if err == nil {
+		t.Error("expected update to fail")
+	}
+	if !strings.Contains(string(out), "failed to update the following indexes: foo") {
+		t.Error("expected index update to fail for foo")
```
ahmetb, line 93: Try doing this for debuggability:

%q doesn’t contain msg=%q

cmd/krew/cmd/update.go
chriskim06: Sure I can do that. Do you think it will look weird if this ends with the `return errors.Wrapf(...)`?

integration_test/update_test.go
chriskim06: What are the two strings in the example you gave? I'm guessing the output of `krew update` and `foo` but want to make sure.

cmd/krew/cmd/update.go
ahmetb: I think it’s fine. I just realized the check is redundant.

integration_test/update_test.go
ahmetb: Out & “failed to ...foo”

You can export latter into a variable.

cmd/krew/cmd/update.go
chriskim06: Ok I'll remove the check then.

ahmetb:
/lgtm
/approve
thanks!

PR #593, opened 2020-04-16
Show new/updated plugins list after index updates

ahmetb:
* Adjusted the showUpdatedPlugins logic to behave well with multi-index mode
* Adjusted the "Updated the local copy of plugin index" message for multi-index
  mode. (We won't show index name if it's 'default' index.)

Related issue: #566
/area multi-index
/assign @corneliusweig 
/assign @chriskim06

### DIFF
--- a/cmd/krew/cmd/update.go
+++ b/cmd/krew/cmd/update.go
@@ -30,7 +30,6 @@ import (
 	"sigs.k8s.io/krew/internal/index/indexscanner"
 	"sigs.k8s.io/krew/internal/installation"
 	"sigs.k8s.io/krew/pkg/constants"
-	"sigs.k8s.io/krew/pkg/index"
 )
 
 // updateCmd represents the update command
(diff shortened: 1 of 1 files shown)


cmd/krew/cmd/update.go
```diff
@@ -59,59 +58,77 @@ func showFormattedPluginsInfo(out io.Writer, header string, plugins []string) {
...
-
 		showFormattedPluginsInfo(out, "New plugins available", s)
 	}
 
 	if len(updatedPlugins) > 0 {
 		var s []string
 		for _, p := range updatedPlugins {
-			old := oldIndex[p.Name]
-			s = append(s, fmt.Sprintf("%s %s -> %s", p.Name, old.Spec.Version, p.Spec.Version))
+			old := oldIndexMap[canonicalName(p.p, p.indexName)]
```
chriskim06, line 98: Minor nit, but I think the keys for `oldIndexMap` could use `displayName`. Then the line below could be brought up and this could just be `oldIndexMap[name]`.

ahmetb: It's probably safer we just use canonicalName in internal parts of the code. displayName is only for displaying, and who knows may start colliding someday.

chriskim06:
Thanks for taking care of this! It LGTM, do you think it would be worth modifying [this](https://github.com/kubernetes-sigs/krew/blob/master/integration_test/update_test.go#L119) integration test to check for upgrade available for plugins from custom indexes or is that not needed?

ahmetb:
Yeah, I got lazy about tests. I will try to update.

cmd/krew/cmd/update.go
```diff
@@ -124,31 +141,27 @@ func ensureIndexesUpdated(_ *cobra.Command, _ []string) error {
...
-
-	fmt.Fprintln(os.Stderr, "Updated the local copy of plugin index.")
 
-	if len(preUpdateIndex) == 0 {
-		return nil
+		if os.Getenv(constants.EnableMultiIndexSwitch) == "" || isDefaultIndex(idx.Name) {
+			fmt.Fprintln(os.Stderr, "Updated the local copy of plugin index.")
+		} else {
+			fmt.Fprintf(os.Stderr, "Updated the local copy of plugin index %q.\n", idx.Name)
+		}
```
corneliusweig, line 149: nit: this message is shown even if the `EnusreUpdated()` call failed with an error.

corneliusweig:
/lgtm
/approve
/hold
In case you want to follow up in a different PR.

cmd/krew/cmd/update.go
ahmetb: Good catch!

ahmetb:
> nit: this message is shown even if the `EnusreUpdated()` call failed with an error.

So I fixed this bug and pushed a change. But I'm not touching the tests. That's mostly because I noticed we'd duplicate a lot of test code (as adding custom indexes takes a lot of code right now).

Ideally we need an utility like:

`NewTest(t).WithDefaultIndex().WithDefaultIndexMirroredAs("foo").WithCustomIndex(name, src). ...`

but since there's a feature gate around the multi-index features, I'm refraining from developing this pattern.

Issue #579, opened 2020-03-31
Krew uninstall behavior for multi-index


corneliusweig:
With multi-index, plugins can come from different indexes. For example, we can have 

* `foo/bar`, plugin `bar` from index `foo` and
* `baz/bar`, plugin `bar` from index `baz`

These two plugins cannot be installed at the same time, because the plugin name `bar` is the same in both cases, and krew does not allow to rename plugins during install. In fact, when installing plugin `bar`, it becomes (almost) oblivious of its index source.

This opens the question, how plugins should properly uninstalled.

**Option 1**
Ignore the index source altogether. Every plugin is uninstalled by its plain plugin name, for example

```bash
krew uninstall bar  # could refer to foo/bar  or  baz/bar alike
```

Pro:
* easy to implement
* users are used to refer to the plugin as `bar` when they use it

Con:
* the commands to install and _un_install are not consistent

**Option 2**
Uninstall goes by its canonical name. For example,

```bash
krew uninstall foo/bar  # fails when baz/bar was installed
krew uninstall bar      # fails bc ambiguous
```

Pro:
* good consistency between uninstall/install commands

Con:
* hard to implement
* hard to use, because users need to find out the actual source of the plugin (`krew info`). Here we could help the users by saying "Did you mean `krew uninstall baz/bar`?"--again, more complexity.

**Option 3**
Uninstall accepts canonical and short names and just ignores the index part. For example

```bash
krew uninstall foo/bar   # succeeds even if baz/bar was installed
krew uninstall bar   # succeeds for baz/bar and foo/bar alike
```

Pro:
* easy to implement
* good consistency between install/uninstall commands

Con:
* can be confusing when mixing `krew install foo/bar` with `krew uninstall baz/bar`, which succeeds.

/cc @ahmetb @chriskim06 
/area multi-index
/kind proposal

chriskim06:
#### Option 1
I think this doesn't require any change which is nice, but I agree that it could be confusing that install and uninstall would be different.

#### Option 2
I'm not sure that this would be too difficult to do. The receipt gets loaded in the uninstall code to verify that the plugin being uninstalled has been installed. I feel like it wouldn't take too many changes to check the index of the receipt and match it with what's being uninstalled.

Adding help text "did you mean `krew uninstall baz/bar`" definitely complicates it, but would be a nice usability feature.

#### Option 3
Agreed that consistency between install/uninstall is nice. I think it would be a little weird if someone does:
```
[...]

PR #594, opened 2020-04-16
cmd: modify upgrade to work with multiple indexes

chriskim06:
Upgrade was already refactored to support adding indexes to receipts so this was a pretty simple change + integration test. Upgrade works with:

- `kubectl krew upgrade`
- `kubectl krew upgrade plugin ... // plugin can be default/plugin or foo/plugin`

Also added warning text to uninstall when user tries to uninstall by canonical name.

Related issue: #566 
/area multi-index

### DIFF
--- a/cmd/krew/cmd/namingutils_test.go
+++ b/cmd/krew/cmd/namingutils_test.go
@@ -114,13 +114,28 @@ func Test_canonicalName(t *testing.T) {
 }
 
 func Test_isCanonicalName(t *testing.T) {
-	if isCanonicalName("foo") {
-		t.Error("expected foo to not be considered a canonical name")
+	tests := []struct {
+		arg      string
+		expected bool
+	}{
+		{"foo", false},
+		{"../index/foo", false},
+		{"index/foo", true},
+		{"", false},
+		{"0-0", false},
+		{"a/", false},
+		{"/b", false},
+		{"a-a/b-b", true},
+		{"a//b", false},
+		{"a / b", false},
+		{"a /b", false},
+		{"a/ b", false},
 	}
-	if isCanonicalName("../index/foo") {
-		t.Error("expected ../index/foo to not be considered a canonical name")
-	}
-	if !isCanonicalName("index/foo") {
-		t.Error("expected index/foo to be considered a canonical name")
+	for _, tt := range tests {
+		t.Run(tt.arg, func(t *testing.T) {
+			if isCanonicalName(tt.arg) != tt.expected {
+				t.Errorf("expected isCanonicalName(%q) to be %t", tt.arg, tt.expected)
+			}
+		})
 	}
 }
--- a/cmd/krew/cmd/namingutils_test.go
+++ b/cmd/krew/cmd/namingutils_test.go
@@ -112,3 +112,15 @@ func Test_canonicalName(t *testing.T) {
 		t.Errorf("expected=%q; got=%q", expected, got)
 	}
 }
+
+func Test_isCanonicalName(t *testing.T) {
+	if isCanonicalName("foo") {
+		t.Error("expected foo to not be considered a canonical name")
+	}
+	if isCanonicalName("../index/foo") {
+		t.Error("expected ../index/foo to not be considered a canonical name")
+	}
+	if !isCanonicalName("index/foo") {
+		t.Error("expected index/foo to be considered a canonical name")
+	}
+}
--- a/cmd/krew/cmd/uninstall.go
+++ b/cmd/krew/cmd/uninstall.go
@@ -41,8 +41,7 @@ Remarks:
 		for _, name := range args {
 			if isCanonicalName(name) {
 				return errors.New("uninstall command does not support INDEX/PLUGIN syntax; just specify PLUGIN")
-			}
-			if !validation.IsSafePluginName(name) {
+			} else if !validation.IsSafePluginName(name) {
 				return unsafePluginNameErr(name)
 			}
 			klog.V(4).Infof("Going to uninstall plugin %s\n", name)
--- a/cmd/krew/cmd/uninstall.go
+++ b/cmd/krew/cmd/uninstall.go
@@ -39,10 +39,10 @@ Remarks:
   Failure to uninstall a plugin will result in an error and exit immediately.`,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		for _, name := range args {
+			if isCanonicalName(name) {
+				return errors.New("uninstall command does not support INDEX/PLUGIN syntax; just specify PLUGIN")
+			}
 			if !validation.IsSafePluginName(name) {
-				if isCanonicalName(name) {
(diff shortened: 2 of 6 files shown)


ahmetb:
This is kinda like #579 (should we do `krew uninstall foo` or `krew uninstall myindex/foo`). In this case, only 1 plugin with a particular name can be installed at any time. So is `kubectl krew upgrade foo/plugin` syntax actually the right call here? (Note, this creates a bit asymmetry with `uninstall` command.)

chriskim06:
Oh right, I forgot that we decided on that in that issue. So just to be clear, we're going to do `krew upgrade plugin` to refer to either default/plugin or foo/plugin right?

ahmetb:
> Oh right, I forgot that we decided on that in that issue. So just to be clear, we're going to do `krew upgrade plugin` to refer to either default/plugin or foo/plugin right?

Yeah, whatever’s installed. Upgrade machinery should be primarily looking at the install receipts to figure out what that name resolves to.

Also, I was hoping to add a check to `uninstall` to remind people not to use `/` in the specified name. You can do it here as well.

chriskim06:
I kinda misread what you meant about uninstall having that warning. I for some reason thought you meant you wanted me to add that in this PR but I realize you meant to add the same type of warning to upgrade. Let me know if you want me to remove that from this PR.

ahmetb:
I think it’s fine to update. But I think we should return it as err not warning. That way we can also test it.

cmd/krew/cmd/upgrade.go
```diff
@@ -53,21 +54,30 @@ kubectl krew upgrade foo bar"`,
...
 				ignoreUpgraded = true
 				skipErrors = true
 			} else {
 				// Upgrade certain plugins
-				pluginNames = args
+				for _, arg := range args {
+					if !validation.IsSafePluginName(arg) {
+						if isCanonicalName(arg) {
+							return errors.New("upgrade command does not support INDEX/PLUGIN syntax; just specify PLUGIN")
+						}
```
corneliusweig, line 67: nit: move that right before the enclosing if

cmd/krew/cmd/upgrade.go
```diff
@@ -53,21 +54,30 @@ kubectl krew upgrade foo bar"`,
...
+						return unsafePluginNameErr(arg)
+					}
+					r, err := receipt.Load(paths.PluginInstallReceiptPath(arg))
+					if err != nil {
+						return errors.Wrapf(err, "receipt not found for %q", arg)
```
corneliusweig, line 72: nit: we don't know what the error is
```suggestion
						return errors.Wrapf(err, "read receipt %q", arg)
```

cmd/krew/cmd/uninstall.go
```diff
@@ -40,6 +40,9 @@ Remarks:
 	RunE: func(cmd *cobra.Command, args []string) error {
 		for _, name := range args {
 			if !validation.IsSafePluginName(name) {
+				if isCanonicalName(name) {
+					return errors.New("uninstall command does not support INDEX/PLUGIN syntax; just specify PLUGIN")
+				}
```
corneliusweig, line 45: nit: move in right before the enclosing if.

chriskim06:
Any final reviews would be much appreciated when you guys get some spare time 😄

integration_test/uninstall_test.go
```diff
@@ -47,6 +47,23 @@ func TestKrewUninstall(t *testing.T) {
 	}
 }
 
+func TestKrewUninstallPluginsFromCustomIndex(t *testing.T) {
```
ahmetb, line 50: thanks for adding these.

integration_test/uninstall_test.go
```diff
@@ -47,6 +47,23 @@ func TestKrewUninstall(t *testing.T) {
...
+	skipShort(t)
+
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	test.WithEnv(constants.EnableMultiIndexSwitch, 1).WithIndex()
+	test.Krew("index", "add", "foo", test.TempDir().Path("index/"+constants.DefaultIndexName)).RunOrFail()
+	test.Krew("install", "foo/"+validPlugin).RunOrFail()
+
```
ahmetb, line 59: I think you don't need this part of the test.

you can just run `uninstall a/b` without `b` being installed and do the

```
	if !strings.Contains(string(b), "INDEX/PLUGIN") {
		t.Error("expected warning about using canonical name to be in output")
	}

```

check you did in `TestKrewUpgradePluginsFromCustomIndex`

integration_test/upgrade_test.go
```diff
@@ -50,6 +51,38 @@ func TestKrewUpgrade(t *testing.T) {
...
+		t.Errorf("expected plugin foo/%s to be upgraded", validPlugin)
+	}
+
+	b, err := test.Krew("upgrade", "foo/"+validPlugin).Run()
+	if err == nil {
+		t.Error("expected error when using canonical name with upgrade")
+	}
+	if !strings.Contains(string(b), "INDEX/PLUGIN") {
+		t.Error("expected warning about using canonical name to be in output")
+	}
```
ahmetb, line 83: this is dope but do you mind extracting into a test named like `TestKrewUpgrade_CannotUseIndexSyntax`

You don't have to (1) add index (2) install plugin (3) specify foo/validPlugin (even x/y works) you should be getting the same error.

integration_test/uninstall_test.go
ahmetb: `TestKrewUninstal_CannotUseIndexSyntax` is a better name though. feel free to mix.

cmd/krew/cmd/namingutils_test.go
```diff
@@ -112,3 +112,15 @@ func Test_canonicalName(t *testing.T) {
 		t.Errorf("expected=%q; got=%q", expected, got)
 	}
 }
+
+func Test_isCanonicalName(t *testing.T) {
```
ahmetb, line 116: while you're at it, some good tests to add:

- ""
- "0-0"
- "a/"
- "/b"
- "a-a/b-b"
- "a//b"
- "a / b"
- "a/ b", "a /b"

and convert it to tabular test format.

cmd/krew/cmd/upgrade.go
```diff
@@ -53,21 +54,30 @@ kubectl krew upgrade foo bar"`,
...
 				ignoreUpgraded = true
 				skipErrors = true
 			} else {
 				// Upgrade certain plugins
-				pluginNames = args
+				for _, arg := range args {
+					if isCanonicalName(arg) {
+						return errors.New("upgrade command does not support INDEX/PLUGIN syntax; just specify PLUGIN")
+					}
+					if !validation.IsSafePluginName(arg) {
```
ahmetb, line 67: can do `else if` here

corneliusweig:
/approve
from my side. Sorry for being so sluggish to respond. Please do @ me, if you want quicker feedback 🙏

PR #595, opened 2020-04-20
multi-index: Add default index if none exists

ahmetb:
When user runs krew {install,update,upgrade} for the first time, if there are
no indexes, default index (with krew-index repo) is now automatically added.

Right now this code path only runs behind multi-index feature gate.

Related issue: #566
/assign @chriskim06
/area multi-index

### DIFF
--- a/cmd/krew/cmd/update.go
+++ b/cmd/krew/cmd/update.go
@@ -44,7 +44,7 @@ plugin index from the internet.
 Remarks:
   You don't need to run this command: Running "krew update" or "krew upgrade"
   will silently run this command.`,
-	RunE: ensureIndexesUpdated,
+	RunE: ensureIndexes,
 }
 
 func showFormattedPluginsInfo(out io.Writer, header string, plugins []string) {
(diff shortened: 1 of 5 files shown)


integration_test/index_test.go
```diff
@@ -171,3 +172,57 @@ func TestKrewIndexRemoveForce_nonExisting(t *testing.T) {
...
+	skipShort(t)
+	test, cleanup := NewTest(t)
+	test = test.WithEnv(constants.EnableMultiIndexSwitch, 1)
+	defer cleanup()
+
+	test.Krew("upgrade").RunOrFail()
+	ensureIndexListHasDefaultIndex(t, string(test.Krew("index", "list").RunOrFailOutput()))
+}
+
+func ensureIndexListHasDefaultIndex(t *testing.T, output string) {
```
chriskim06, line 223: Seems like you could just pass `*ITest` here as the param.

integration_test/index_test.go
```diff
@@ -171,3 +172,57 @@ func TestKrewIndexRemoveForce_nonExisting(t *testing.T) {
...
+	t.Helper()
+	if !regexp.MustCompile(`(?m)^default\b`).MatchString(output) {
```
chriskim06, line 225: Nit: isn't it better to put this at the package level so that the regular expression is only parsed once instead of for each time this function is called?

integration_test/index_test.go
ahmetb: 🤔 why? Only testing.T seems sufficient?

chriskim06: Ok, I only mentioned since the `string(test.Krew("index", "list").RunOrFailOutput())` is being passed to each one. I thought it might be better to put that in the `ensureIndexListHasDefaultIndex` function rather than relying on the caller to pass the desired output.

ahmetb: yeah got it now. I think that's fine. this way we can actually see the error from RunOrFail at the correct line number.

integration_test/index_test.go
ahmetb: since it's only part of integration tests, I figured it doesn't matter a whole lot.

cmd/krew/cmd/update.go
```diff
@@ -104,9 +104,38 @@ func showUpdatedPlugins(out io.Writer, preUpdate, posUpdate []index.Plugin, inst
...
+		return nil
+	}
 
+	klog.V(3).Infof("No index found, add default index.")
+	fmt.Fprintf(os.Stderr, "Adding \"default\" plugin index from %s.\n", constants.DefaultIndexURI)
+	return errors.Wrap(indexoperations.AddIndex(paths, constants.DefaultIndexName, constants.DefaultIndexURI),
+		"failed to add default plugin index in absence of no indexes")
+}
+
+// ensureDefaultIndexIfNoneExist iterates over all indexes and updates them
```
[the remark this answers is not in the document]
chriskim06: Hey just wanted to check in on this, it LGTM except this minor typo in the godoc.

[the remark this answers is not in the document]
ahmetb: Ah fixing now. I'll make sure cornelius reviews too.

ahmetb:
/hold
/assign @corneliusweig 
quick reviews appreciated.

ahmetb:
@corneliusweig please re-lgtm.

corneliusweig:
/lgtm
/approve
/hold cancel
Thanks for taking care of this!

PR #598, opened 2020-04-30
integration_test: Create helpers to initialize tests with custom indexes

### LINKED ISSUE #565: integration_test: Create helpers to initialize tests with custom indexes
Currently integration_test test cases for:

- `krew index`
- other krew cmds that test multi-index behavior

keep repeating the commands to create multiple indexes, such as:

```go
	localIndex := test.TempDir().Path("index/" + constants.DefaultIndexName)
	test.Krew("index", "add", "foo", localIndex).RunOrFail()
```

Right now, this is achieved by doing `test = NewTest(t).WithIndex()` and aliasing the cloned `default` index directory into another index.

So perhaps we need to:

- modify: WithIndex() → WithDefaultIndex()
- add: WithDefaultIndexAlias(name string) 

the last name could use improvement, like maybe WithCustomIndex.

This way the test author does not need to worry about where the custom index is coming from. Roughly related to #562.

/kind cleanup
/help
cc: @chriskim06 
/area multi-index

chriskim06:
Two changes here:
- `WithIndex` -> `WithDefaultIndex`
- added `WithCustomIndex` function that does the cloning from the default index

I wasn't really sure if `WithCustomIndex` should call `initializeIndex` if it hasn't been initialized yet. It would clean up the code so that people don't have to call `WithDefaultIndex().WithCustomIndex("foo")` but I was worried that the cloning of krew-index was a little too hidden this way.

If I'm overthinking it let me know and I can change it up so that `WithCustomIndex` initializes krew-index if it hasn't been initialized yet.

Fixes #565
Related issue: #566

/area multi-index

### DIFF
--- a/integration_test/testutil_test.go
+++ b/integration_test/testutil_test.go
@@ -190,7 +190,8 @@ func (it *ITest) WithDefaultIndex() *ITest {
 }
 
 // WithCustomIndexFromDefault initializes a new index by cloning the default index. WithDefaultIndex needs
-// to be called before this function.
+// to be called before this function. This is a helper function for working with custom indexes in the
+// integration tests so that developers don't need to alias the cloned default index each time.
 func (it *ITest) WithCustomIndexFromDefault(name string) *ITest {
 	// TODO(chriskim06) remove this once index migration happens
 	if !isMultiIndexEnabled(it.env) {
@@ -318,13 +319,12 @@ func (it *ITest) initializeIndex() {
 }
 
 func isMultiIndexEnabled(env []string) bool {
-	enabled := false
 	for _, e := range env {
 		if strings.Contains(e, constants.EnableMultiIndexSwitch) {
-			enabled = true
+			return true
 		}
 	}
-	return enabled
+	return false
 }
 
 func initFromGitClone(t *testing.T) ([]byte, error) {
--- a/integration_test/testutil_test.go
+++ b/integration_test/testutil_test.go
@@ -189,9 +189,9 @@ func (it *ITest) WithDefaultIndex() *ITest {
 	return it
 }
 
-// WithCustomIndex initializes a new index by cloning the default index. WithDefaultIndex needs to be called
-// before this function.
-func (it *ITest) WithCustomIndex(name string) *ITest {
+// WithCustomIndexFromDefault initializes a new index by cloning the default index. WithDefaultIndex needs
+// to be called before this function.
+func (it *ITest) WithCustomIndexFromDefault(name string) *ITest {
 	// TODO(chriskim06) remove this once index migration happens
 	if !isMultiIndexEnabled(it.env) {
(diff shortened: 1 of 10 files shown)


ahmetb:
I think this mostly LGTM, and I wouldn't mind if `WithCustomIndex()` failed with error if default index is not initialized.

Also WithCustomIndex() name isn't the best IMO, it doesn't indicate that it's created by mirroring the default index. I'm open to other proposals.

chriskim06:
How about `WithCustomIndexFromDefault`? Its kind of a wordy/long function name but I think it captures the intent behind it (new custom index "from" default index).

I thought I'd just throw out the most obvious correction to what I had to get your opinion on that, but I'll keep giving this some thought throughout the day.

ahmetb:
We (maintainers) read the tests so it's fine. it also now makes sense why  WithDefaultIndex is there :)

chriskim06:
Ok I'll change it to that in a bit

chriskim06:
Here's something a little strange. The integration tests took [75 seconds](https://github.com/kubernetes-sigs/krew/pull/598/checks?check_run_id=634716168#step:15:6) to run here, but the action that got launched in my fork only took [36 seconds](https://github.com/chriskim06/krew/runs/634715990?check_suite_focus=true#step:15:6).

ahmetb:
Could be anything. If it's one off, nothing to worry, could be throttling (by repo, by org, global). But if it repeats consistently after merged, then we should probably investigate.

integration_test/testutil_test.go
```diff
@@ -299,13 +310,21 @@ func (it *ITest) initializeIndex() {
...
+	enabled := false
+	for _, e := range env {
 		if strings.Contains(e, constants.EnableMultiIndexSwitch) {
-			if err := indexmigration.Migrate(environment.NewPaths(it.Root())); err != nil {
-				it.t.Fatalf("error migrating index: %s", err)
-			}
+			enabled = true
 		}
 	}
+	return enabled
```
corneliusweig, line 327:
```suggestion
			return true
		}
	}
	return false
```

integration_test/testutil_test.go
```diff
@@ -183,12 +183,23 @@ func (it *ITest) Root() string {
...
-// WithIndex initializes the index with the actual krew-index from github/kubernetes-sigs/krew-index.
-func (it *ITest) WithIndex() *ITest {
+// WithDefaultIndex initializes the index with the actual krew-index from github/kubernetes-sigs/krew-index.
+func (it *ITest) WithDefaultIndex() *ITest {
 	it.initializeIndex()
 	return it
 }
 
+// WithCustomIndexFromDefault initializes a new index by cloning the default index. WithDefaultIndex needs
+// to be called before this function.
```
corneliusweig, line 193: I think it's worth explaining _why_ we want this testutil in the comment. This is another good reason to move this logic into a function, because it gives us a good place where we can document this.

ahmetb:
/lgtm
/approve
Thank you. 🙏

Issue #600, opened 2020-05-06
Index directory layout migration


chriskim06:
Currently there is a `krew system index-upgrade` command to migrate users to the new directory structure needed to support multiple indexes. It's been mentioned before about potentially making this migration happen automatically.

Both options seem fine to me. The receipts migration used a subcommand to perform the migration but I also think it might be more user friendly to have the index migration happen under the hood. @ahmetb @corneliusweig what are your thoughts?

/area multi-index

ahmetb:
I'm currently indifferent. We haven't heard of many complaints about the explicit migration in the past, right @corneliusweig ? My main concern is people leaping forward 3 versions at a time (think: they don't have `indexmigration` yet). 

So we need to probably manually test direct migration from 0.2 to latest to see if it's gracefully migrated (either explicitly or implicitly).

chriskim06:
I noticed a bug when I tried the update from 0.2 to master. Here's what I did:
- downloaded 0.2 version
- installed a plugin
- built master and tried running `krew list`
- prompted to run `receipts-upgrade`
- ran `receipts-upgrade` and got prompted to run `index-upgrade`
- ran `index-upgrade` and got prompted to run `receipts-upgrade`
- neither migration happened (although receipts/ dir was created)

Here's the output I saw after using the build from master (I kept both binaries in this folder so that's why this one has `-master` appended to the end):
```
$ KREW_ROOT='/home/kimc/oldkrew' X_KREW_ENABLE_MULTI_INDEX=1 ./krew-linux_amd64-master list
WARNING: To be able to run kubectl plugins, you need to add
the following to your ~/.bash_profile or ~/.bashrc:

    export PATH="${PATH}:${HOME}/.krew/bin"

and restart your shell.

You need to perform a migration to continue using krew.
Please run `kubectl krew system receipts-upgrade`
F0508 16:48:41.225263   27413 root.go:79] krew home outdated

$ KREW_ROOT='/home/kimc/oldkrew' X_KREW_ENABLE_MULTI_INDEX=1 ./krew-linux_amd64-master system receipts-upgrade
WARNING: To be able to run kubectl plugins, you need to add
the following to your ~/.bash_profile or ~/.bashrc:

    export PATH="${PATH}:${HOME}/.krew/bin"

and restart your shell.

You need to perform a migration to continue using krew.
Please run `kubectl krew system index-upgrade`
F0508 16:48:47.981024   27449 root.go:79] krew index outdated

[...]

PR #603, opened 2020-05-30
Remove receipts migration

chriskim06:
This removes the receiptsmigration code like we discussed in the related issue. I'm leaving the `system.go` file for now and leaving the integration test with the same name, but I might end up changing it in a follow up PR when I switch the index migration to be automatic.

Related issue: #600


cmd/krew/cmd/root.go
```diff
@@ -141,8 +141,12 @@ func preRun(cmd *cobra.Command, _ []string) error {
 	if err != nil {
 		return err
 	}
-	if !isMigrated && cmd.Use != "receipts-upgrade" {
-		fmt.Fprintln(os.Stderr, "You need to perform a migration to continue using krew.\nPlease run `kubectl krew system receipts-upgrade`")
+	if !isMigrated {
+		fmt.Fprintln(os.Stderr, `This version of Krew is not supported anymore. Please manually migrate:
+1. Backup plugin list: kubectl krew list > backup.txt
```
corneliusweig, line 146: So here's the problem: `krew list` will not work, because the krew we're running here is a new one, but it is faced with an old index. Instead of listing plugins, `krew list` will simply produce this error message and bail out. That's the reason why I was advocating for a unified migration.

The instructions here cannot rely on any `krew` command and must rely on `bash` (BSD flavor, so that OSX works) and PowerShell or other Windows instructions.

cmd/krew/cmd/system.go
```diff
@@ -72,6 +54,5 @@ func init() {
 	if _, ok := os.LookupEnv(constants.EnableMultiIndexSwitch); ok {
 		systemCmd.AddCommand(indexUpgradeCmd)
 	}
-	systemCmd.AddCommand(receiptsUpgradeCmd)
 	rootCmd.AddCommand(systemCmd)
```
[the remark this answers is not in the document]
chriskim06: Ya I was planning on removing it in the follow up PR where I make the index migration automatic.

cmd/krew/cmd/root.go
chriskim06: Ah good catch, I can change this. I can come up with something for bash, but I don't really have any experience in windows :sweat_smile:. I'll try to figure out what it would look like there, but do you or @ahmetb know what the windows equivalent of `ls -d $KREW_HOME/store/ > plugins.tmp`/`kubectl krew install < plugins.tmp` would be?

internal/receiptsmigration/migration.go
```diff
@@ -16,25 +16,9 @@
 package receiptsmigration
```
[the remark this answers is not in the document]
chriskim06: I think having `Done` is useful to be able to detect when to guide 0.2 users on how to manually reinstall krew/plugins right? That's the reason I left this in.

cmd/krew/cmd/root.go
ahmetb: Nice catch.
I'd say don't bother with this and just talk about:

1. uninstall...
2. reinstall...
3. install the plugins you used

(don't mention backups).

cmd/krew/main.go
```diff
@@ -22,5 +22,5 @@ import (
 
 func main() {
 	cmd.Execute()
-	defer klog.Flush()
+	klog.Flush()
```
ahmetb, line 25: 🤔 let's move this 1 above maybe and keep as defer.

there might be some os.Exit/panics causing klog.Flush to not be called

chriskim06: Sounds good. For some reason the linter started complaining about the defer being at the end of the function
```
$ hack/run-lint.sh
cmd/krew/main.go:25:2: unnecessaryDefer: defer klog.Flush() is placed just before return (gocritic)
        defer klog.Flush()
```

I'll do as you suggested :+1:

PR #607, opened 2020-06-05
Automatic index migration

### LINKED ISSUE #600: Index directory layout migration

chriskim06:
- replace the part in `root.go` that previously informed the user that they needed to migrate with the actual `indexmigration.Migrate` function
- add integration tests (1 for checking auto migration happens and another for the only scenario I can think of where it doesn't happen)

Fixes #600 
Related issue: #566


cmd/krew/cmd/root.go
```diff
@@ -154,9 +154,10 @@ func preRun(cmd *cobra.Command, _ []string) error {
 		if err != nil {
 			return errors.Wrap(err, "error getting file info")
 		}
-		if !isMigrated && cmd.Use != "index-upgrade" {
-			fmt.Fprintln(os.Stderr, "You need to perform a migration to continue using krew.\nPlease run `kubectl krew system index-upgrade`")
-			return errors.New("krew index outdated")
+		if !isMigrated {
```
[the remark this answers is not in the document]
chriskim06: Ah good catch!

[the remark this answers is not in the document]
chriskim06: Should this logging be leveled or always displayed to the user? The confirmation at the end was displayed to the user back when this was a manual process through `system index-upgrade` but now that its automatic should this happen at some log level?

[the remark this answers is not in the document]
ahmetb: I think we can show  the one-time operations like these to the user (after all they would be shown only "once"):

- migrating krew index layout 
- migration completed

internal/indexmigration/migration.go
```diff
@@ -36,14 +36,16 @@ func Done(paths environment.Paths) (bool, error) {
...
 // Migrate removes the index directory and then clones krew-index to the new default index path.
 func Migrate(paths environment.Paths) error {
-	isMigrated, err := Done(paths)
+	isMigrated, err := done(paths)
 	if err != nil {
 		return errors.Wrap(err, "failed to check if index migration is complete")
 	}
 	if isMigrated {
-		klog.V(2).Infoln("Already migrated")
+		klog.V(2).Infoln("already migrated")
```
corneliusweig, line 44: IMO, log messages can be capitalized and end with a dot. Errors on the other hand should not, because they can be wrapped and then it looks weird.

cmd/krew/cmd/root.go
```diff
@@ -150,13 +150,8 @@ func preRun(cmd *cobra.Command, _ []string) error {
...
 	if _, ok := os.LookupEnv(constants.EnableMultiIndexSwitch); ok {
-		isMigrated, err = indexmigration.Done(paths)
-		if err != nil {
-			return errors.Wrap(err, "error getting file info")
-		}
-		if !isMigrated && cmd.Use != "index-upgrade" {
-			fmt.Fprintln(os.Stderr, "You need to perform a migration to continue using krew.\nPlease run `kubectl krew system index-upgrade`")
-			return errors.New("krew index outdated")
+		if err := indexmigration.Migrate(paths); err != nil {
+			return errors.Wrap(err, "failed to automatically migrate index")
```
corneliusweig, line 154: This error is no longer accurate. The index could already have been migrated, but the check failed. In this particular situation, I think it's best not to wrap at all, because the errors from `Migrate` are already descriptive. If you feel strongly about wrapping, I'd suggest something like
```suggestion
			return errors.Wrap(err, "index initialization")
```

integration_test/migration_test.go
```diff
@@ -19,8 +19,41 @@ import (
...
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	test.WithEnv(constants.EnableMultiIndexSwitch, 1).WithDefaultIndex()
+	prepareOldIndexLayout(test)
+
+	// any command here should cause the index migration to occur
+	test.Krew("index", "list").RunOrFail()
+	if !isIndexMigrated(test) {
+		t.Error("output should include the default index after migration")
```
corneliusweig, line 38:
```suggestion
		t.Error("index should have been auto-migrated")
```

integration_test/migration_test.go
```diff
@@ -19,8 +19,41 @@ import (
...
+	skipShort(t)
+
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	test.WithEnv(constants.EnableMultiIndexSwitch, 1).WithDefaultIndex()
+	prepareOldIndexLayout(test)
+
+	test.Krew().RunOrFail()
+	if isIndexMigrated(test) {
```
corneliusweig, line 52: Is that behaviour something we really require or just a coincidence of the implementation? If the latter is the case, consider removing the test, because it doesn't test a relevant behaviour.

chriskim06: I think just a coincidence of the implementation (cobra doesn't seem to execute the prerun in the case of no subcommand or `krew -h`). It was something I noticed when I was testing the migration locally and wasn't really sure if I should add a test for or not. I can remove this if it seems unnecessary.

internal/indexmigration/migration.go
chriskim06: Ah ok I can change it, I misremembered reading that log messages shouldn't be capitalized in addition to error messages.

cmd/krew/cmd/root.go
```diff
@@ -150,13 +150,8 @@ func preRun(cmd *cobra.Command, _ []string) error {
...
 	if _, ok := os.LookupEnv(constants.EnableMultiIndexSwitch); ok {
-		isMigrated, err = indexmigration.Done(paths)
-		if err != nil {
-			return errors.Wrap(err, "error getting file info")
-		}
-		if !isMigrated && cmd.Use != "index-upgrade" {
-			fmt.Fprintln(os.Stderr, "You need to perform a migration to continue using krew.\nPlease run `kubectl krew system index-upgrade`")
-			return errors.New("krew index outdated")
+		if err := indexmigration.Migrate(paths); err != nil {
+			return err
```
ahmetb, line 154: this err needs to be wrapped. I don't see any indication of what was being done for err messages inside Migrate() (saying "migration failed"). So it's good to add that here.

integration_test/migration_test.go
```diff
@@ -42,6 +60,27 @@ func TestKrewUnsupportedVersion(t *testing.T) {
 	}
 }
 
+func isIndexMigrated(it *ITest) bool {
+	indexPath := it.TempDir().Path("index/default")
+	_, err := os.Stat(indexPath)
+	return err == nil
+}
+
+func prepareOldIndexLayout(it *ITest) {
```
ahmetb, line 69: is this something we should eventually delete?
if so please add a TODO (e.g. // TODO remove when testing indexmigration is no longer necessary).

internal/indexmigration/migration.go
```diff
@@ -36,14 +36,16 @@ func Done(paths environment.Paths) (bool, error) {
...
 // Migrate removes the index directory and then clones krew-index to the new default index path.
 func Migrate(paths environment.Paths) error {
-	isMigrated, err := Done(paths)
+	isMigrated, err := done(paths)
 	if err != nil {
 		return errors.Wrap(err, "failed to check if index migration is complete")
 	}
 	if isMigrated {
-		klog.V(2).Infoln("Already migrated")
+		klog.V(2).Infoln("Already migrated.")
```
ahmetb, line 44: so imagine every time krew runs in -v≥2, this will show up in the output.
but it's missing context:

- "Already migrated" → what’s migrated
- when is it started (there's no "checking index migration" msg)

ahmetb:
/approve
minor issues here, too.

cmd/krew/cmd/root.go
corneliusweig: I don't think we should say "migration failed", because the index could already have been migrated, but the check failed. Wrapping with a message which communicates this ambiguity would be fine though.

Maybe the function is also not named well. It checks and maybe migrates instead of just migrate.

chriskim06: What do you guys think about moving the migration needed check out of `Migrate` then? I think that would solve these issues.

corneliusweig: Yes, let's separate the check from the migration. That will make it easier to understand what is actually happening. :+1:

PR #611, opened 2020-06-08
Add common editor directories into gitignore

ferhatelmas:
idea/vs code according to https://blog.golang.org/survey2019-results

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,7 @@
 
 out/
 build/
+
+# Editors
+.idea/
+.vscode/


ahmetb:
Ideally people should learn to use global gitignore files on their dev machines. :)

I don’t think it belong here given we only have a handful number of maintainers. L

Also probably why github’s own dotfiles repo doesn’t have these as they should be solved on local level. A duplicated copy of IDE ignores can’t cope with this ever growing list.

ferhatelmas:
:+1: Fair enough.

Issue #576, opened 2020-03-31
Proposal: trust model around custom plugin indexes


ahmetb:
Related: #483 

Currently every plugin installed/upgraded from krew-index comes with a security warning:

```
WARNING: You installed plugin "ns" from the krew-index plugin repository.
   These plugins are not audited for security by the Krew maintainers.
   Run them at your own risk.
```

The reason we added this is because users must understand that they are downloading potentially unvetted binaries.

With the fact that multiple indexes are now happening, this opens several possibilities to rethink this experience, given:

- adding a plugin index is an explicit operation for the user (or IT admin)
- default index (krew-index) becomes just another index that user (or IT admin) can remove

## Proposal 1

- Trust all "custom" indexes implicitly.
  - do not print security warning during install
  - (maybe) add a warning while the index is added via "krew index add" command
- `default` index (i.e. krew-index) would still print installation warnings

  1. krew-index is still what 90% of our users will be using
  2. krew-index is added automatically (though, we haven't discussed how that looks like in a multi-index mode, but it's likely we don't want explicit user interaction as that'd break many user journeys)

Notes:

- does a custom treatment of "default" index
- what if "default" isn't actually pointing to krew-index repository?

## Proposal 2

Trust "all" indexes explicitly on the first use.

- First time someone tries to install a plugin through "krew install", we present an interactive (Y/n?) consent with the security message.
- The "trust" action is stored on disk (likely through a file indicator)
- Successive "krew install" commands using that index don't cause security messages.

Notes:

- Most first-time users of krew will see an interactive warning, once (as they're installing via krew-index)
  - Does an "interactive prompt" actually achieve something –or are we just covering our back with this warning? If the user meant to run a command, they are very likely to say "no" –but they should be aware of the risks.
- Is trusting an index _once_ a good model? e.g. users might forget the risks several months after acknowledging the centralized krew-index
- How does it work non-interactively (e.g. `krew install < myplugins.txt`)? (Likely we'd need to fail and add a --trust/--quiet flag to install which I'm not super eager about.)
- Introduces new files and logic to store this "trust".
  - The logic must be able to handle when index is removed and re-added (potentially with a new git remote)
- How does the user revoke "trust" action? (I don't want to add cmds or flags about _trust_.)
- No special treatment of krew-index (existing users also go through this)

/kind proposal
/area multi-index
/priority important-soon
/cc @corneliusweig 
/cc @chriskim06

corneliusweig:
[...]

PR #616, opened 2020-06-23
Print security notice when index is added

### LINKED ISSUE #615: Update security warning for custom indexes
Right now a security warning is printed anytime a plugin is installed/upgrade. This should be changed in the case of custom indexes to:
- the warning should be printed when the index is added
- plugin install/upgrade should only print the warning if the plugin is from the default index

Related to: #576 

/area multi-index

chriskim06:
Right now I have the warning when the index is added as:
```
WARNING: You have added a new index from "/home/kimc/playground/temp"
The plugins in this index are not audited for security by the Krew maintainers.
Install them at your own risk.
```

let me know if you guys think I should change this.

Fixes #615 
Related issue: #576

/area multi-index

### DIFF
--- a/cmd/krew/cmd/index.go
+++ b/cmd/krew/cmd/index.go
@@ -81,7 +81,8 @@ var indexAddCmd = &cobra.Command{
 		}
 		internal.PrintWarning(os.Stderr, `You have added a new index from %q
 The plugins in this index are not audited for security by the Krew maintainers.
-Install them at your own risk.`+"\n", args[1])
+Install them at your own risk.
+`, args[1])
 		return nil
 	},
 }
--- a/cmd/krew/cmd/index.go
+++ b/cmd/krew/cmd/index.go
@@ -76,12 +76,13 @@ var indexAddCmd = &cobra.Command{
 			return errInvalidIndexName
 		}
 		err := indexoperations.AddIndex(paths, name, args[1])
-		if err == nil {
-			internal.PrintWarning(os.Stderr, `You have added a new index from %q
+		if err != nil {
+			return err
+		}
+		internal.PrintWarning(os.Stderr, `You have added a new index from %q
 The plugins in this index are not audited for security by the Krew maintainers.
 Run them at your own risk.`+"\n", args[1])
-		}
-		return err
+		return nil
 	},
 }
 
--- a/cmd/krew/cmd/index.go
+++ b/cmd/krew/cmd/index.go
@@ -81,7 +81,7 @@ var indexAddCmd = &cobra.Command{
 		}
 		internal.PrintWarning(os.Stderr, `You have added a new index from %q
 The plugins in this index are not audited for security by the Krew maintainers.
-Run them at your own risk.`+"\n", args[1])
+Install them at your own risk.`+"\n", args[1])
 		return nil
 	},
 }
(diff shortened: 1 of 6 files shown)


ahmetb:
I think I'm good with this in principle.
/approve
/assign @corneliusweig

cmd/krew/cmd/index.go
```diff
@@ -86,7 +92,7 @@ var indexDeleteCmd = &cobra.Command{
 
 It is only safe to remove indexes without installed plugins. Removing an index
 while there are plugins installed will result in an error, unless the --force
-option is used ( not recommended).`,
+option is used (not recommended).`,
```
ahmetb, line 95: adding a TestKrewInstallDoesntShowSecurityWarningForCustomIndex would be nice.

cmd/krew/cmd/upgrade.go
```diff
@@ -104,7 +105,9 @@ kubectl krew upgrade foo bar"`,
 					return errors.Wrapf(err, "failed to upgrade plugin %q", pluginDisplayName)
 				}
 				fmt.Fprintf(os.Stderr, "Upgraded plugin: %s\n", pluginDisplayName)
-				internal.PrintSecurityNotice(plugin.Name)
+				if indexName == constants.DefaultIndexName {
```
ahmetb, line 108: TestKrewInstallDoesntShowSecurityWarningForCustomIndex

cmd/krew/cmd/index.go
```diff
@@ -75,7 +75,13 @@ var indexAddCmd = &cobra.Command{
 		if !indexoperations.IsValidIndexName(name) {
 			return errInvalidIndexName
 		}
-		return indexoperations.AddIndex(paths, name, args[1])
+		err := indexoperations.AddIndex(paths, name, args[1])
+		if err == nil {
```
ahmetb, line 79: I think `if err != nil { return err}`  then printing this outside the if-block would be better.

also worth adding tests for (just to check WARNING: exists)

cmd/krew/cmd/index.go
```diff
@@ -75,7 +75,14 @@ var indexAddCmd = &cobra.Command{
...
 			return errInvalidIndexName
 		}
-		return indexoperations.AddIndex(paths, name, args[1])
+		err := indexoperations.AddIndex(paths, name, args[1])
+		if err != nil {
+			return err
+		}
+		internal.PrintWarning(os.Stderr, `You have added a new index from %q
+The plugins in this index are not audited for security by the Krew maintainers.
+Run them at your own risk.`+"\n", args[1])
```
ahmetb, line 84: It might be better to say `Install them at your own risk.` for this particular msg.

integration_test/index_test.go
```diff
@@ -66,6 +66,19 @@ func TestKrewIndexAddUnsafe(t *testing.T) {
...
 
+func TestKrewIndexAddShowsSecurityWarning(t *testing.T) {
+	skipShort(t)
+
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	test.WithEnv(constants.EnableMultiIndexSwitch, 1).WithDefaultIndex()
+	out := string(test.Krew("index", "add", "foo", test.TempDir().Path("index/"+constants.DefaultIndexName)).RunOrFailOutput())
+	if !strings.Contains(out, "WARNING") {
```
corneliusweig, line 77: Can you also check for a more specific message? Just in case when `add` prints two warnings, we want to make sure the test asserts on the correct one.

cmd/krew/cmd/upgrade.go
corneliusweig: +1
Even if it looks redundant, it is an important documentation of our intended logic.

ahmetb:
/lgtm
Thanks!

PR #618, opened 2020-07-01
site: custom index docs

chriskim06:
Let me know what you guys think. Also what do I need to do to mark this as draft?

Related issue: #566 

/area multi-index
/area docs


ahmetb:
I'll make a pass by adding commits, stay tuned.

ahmetb:
Just did a pass by:

- merge 2 user guides into 1
- remove some cmd outputs that weren't useful
- rewrite some paragraphs
- change titles to be task oriented "Using Custom Plugin Indexes" and "Hosting Custom Plugin Indexes"
- emphasis on calling them "Custom Plugin Indexes" and not "custom indexes" which causes a new term to be coined.

ahmetb:
Actually could not push the commit. @chriskim06 do you mind choosing "allow maintainers to push additional commits" on the right?

ahmetb:
Preview:
- https://5efc0af6673a52000884afb5--kubernetes-sigs-krew.netlify.app/docs/user-guide/custom-indexes/
- https://5efc0af6673a52000884afb5--kubernetes-sigs-krew.netlify.app/docs/developer-guide/custom-indexes/

ahmetb:
Pushed some more changes to https://5efc0af6673a52000884afb5--kubernetes-sigs-krew.netlify.app/docs/developer-guide/custom-indexes/.

ahmetb:
/lgtm cancel

we need to add `draft: true` to frontmatters

chriskim06:
Is that setting external to this PR?

ahmetb:
No, you'll need to update the .md files you added and add the property to their header section.

site/content/docs/user-guide/using-custom-indexes.md
```diff
@@ -0,0 +1,84 @@
...
+You can remove a custom plugin index by passing the name it was added with to
+the remove command:
+
+```sh
+{{<prompt>}}kubectl krew index remove foo
+```
+
+## Listing indexes
+
+To see what indexes you have added you can run the list command
```
corneliusweig, line 38:
```suggestion
To see what indexes you have added run the list command
```

site/content/docs/user-guide/using-custom-indexes.md
```diff
@@ -0,0 +1,84 @@
...
+foo      https://github.com/foo/custom-index.git{{</output>}}
+```
+
+## Installing plugins from custom indexes
+
+Commands for managing plugins (e.g. `install`, `upgrade`) work with custom
+indexes as well.
+
+By default, Krew prefixes plugins with a `default/` prefix. However, if you have
+a plugin installed from a custom index, you need to specify it in format
```
corneliusweig, line 53:
```suggestion
By default, Krew prefixes plugins with a `default/` prefix. So if
a plugin comes from a custom index, you need to specify it in format
```

site/content/docs/user-guide/using-custom-indexes.md
```diff
@@ -0,0 +1,84 @@
...
+- To remove a plugin, you don't need to specify its index:
+
+    ```sh
+    {{<prompt>}}kubectl krew uninstall PLUGIN_NAME
+    ```
+
+- To get information about a plugin from a custom index:
+
+    ```sh
+    {{<prompt>}}kubectl krew info INDEX_NAME/PLUGIN_NAME
```
corneliusweig, line 79: Is that also correct if the plugin is already installed? I don't remember..

site/content/docs/user-guide/using-custom-indexes.md
```diff
@@ -0,0 +1,84 @@
...
+
+Plugin indexes contain plugin manifests, which are documents that describe
+installation procedure for a plugin. For discovery purposes, Krew comes with a
+`default` plugin index,  plugins hosted in the [`krew-index`
+repository](https://github.com/kubernetes-sigs/krew-index).
+
+However, some plugin authors may choose to host their own indexes that contain
+their curation of kubectl plugins. These are called "custom plugin indexes".
+
+## Adding a custom index
```
corneliusweig, line 16: Can you also add an "advanced" section, where you explain how users/vendors could substitute the krew default index with their own?

site/content/docs/developer-guide/custom-indexes.md
```diff
@@ -0,0 +1,54 @@
+---
+title: Hosting Custom Plugin Indexes
+slug: custom-indexes
+weight: 500
```
corneliusweig, line 4:
```suggestion
weight: 900
```
To me this sounds like a much more specialized topic than using custom indexes. So I would put it last.

site/content/docs/developer-guide/custom-indexes.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Duplicate plugin names
+
+Your custom index can contain plugins that have the same name as the ones in
+`krew-index`.
+
+Users of your index will need to install your plugin using the
+explicit `<INDEX>/<PLUGIN>` syntax. See the [user guide][ug].
+
+[ug]: {{< relref "../user-guide/using-custom-indexes.md">}}
```
corneliusweig, line 54: This has considerable overlap with `using-custom-indexes`, and duplicates some information. Do we really need it?

(also the heading does not fit to the second paragraph)

site/content/docs/user-guide/using-custom-indexes.md
chriskim06: Yes since they can have name conflicts, we did it this way so that you can get info for plugin `foo` in default and custom index.

site/content/docs/developer-guide/custom-indexes.md
chriskim06: I think the weight only applies to the folder that it lives in. I added it here in the developer guide above the plugin usage stats page.

site/content/docs/developer-guide/custom-indexes.md
chriskim06: I'm ok with whatever. I think it might be good to have it repeated here as a consideration for people writing plugins for custom indexes and to let them know that people will need to install their plugin differently.

site/content/docs/user-guide/using-custom-indexes.md
chriskim06: Whoops I missed this comment when I was looking at your review yesterday. Sure I can add this

site/content/docs/user-guide/using-custom-indexes.md
```diff
@@ -0,0 +1,91 @@
...
+
+Plugin indexes contain plugin manifests, which are documents that describe
+installation procedure for a plugin. For discovery purposes, Krew comes with a
+`default` plugin index, plugins hosted in the [`krew-index`
+repository](https://github.com/kubernetes-sigs/krew-index).
+
+However, some plugin authors may choose to host their own indexes that contain
+their curation of kubectl plugins. These are called "custom plugin indexes".
+
+## Adding a custom index
```
ahmetb, line 16: I think the title isn’t good. 
It would be better to explain in this section “The default index” and

* how it can be removed
* or pointed to another repo
* and how plugins without an explicit index name mean “default”

chriskim06: I feel like the adding/removing/listing indexes sections are important to showcase the new index command right? I agree though that a section explaining the default index would be good, I can move the advanced usage section I had and change it to reflect the points you listed.

corneliusweig: Yes, I think this should work well. You could also explain how plugin names are resolved. Iow, that a `default/` prefix is added implicitly, if no index name is given. Then you could also explain how plugin name clashes are handled and remove the `Duplicate plugin names` section in the other page.

site/content/docs/user-guide/using-custom-indexes.md
```diff
@@ -0,0 +1,97 @@
...
+To see what indexes you have added run the list command:
+
+```sh
+{{<prompt>}}kubectl krew index list
+{{<output>}}INDEX    URL
+default  https://github.com/kubernetes-sigs/krew-index.git
+foo      https://github.com/foo/custom-index.git{{</output>}}
+```
+
+## The default index
```
ahmetb, line 46: Can we perhaps use code block notation for all “default”s in this paragraph?

site/content/docs/user-guide/using-custom-indexes.md
```diff
@@ -0,0 +1,97 @@
...
+
+When a plugin doesn't have an explicit `INDEX_NAME` prefix it refers to a plugin
```
ahmetb, line 48: , it

site/content/docs/user-guide/using-custom-indexes.md
```diff
@@ -0,0 +1,97 @@
...
+from the default index. These plugins have an implicit `default/` prepended to
+them in Krew commands. The `INDEX_NAME` prefix is used to differentiate plugins
+with the same name across different indexes.
+
+Krew ships with [`krew-index`][ki] as the default index, but this can be removed
+by passing `default` to the remove command. Once this is removed, you can add
```
ahmetb, line 54: > ...Removed using the “kubectl krew index remove default” command

site/content/docs/user-guide/using-custom-indexes.md
ahmetb, line 46: This section should ideally come **after** installing plugins from custom indexes (keep in mind most people might not care about the default index nuances)

PR #624, opened 2020-07-24
Remove multi index flag

### LINKED ISSUE #597: Update help messages to reflect custom index syntax
Here are a list of commands that I think need to have their help messages updated:
- install

Here's a list of maybes since a warning is already printed when someone tries to use custom index syntax (but might be nice to include something in the help message as well):
- uninstall
- upgrade

The examples currently show something along the lines of `kubectl krew [command] PLUGIN`. Once the multi-index feature is released these will need to be updated to include examples of using custom index syntax. Should the help messages be updated to mention something about the concept of a "default" index? Like:
```
Examples:
  To install one or more plugins from the default index, run:
    kubectl krew install NAME [NAME...]

  To install one or more plugins from a custom index named "foo", run:
    kubectl krew install foo/NAME [foo/NAME...]
```

chriskim06:
Changes:
- remove the multi index flag
- update the `install` help message with an example for installing from a custom index
- update integration test function `WithDefaultIndex` to create an index in the migrated layout

Fixes #597 
Related issue: #566 

/area multi-index
/cc @ahmetb 
/cc @corneliusweig

### DIFF
--- a/cmd/krew/cmd/install.go
+++ b/cmd/krew/cmd/install.go
@@ -51,8 +51,11 @@ func init() {
 		Long: `Install one or multiple kubectl plugins.
 
 Examples:
-  To install one or multiple plugins, run:
+  To install one or multiple plugins from the default index, run:
     kubectl krew install NAME [NAME...]
+
+  To install one or multiple plugins from a custom index named "foo", run:
+    kubectl krew install foo/NAME [foo/NAME...]
 
   To install plugins from a newline-delimited file, run:
     kubectl krew install < file.txt
--- a/cmd/krew/cmd/install.go
+++ b/cmd/krew/cmd/install.go
@@ -54,11 +54,11 @@ Examples:
   To install one or multiple plugins from the default index, run:
     kubectl krew install NAME [NAME...]
 
-  To install one or multiple plugins from a custom index, run:
-    kubectl krew install INDEX/NAME [INDEX/NAME...]
-
   To install plugins from a newline-delimited file, run:
     kubectl krew install < file.txt
+
+  To install one or multiple plugins from a custom index, run:
+    kubectl krew install INDEX/NAME [INDEX/NAME...]
 
   (For developers) To provide a custom plugin manifest, use the --manifest or
   --manifest-url arguments. Similarly, instead of downloading files from a URL,
--- a/cmd/krew/cmd/install.go
+++ b/cmd/krew/cmd/install.go
@@ -54,8 +54,8 @@ Examples:
   To install one or multiple plugins from the default index, run:
     kubectl krew install NAME [NAME...]
 
-  To install one or multiple plugins from a custom index named "foo", run:
-    kubectl krew install foo/NAME [foo/NAME...]
+  To install one or multiple plugins from a custom index, run:
+    kubectl krew install INDEX/NAME [INDEX/NAME...]
 
   To install plugins from a newline-delimited file, run:
     kubectl krew install < file.txt
@@ -63,7 +63,7 @@ Examples:
   (For developers) To provide a custom plugin manifest, use the --manifest or
   --manifest-url arguments. Similarly, instead of downloading files from a URL,
   you can specify a local --archive file:
-	kubectl krew install --manifest=FILE [--archive=FILE]
+    kubectl krew install --manifest=FILE [--archive=FILE]
 
 Remarks:
   If a plugin is already installed, it will be skipped.
--- a/cmd/krew/cmd/index.go
+++ b/cmd/krew/cmd/index.go
@@ -35,11 +35,10 @@ var (
 
 // indexCmd represents the index command
 var indexCmd = &cobra.Command{
-	Use:    "index",
-	Short:  "Manage custom plugin indexes",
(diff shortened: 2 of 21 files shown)


cmd/krew/cmd/install.go
```diff
@@ -51,16 +51,19 @@ func init() {
 		Long: `Install one or multiple kubectl plugins.
 
 Examples:
-  To install one or multiple plugins, run:
+  To install one or multiple plugins from the default index, run:
     kubectl krew install NAME [NAME...]
 
+  To install one or multiple plugins from a custom index, run:
+    kubectl krew install INDEX/NAME [INDEX/NAME...]
```
ahmetb, line 58: let's move this after the next example

ahmetb:
/lgtm
/approve

image

ahmetb:
So far this went fine for me on master.

Something I noticed is we print the migration messages as Info (user-visible).
Was that our intention? It doesn't bug me and it's also one-time, so maybe it's ok.

```
krew git:(master) $ tree -d ~/.krew/index
/Users/ahmetb/.krew/index
└── plugins
1 directory
```
```
krew git:(master) $ hack/make-binary.sh
Building binaries for: darwin/amd64
(Stamping with git tag=v0.3.4-78-g7ba6715 rev=7ba6715)

Number of parallel builds: 15
-->    darwin/amd64: sigs.k8s.io/krew/cmd/krew
```
```
krew git:(master) $ out/bin/krew-darwin_amd64 --help
krew is the kubectl plugin manager.
You can invoke krew through kubectl: "kubectl krew [command]..."

Usage:
  kubectl krew [command]

Available Commands:
  help        Help about any command
  index       Manage custom plugin indexes
  info        Show information about an available plugin
  install     Install kubectl plugins
  list        List installed kubectl plugins
  search      Discover kubectl plugins
  uninstall   Uninstall plugins
  update      Update the local copy of the plugin index
  upgrade     Upgrade installed plugins to newer versions
  version     Show krew version and diagnostics

Flags:
  -h, --help      help for krew
  -v, --v Level   number for the log level verbosity

Use "kubectl krew [command] --help" for more information about a command.
```
```
krew git:(master) $ out/bin/krew-darwin_amd64 list
I0724 12:01:24.987470   63847 migration.go:41] Migrating krew index layout.
I0724 12:01:24.987938   63847 migration.go:58] Migration completed successfully.
PLUGIN                 VERSION
access-matrix          v0.4.4
ctx                    v0.9.1
detached/cert-manager  v0.16.0-alpha.0
detached/chonk         v0.1.1
detached/example       v1.0.0
[...]
```
```
krew git:(master) $ tree -d ~/.krew/index
/Users/ahmetb/.krew/index
└── default
    └── plugins

2 directories
````

**Comment by ahmetb** [comment]
Actually I think I found a bug.

After migration I ran a `out/bin/krew-darwin_amd64 upgrade` and got errors so I immediately terminated:


```
Upgrading plugin: access-matrix
Skipping plugin access-matrix, it is already on the newest version
WARNING: failed to upgrade plugin "detached/", skipping (error: open /Users/ahmetb/.krew/index/detached/plugins/cert-manager.yaml: no such file or directory)
WARNING: failed to upgrade plugin "detached/", skipping (error: open /Users/ahmetb/.krew/index/detached/plugins/chonk.yaml: no such file or directory)
Upgrading plugin: ctx
Skipping plugin ctx, it is already on the newest version
WARNING: failed to upgrade plugin "detached/", skipping (error: open /Users/ahmetb/.krew/index/detached/plugins/example.yaml: no such file or directory)
Upgrading plugin: fleet
Upgraded plugin: fleet
WARNING: You installed plugin "fleet" from the krew-index plugin repository.
   These plugins are not audited for security by the Krew maintainers.
   Run them at your own risk.
Upgrading plugin: get-all
Skipping plugin get-all, it is already on the newest version
Upgrading plugin: gs
^C
```

chriskim06:
Ya that one is because when you install via manifest it gets the detached index. Then when upgrade with no args happens [this line](https://github.com/kubernetes-sigs/krew/blob/master/cmd/krew/cmd/upgrade.go#L580) causes it to try to upgrade from a non existent index.

chriskim06:
I thought we agreed on that when we added the detached index

ahmetb:
> I don't think we can assume that plugins installed from manifest are necessarily upgraded from krew-index.

We can.

Where else would they be installed from at the time of the migration from a time where we only allowed 1 index? (They could indeed be installed manually with `--archive|--manifest`?)

We definitely should not suddenly orphan all plugins. Most users have >0 plugins installed. And likely <1% of those are actually "detached". We definitely should not assume all is "detached", that would break ~everyone for no reason.

ahmetb:
Actually, in my case those plugins `chonk`, `cert-manager` and `example` were definitely "detached" at the time of installation.

But the message is confusing:

> WARNING: failed to upgrade plugin "detached/", skipping (error: open /Users/ahmetb/.krew/index/detached/plugins/chonk.yaml: no such file or directory)

- It tells me plugin name is `detached/`
- It tells me there is supposed to be a dir `/Users/ahmetb/.krew/index/detached`

that’ll be confusing (to a very small % of the users who install detached plugins)

PR #626, opened 2020-07-28
Skip upgrades for plugins installed via manifest

### LINKED ISSUE #625: Not friendly "upgrade" message for detached plugins
After doing multi-index migration on my local index, I ran `kubectl krew upgrade`.

```
Upgrading plugin: access-matrix
Skipping plugin access-matrix, it is already on the newest version
WARNING: failed to upgrade plugin "detached/", skipping (error: open /Users/ahmetb/.krew/index/detached/plugins/cert-manager.yaml: no such file or directory)
WARNING: failed to upgrade plugin "detached/", skipping (error: open /Users/ahmetb/.krew/index/detached/plugins/chonk.yaml: no such file or directory)
Upgrading plugin: ctx
Skipping plugin ctx, it is already on the newest version
WARNING: failed to upgrade plugin "detached/", skipping (error: open /Users/ahmetb/.krew/index/detached/plugins/example.yaml: no such file or directory)
Upgrading plugin: fleet
Upgraded plugin: fleet
WARNING: You installed plugin "fleet" from the krew-index plugin repository.
   These plugins are not audited for security by the Krew maintainers.
   Run them at your own risk.
Upgrading plugin: get-all
Skipping plugin get-all, it is already on the newest version
Upgrading plugin: gs
^C
```

Specifically:

> WARNING: failed to upgrade plugin "detached/", skipping (error: open /Users/ahmetb/.krew/index/detached/plugins/cert-manager.yaml: no such file or directory)

In this case, yes this plugin is indeed `detached`, but the error message says:

1. it's looking at a plugin named `detached/`
2. there should be a dir at `$HOME/.krew/index/detached/`

I forgot what the expected behavior should be while running `upgrade` on all plugins that include detached ones (skip or fail?) but regardless, the error message is less than ideal.

Would be good to fix before we tag v0.4. (Though, this only affects plugin developer persona.)

/cc @chriskim06 
/cc @corneliusweig 
/kind bug
/priority p1

chriskim06:
Fixes #625 

/area multi-index
/cc @ahmetb 
/cc @corneliusweig

### DIFF
--- a/cmd/krew/cmd/upgrade.go
+++ b/cmd/krew/cmd/upgrade.go
@@ -78,6 +78,11 @@ kubectl krew upgrade foo bar"`,
 			var nErrors int
 			for _, name := range pluginNames {
 				indexName, pluginName := pathutil.CanonicalPluginName(name)
+				if indexName == "detached" {
+					fmt.Fprintf(os.Stderr, "Skipping upgrade for %q because it was installed via manifest\n", pluginName)
+					continue
+				}
+
 				plugin, err := indexscanner.LoadPluginByName(paths.IndexPluginsPath(indexName), pluginName)
 				if err != nil {
 					if !os.IsNotExist(err) {
--- a/cmd/krew/cmd/upgrade.go
+++ b/cmd/krew/cmd/upgrade.go
@@ -79,7 +79,7 @@ kubectl krew upgrade foo bar"`,
 			for _, name := range pluginNames {
 				indexName, pluginName := pathutil.CanonicalPluginName(name)
 				if indexName == "detached" {
-					fmt.Fprintf(os.Stderr, "Skipping upgrade for %q because it was installed via manifest\n", pluginName)
+					klog.Warningf("Skipping upgrade for %q because it was installed via manifest\n", pluginName)
 					continue
 				}
 
--- a/integration_test/upgrade_test.go
+++ b/integration_test/upgrade_test.go
@@ -81,6 +81,22 @@ func TestKrewUpgradePluginsFromCustomIndex(t *testing.T) {
 	}
 }
 
+func TestKrewUpgradeSkipsManifestPlugin(t *testing.T) {
+	skipShort(t)
+
+	test, cleanup := NewTest(t)
+	defer cleanup()
+
+	test.WithDefaultIndex().
+		Krew("install", "--manifest", filepath.Join("testdata", validPlugin+constants.ManifestExtension)).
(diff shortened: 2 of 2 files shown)


cmd/krew/cmd/upgrade.go
```diff
@@ -78,6 +78,11 @@ kubectl krew upgrade foo bar"`,
 			var nErrors int
 			for _, name := range pluginNames {
 				indexName, pluginName := pathutil.CanonicalPluginName(name)
+				if indexName == "detached" {
+					fmt.Fprintf(os.Stderr, "Skipping upgrade for %q because it was installed via manifest\n", pluginName)
```
ahmetb, line 82: klog.Warnf might be preferable for our output style.

chriskim06: Do you think it should be behind some log level then and use `Infof` (doesn't seem like `Warningf` is available when using levels)? If not the output looks a little strange IMO
```
Updated the local copy of plugin index "local".
Upgrading plugin: local/fields
Skipping plugin local/fields, it is already on the newest version
W0727 17:55:55.211941   25668 upgrade.go:82] Skipping upgrade for "grep" because it was installed via manifest
Upgrading plugin: whoami
Skipping plugin whoami, it is already on the newest version
```

ahmetb: 🤔 yeah we don't have a pretty way of printing warnings.
maybe search the source tree about how other warnings are printed.
we def use klog.warnf in several places and these are _mostly_ visible to plugin developers (exceptions exist).

chriskim06: Ya that's a good point. This also happens in `install` so its probably not that big of a deal, I'll switch it to `Warningf`
```
$ kubectl krew install who-can whoami
Updated the local copy of plugin index "local".
Installing plugin: who-can
Installed plugin: who-can
\
 | Use this plugin:
 |      kubectl who-can
 | Documentation:
 |      https://github.com/aquasecurity/kubectl-who-can
 | Caveats:
 | \
 |  | The plugin requires the rights to list (Cluster)Role and (Cluster)RoleBindings.
 | /
/
WARNING: You installed plugin "who-can" from the krew-index plugin repository.
   These plugins are not audited for security by the Krew maintainers.
   Run them at your own risk.
Installing plugin: whoami
W0727 19:34:59.417519   11450 install.go:160] Skipping plugin "whoami", it is already installed
```

PR #629, opened 2020-08-04
fix path warning on windows

ahmetb:
Previously this was rendering as

    %!U(MISSING)SERPROFILE%!\(MISSING).krew\bin

while installing on windows. Using %% solves that.

### DIFF
--- a/cmd/krew/cmd/internal/setup_check_test.go
+++ b/cmd/krew/cmd/internal/setup_check_test.go
@@ -45,7 +45,7 @@ func TestIsBinDirInPATH_secondRun(t *testing.T) {
 }
 
 func TestSetupInstructions_windows(t *testing.T) {
-	const instructionsContain = "add the\n\"%USERPROFILE%\\.krew\\bin\" directory to your PATH environment variable"
+	const instructionsContain = `USERPROFILE`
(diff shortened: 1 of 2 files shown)


ahmetb:
I also had to update the tests because the function actually has to return `%%USERPROFILE%%` but the test was looking for the full message that doesn't have `%%` notation.

The `%%` gets reduced to `%` at the caller site of this function that uses fmt.Sprintf/Printf.

/assign @corneliusweig

cmd/krew/cmd/internal/setup_check_test.go
```diff
@@ -45,10 +45,11 @@ func TestIsBinDirInPATH_secondRun(t *testing.T) {
 }
 
 func TestSetupInstructions_windows(t *testing.T) {
-	const instructionsContain = "add the\n\"%USERPROFILE%\\.krew\\bin\" directory to your PATH environment variable"
+	const instructionsContain = `USERPROFILE`
 	os.Setenv("KREW_OS", "windows")
 	defer func() { os.Unsetenv("KREW_OS") }()
 	instructions := SetupInstructions()
+	t.Log(instructions)
```
corneliusweig, line 52: nit: intentional?

PR #633, opened 2020-08-15
Use go 1.14 `t.Cleanup()` for simpler tests

corneliusweig:
Go 1.14 comes with a `t.Cleanup()` method to register callbacks to be executed when the test is done (see https://golang.org/doc/go1.14#testing). This PR switches to this pattern.

### DIFF
--- a/internal/testutil/tempdir.go
+++ b/internal/testutil/tempdir.go
@@ -34,20 +34,19 @@ type TempDir struct {
 
 // NewTempDir creates a temporary directory and a cleanup function.
 // It is the responsibility of calling code to call cleanup when done.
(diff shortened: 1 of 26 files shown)


internal/testutil/tempdir.go
```diff
@@ -34,20 +34,19 @@ type TempDir struct {
 
 // NewTempDir creates a temporary directory and a cleanup function.
 // It is the responsibility of calling code to call cleanup when done.
```
chriskim06, line 36: This godoc should be updated.

Very cool though!

corneliusweig: Done. Thanks for spotting!

ahmetb:
Wow nice feature of Go TIL. 
/lgtm
/approve

PR #640, opened 2020-08-28
add Go Vet tool

### LINKED ISSUE #639: Consider add Go Vet tool in the CI
Since we already use _golangci-lint_ static analysis in the _run-lint.sh_ file, adding the yet another static analysis tool _(a.k.a go vet)_ would still makes sense. In order to detect suspicious constructs (too see _registered analyzers_: `go tool vet help`), we should use that official tool. (We are still covering some cases, by using [.golangci.yaml](https://github.com/kubernetes-sigs/krew/blob/master/.golangci.yml) config file.)

Dentrax:
🎉 This PR adds Go Vet tool in the build pipeline.

Fixes #639

### DIFF
--- a/hack/run-tests.sh
+++ b/hack/run-tests.sh
@@ -40,6 +40,9 @@ trap print_status EXIT
 print_with_color "$color_blue" 'Checking boilerplate'
 "$SCRIPTDIR"/verify-boilerplate.sh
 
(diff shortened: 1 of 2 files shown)


ahmetb:
Lets follow the guidance at https://github.com/kubernetes-sigs/krew/issues/639#issuecomment-683203165
/close

PR #642, opened 2020-08-29
validate-krew-manifest: add license check

### LINKED ISSUE #641: Check for presence of license file in CI
Currently, checking for license file in submissions is a manual process. Especially, if manifests use a wildcard to copy out all files, we need to download the tarball, and check if a license file is there.
However, new versions could silently remove the license file and we would not notice.

---

Almost all plugins now have a license file, or claimed to include one in the next version. I think we should therefore automate the license check by including such a check in our install validation automation.

The location and name of licenses has unfortunately some variance:

- sometimes files are nested in `sub/folder/LICENSE`
- some have a suffix `LICENSE.txt`
- some differ in upper/lower casing `license`, `License`

> I suggest to walk all files in the install folder, case-fold everything to upper-case, and match by  prefix `LICENSE`. If no match can be found, fail the manifest validation.

ahmetb:
Ensure platform spec extracts a (case-insensitive) {LICENSE,COPYING}{,.txt}
file by looking at the installation directory.

Fixes #641
/assign @corneliusweig

### DIFF
--- a/cmd/validate-krew-manifest/main.go
+++ b/cmd/validate-krew-manifest/main.go
@@ -165,11 +165,12 @@ func installPlatformSpec(manifestFile string, p index.Platform) error {
 	return errors.Wrap(err, "LICENSE (or alike) file is not extracted from the archive as part of installation")
 }
 
-var licenseFiles = []string{
-	"license",
-	"license.txt",
-	"copying",
-	"copying.txt",
+var licenseFiles = map[string]struct{}{
+	"license":     {},
+	"license.txt": {},
+	"license.md":  {},
+	"copying":     {},
+	"copying.txt": {},
 }
 
 func validateLicenseFileExists(krewRoot string) error {
@@ -190,11 +191,9 @@ func validateLicenseFileExists(krewRoot string) error {
 
 	for _, f := range files {
 		klog.V(8).Infof("found installed file: %s", f)
-		f = strings.ToLower(filepath.Base(f))
-		for _, lf := range licenseFiles {
-			if lf == f {
-				return nil
-			}
+		if _, ok := licenseFiles[strings.ToLower(filepath.Base(f))]; ok {
+			klog.V(8).Infof("found license file %q", f)
+			return nil
 		}
 	}
 	return errors.Errorf("could not find license file among [%s]", strings.Join(files, ", "))
--- a/cmd/validate-krew-manifest/main.go
+++ b/cmd/validate-krew-manifest/main.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/klog"
 
+	"sigs.k8s.io/krew/internal/environment"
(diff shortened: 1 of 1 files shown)


cmd/validate-krew-manifest/main.go
```diff
@@ -159,7 +160,44 @@ func installPlatformSpec(manifestFile string, p index.Platform) error {
...
 	}
-	return nil
+
+	err = validateLicenseFileExists(tmpDir)
+	return errors.Wrap(err, "LICENSE (or alike) file is not extracted from the archive as part of installation")
+}
+
+var licenseFiles = []string{
+	"license",
+	"license.txt",
```
corneliusweig, line 170: Maybe also allow `license.md`?

cmd/validate-krew-manifest/main.go
```diff
@@ -159,7 +160,44 @@ func installPlatformSpec(manifestFile string, p index.Platform) error {
...
+
+	for _, f := range files {
+		klog.V(8).Infof("found installed file: %s", f)
+		f = strings.ToLower(filepath.Base(f))
+		for _, lf := range licenseFiles {
+			if lf == f {
+				return nil
+			}
+		}
+	}
```
corneliusweig, line 199: Maybe use a map instead?

```go
var licenseFiles = map[string]bool{
  "license": true,
  "license.txt": true,
  "copying": true,
  "copying.txt":true,
}
// ...
if licenseFiles[f] {
  return nil
}
```

cmd/validate-krew-manifest/main.go
```diff
@@ -159,7 +160,44 @@ func installPlatformSpec(manifestFile string, p index.Platform) error {
...
+	if err != nil {
+		return errors.Wrap(err, "failed to walk installation directory")
+	}
+
+	for _, f := range files {
+		klog.V(8).Infof("found installed file: %s", f)
+		f = strings.ToLower(filepath.Base(f))
+		for _, lf := range licenseFiles {
+			if lf == f {
+				return nil
```
corneliusweig, line 196: nit:
```go
klog.V(8).Infof("found license file: %s", f)
```

corneliusweig (approved):
Heh, thanks for getting this done right away! I totally didn't expect that :D

cmd/validate-krew-manifest/main.go
ahmetb: done

cmd/validate-krew-manifest/main.go
ahmetb: done

cmd/validate-krew-manifest/main.go
ahmetb: done

corneliusweig:
/lgtm
/approve
Nice work :)

ahmetb:
Do you mind updating the go-get in krew-index CI?

corneliusweig:
Sure, no problem. Will do later today.

PR #656, opened 2020-10-07
Support arm architecture install

rothgar:
Detect architecture with `uname -m` and use amd64 or arm to extract the binaries depending on the current system architecture.


corneliusweig:
/lgtm
/approve
Thanks for this refinement!

PR #658, opened 2020-10-09
Dynamic plugins list in the website

ahmetb:
This relies on Netlify's serverless functions capability to render
plugin count (in the homepage) and plugin list (/docs/plugins) in
the website!

SIG Docs TL has cleared the usage of Functions feature here: 
https://groups.google.com/g/kubernetes-sig-docs/c/cXqQm60tGs8

The serverless function is called "api" and it's accessible on path
/.netlify/functions/api/* and has two subpaths that return JSON responses:

1. */pluginCount
2. */plugins

This serverless function is built by placing a binary in /site/functions
named "api" and the build instruction is in 'netlify.toml' file.

Since we get plugin list by calling GitHub API, by default we run without
an authentication token. However after this is merged, I will go and add a
permissionless $GITHUB_ACCESS_TOKEN that I have created from my account.

Unauthenticated GitHub calls are limited to 60 per IP, but since I'll add a
personal access token, that should not be an issue. When debugging locally,
60 is also fine. (deploy previews for PRs from non-maintainers will not have
an access token available, so they will make unauthenticated calls to GH API.)

The responses from the dynamic functions are actually cached in Netlify's
CDN and that helps with the rate-limiting as well. I have currently set the
CDN cache duration to 1 hour for each response.

Please take a look at the previews and let me know if there is anything that
doesn't look good.

I might delete `generate-plugin-overview` tool and update the current
plugins.md in a followup PR.

/assign @chriskim06
/assign @corneliusweig


ahmetb:
/hold
Not entirely ready for review, requires some setup. I'll ping again later.

ahmetb:
/hold cancel
Ready to preview at https://deploy-preview-658--kubernetes-sigs-krew.netlify.app/.
Feel free to add commits to my branch.

site/functions/server/main.go
```diff
@@ -0,0 +1,287 @@
...
+	var out PluginCountResponse
+	out.Data.Count = count
+
+	w.Header().Set("Cache-Control", fmt.Sprintf("public, max-age=%d", cacheSeconds))
+	writeJSON(w, out)
+}
+
+func writeJSON(w io.Writer, v interface{}) {
+	e := json.NewEncoder(w)
+	e.SetIndent("", "  ")
```
corneliusweig, line 116: Let's not do this and save a few bits on the wire.
If we need to look at this, it will most certainly be in a browser console and they print json objects always the same.
```suggestion
```

site/functions/server/main.go
```diff
@@ -0,0 +1,287 @@
...
+		}
+		out.Data.Plugins = append(out.Data.Plugins, pi)
+	}
+
+	w.Header().Set("Cache-Control", fmt.Sprintf("public, max-age=%d", cacheSeconds))
+	writeJSON(w, out)
+}
+
+func fetchPlugins(entries []*github.RepositoryContent) ([]*krew.Plugin, error) {
+	ctx := context.Background()
```
corneliusweig, line 180: `ctx` should be passed in from `req.Context()`

site/functions/server/main.go
```diff
@@ -0,0 +1,287 @@
...
+		out    []*krew.Plugin
+		retErr error
+	)
+
+	queue := make(chan string)
+	var wg sync.WaitGroup
+
+	for i := 0; i < urlFetchBatchSize; i++ {
+		wg.Add(1)
+		go func(j int) {
```
corneliusweig, line 195:
```suggestion
		go func() {
```
`j` is unused?

site/functions/server/main.go
```diff
@@ -0,0 +1,287 @@
...
+	var (
+		mu     sync.Mutex
+		out    []*krew.Plugin
+		retErr error
+	)
+
+	queue := make(chan string)
+	var wg sync.WaitGroup
+
+	for i := 0; i < urlFetchBatchSize; i++ {
```
corneliusweig, line 193: nit: urlFetchBatchSize -> pluginFetchParallelism
```suggestion
	for i := 0; i < pluginFetchParallelism; i++ {
```

site/functions/server/main.go
```diff
@@ -0,0 +1,287 @@
...
+	defer cancel()
+
+	var (
+		mu     sync.Mutex
+		out    []*krew.Plugin
+		retErr error
+	)
+
+	queue := make(chan string)
+	var wg sync.WaitGroup
```
corneliusweig, line 191: Maybe use errgroup instead and avoid the `context.WithCancel`?

site/functions/server/main.go
```diff
@@ -0,0 +1,287 @@
...
+					if err != nil {
+						retErr = err
+						cancel()
+						return
+					}
+					mu.Lock()
+					out = append(out, p)
+					mu.Unlock()
+				}
+			}
```
corneliusweig, line 215: When using errgroup, this could be simplified to

```go
for url := range queue {
  p, err := readPlugin(url)
  if err != nil {
    return err
  }
  mu.Lock()
  out = append(out, p)
  mu.Unlock()
}
```

site/functions/server/main.go
```diff
@@ -0,0 +1,287 @@
...
+	close(queue)
+	wg.Wait()
+
+	return out, retErr
+}
+
+func readPlugin(url string) (*krew.Plugin, error) {
+	resp, err := http.Get(url)
+	if err != nil {
+		return nil, errors.Wrapf(err,"failed to get %s", url)
```
corneliusweig, line 237: nit: run gofmt :)

site/functions/server/main.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+	if resp.Body != nil {
+		defer resp.Body.Close()
+	}
```
corneliusweig, line 241: nit:
```suggestion
	defer resp.Body.Close()
```

From the spec:
> // The http Client and Transport guarantee that Body is always
	// non-nil, even on responses without a body or responses with
	// a zero-length body. It is the caller's responsibility to
	// close Body.

site/functions/server/main.go
```diff
@@ -0,0 +1,287 @@
...
+	var v krew.Plugin
```
corneliusweig, line 242: nit: move `v` right before `yaml.Unmarshal`

site/functions/server/main.go
```diff
@@ -0,0 +1,287 @@
...
+
+func main() {
+	port := flag.Int("port", -1, "specify a port to use http rather than AWS Lambda")
+	flag.Parse()
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/.netlify/functions/api/pluginCount", pluginCountHandler)
+	mux.HandleFunc("/.netlify/functions/api/plugins", pluginsHandler)
+	// To debug locally, you can run this server with -port=:8080 and run "hugo serve" and uncomment this:
+	mux.Handle("/", httputil.NewSingleHostReverseProxy(&url.URL{Scheme: "http", Host: "localhost:1313"}))
```
corneliusweig, line 280: The comment reads like this should be commented out?

Anyway, I think it'd be much nicer like that, then no commenting is necessary:
```go
if *port > -1 {
   ...
}
```

corneliusweig (approved):
This is really nice work! A few nits, but otherwise this is perfect stuff.

Very cool that you found out about netlify functions. This will save us so much pointless work.

---

I agree that `generate-plugin-overview` should be deleted asap as soon as this is merged.

site/layouts/partials/footer.html
```diff
@@ -44,5 +44,77 @@
...
+    document.addEventListener('DOMContentLoaded', (event) => {
+        var pluginCountElem = document.getElementById("krew-plugin-count");
+        if (!pluginCountElem){
+            return;
+        }
+        fetch('/.netlify/functions/api/pluginCount')
+            .then(response => response.json())
+            .then(response => {
+                console.log(`${response.data.count} plugins found`);
+                pluginCountElem.innerText=response.data?.count || 0;
```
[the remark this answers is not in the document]
ahmetb: Might do `<error>` :D

site/functions/server/main.go
ahmetb: It can be commented out but I left it on just in case. You're right.

ahmetb:
@corneliusweig I think it would be better if you do a pass with your commits on the frontend pieces. My frontend skills are weak at best. :)

site/functions/server/main.go
ahmetb: might use errgroup, but I don't see a way to limit parallelism with errgroup.
feel free to add a patch.

corneliusweig:
@ahmetb I added some commits but had to create a PR to your repo: https://github.com/ahmetb/krew/pull/2
Please have a look

ahmetb:
🤔 I see "Allow edits and access to secrets by maintainers" so you should be able to push to my branch. Have you tried over ssh (git://) ?

site/functions/server/main.go
```diff
@@ -0,0 +1,287 @@
...
+		writeJSON(w, PluginsResponse{Error: ErrorResponse{Message: fmt.Sprintf("failed to fetch plugins: %v", err)}})
+		return
+	}
+
+	for _, v := range plugins {
+		pi := pluginInfo{
+			Name:             v.Name,
+			Homepage:         v.Spec.Homepage,
+			ShortDescription: v.Spec.ShortDescription,
+			GithubRepo:       findRepo(v.Spec.Homepage),
```
chriskim06, line 170: Wouldn't it be fine to just pass `v.Spec.Homepage` here? Then plugins from gitlab or somewhere not in the `knownHomePages` map could still be linked

chriskim06: wait nvm I just realized there's also Homepage being passed here and this is just for the github stars link

corneliusweig:
So I tried again various approaches---nothing will update this PR. However, now there's also a new branch in krew, called `netlify-dynamic-content` :(.

So I think it's easiest if you merge my PR into your PR :roll_eyes:

site/functions/go.mod
```diff
@@ -0,0 +1,17 @@
...
+
+require (
+	github.com/apex/gateway v1.1.1
+	github.com/aws/aws-lambda-go v1.19.1
+	github.com/google/go-github/v32 v32.1.0
+	github.com/pkg/errors v0.9.1
+	golang.org/x/oauth2 v0.0.0-20180821212333-d2e6202438be
+	golang.org/x/sync v0.0.0-20181221193216-37e7f081c4d4
+	gopkg.in/yaml.v2 v2.2.8
+	sigs.k8s.io/krew v0.4.0
```
ahmetb, line 13: this is a bit worrisome to pin the site to a particular version of `master`. (should be ok as long as we keep the types/consts backwards compatible)

this `replace` statement works fine locally, but once I uncommented the `replace` statement and deploy, it fails to build on Netlify sadly. not sure why.

site/functions/server/main.go
```diff
@@ -0,0 +1,272 @@
...
+
+	knownHomePages := map[string]string{
+		`https://krew.sigs.k8s.io/`:                                  "kubernetes-sigs/krew",
+		`https://sigs.k8s.io/krew`:                                   "kubernetes-sigs/krew",
+		`https://kubernetes.github.io/ingress-nginx/kubectl-plugin/`: "kubernetes/ingress-nginx",
+		`https://kudo.dev/`:                                          "kudobuilder/kudo",
+		`https://kubevirt.io`:                                        "kubevirt/kubectl-virt-plugin",
+		`https://popeyecli.io`:                                       "derailed/popeye",
+		`https://soluble-ai.github.io/kubetap/`:                      "soluble-ai/kubetap",
+	}
```
ahmetb, line 251: @corneliusweig I wonder if we should move this to somewhere like a `var` declaration at the top. I suspect it'll get updated frequently. :D

site/layouts/partials/footer.html
```diff
@@ -44,5 +44,58 @@
     </script>
     {{ end }}
 
+    <script>
```
ahmetb, line 47: @corneliusweig should we move these to a .js file at this point?

site/layouts/partials/footer.html
```diff
@@ -44,5 +44,58 @@
...
+            count = json.data?.count ?? '<error>';
+        } catch (e) {
+            console.warn(e);
+        }
+        elems.forEach(elem => elem.innerText=count);
+    });
+    </script>
+
+    <script type="module">
+    import {html, render} from 'https://unpkg.com/lit-html?module';
```
ahmetb, line 66: I must admit I don't fully understand how imports work in the browser so @corneliusweig you’ll probably be on point for maintaining these :D I barely can get the vanilla js working.

site/functions/go.mod
corneliusweig: Yes, it will be annoying if we forget about it. I don't see a better alternative though.

site/functions/server/main.go
corneliusweig: I wonder if we should even go one step further and extract this configuration into non-code. In `krew-index-tracker`, I also need this list and it is error-prone to redeploy every time this changes.

Maybe we can keep this map in a very dumb format in a text file inside `krew-index` or `krew`. Something like
```txt
https://krew.sigs.k8s.io/ kubernetes-sigs/krew
https://sigs.k8s.io/krew kubernetes-sigs/krew
...
```

site/layouts/partials/footer.html
corneliusweig: Yes, that makes sense to me. We don't have to do it in this PR though. Maybe open a new issue? [[main reason: collaboration on the shared PR seems to be not straightforward]]

site/layouts/partials/footer.html
corneliusweig: Yeah, I'm not too worried about it ((support stats)[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import]).

site/functions/server/main.go
```diff
@@ -0,0 +1,272 @@
...
+	return knownHomePages[homePage]
+}
+
+func main() {
+	port := flag.Int("port", -1, `To debug locally set --port=8080 and run "hugo serve"`)
```
corneliusweig, line 256: Running the server locally with `hugo` did not work for me, I always got `not found`, because the site was trying to query the server on port 1313 (hugo) instead of 8080. Can you write down the instructions in a step-by-step manner?

site/layouts/partials/footer.html
```diff
@@ -44,5 +44,58 @@
...
+    document.addEventListener('DOMContentLoaded', async () => {
+        const elems = document.querySelectorAll(".krew-plugin-count");
+        if (elems.length === 0){
+            return;
+        }
+        let count = '<error>';
+        try {
+            const resp = await fetch('/.netlify/functions/api/pluginCount');
+            const json = await resp.json();
+            count = json.data?.count ?? '<error>';
```
chriskim06, line 57: It seems like optional chaining is relatively [new](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#Browser_compatibility) (or at least those version numbers seem a little high). Do either of you think this should be changed to the old way or is it ok?

corneliusweig: The question immediately becomes: how far back do we want to be backwards compatible.

This case is trivial, but `import` and `async/await` is already more involved and the old style quickly becomes kind of ugly. We could warn page visitors by including a `<script nomodule ..>` script to inform the user that the page is not working as intended and to switch to a newer browser.

ahmetb: I think these are mostly fine. I tend to assume developers use latest browser versions most of the time. In that regard they're more savvy than average computer user.

site/functions/server/main.go
ahmetb: pushed a ./site/functions/README.md . PTAL

site/layouts/partials/footer.html
ahmetb: ack

site/functions/server/main.go
ahmetb: I'm -1 for a new file/format.
I think we can hardcode these still in a centralized source code file.
We can do a variable on top of main.go for now.

ahmetb:
added two more commits:

1. 543f132 move from /docs/plugins to /plugins → this way plugin list isn't part of docs
1. fb0b188 add /plugins to navbar → show a "Plugins" item on the top navigation bar.

corneliusweig:
Oh no! I'm so slow. Apologies 🙇‍♂️ 
/lgtm
/approve

PR #659, opened 2020-10-15
Fixed use of fish shell evaluation with (...) instead of bash style $(…)

vicenteherrera:
Fixed use of fish shell evaluation with (...) instead of bash style $(…) in installation script in the specific fish section.

Tested and working well.


vicenteherrera:
Signed the individual Contributor License Agreement.

ahmetb:
Someone tested the old version and it worked well back in the day too. Can you provide an explanation why this works and the other one does not?

corneliusweig:
The installation instruction was modified in #656, adding arm support. Nobody saw that we used bash syntax for this fish snippet.

Thanks for fixing!

/lgtm
/approve

PR #665, opened 2020-11-14
Fix : sanitize PATH dir

### LINKED ISSUE #664: Sanitize path, before comparing with krew binary folder
Hi, 

First of all thanks for this plugin manager it is simple and really convenient. This issue is about warning display when i use krew : 

```
--> kubectl krew list
WARNING: To be able to run kubectl plugins, you need to add
the following to your ~/.bash_profile or ~/.bashrc:

    export PATH="${PATH}:${HOME}/.krew/bin"

and restart your shell.
```

This warning is due to my PATH variable which contain `$PATH:/home/user//go/projects/gopath/.krew/bin` => note double slash for example. And the code here : 
https://github.com/kubernetes-sigs/krew/blob/7ba67155685d978c3661526ed5b9725425c42a97/internal/environment/environment.go#L43-L47
and here : 
https://github.com/kubernetes-sigs/krew/blob/7ba67155685d978c3661526ed5b9725425c42a97/cmd/krew/cmd/internal/setup_check.go#L60-L64

who is not symmetric for bin and path directory. My proposal here https://github.com/RouxAntoine/krew/commit/97099d1d97aa332653b96aef6b63dd6d1041878b is to use function `filepath.Abs(dirInPATH)` to sanitize dir segment of PATH.

Do you think it's a good idea ? I also add unit test `TestIsBinDirInPATH_Custom` for demonstrate this case.

RouxAntoine:
Actually comparison between .krew/bin directory and path segment directory
is done with equal operator. This PR allow method filepath. Abs to be call during krew path creation
`/internal/environment/environment.go#L43` and also in `/cmd/krew/cmd/internal/setup_check.go#L62` path directory iteration

Fixes #664

### DIFF
--- a/cmd/krew/cmd/internal/setup_check.go
+++ b/cmd/krew/cmd/internal/setup_check.go
@@ -58,7 +58,12 @@ func IsBinDirInPATH(paths environment.Paths) bool {
 
 	binPath := paths.BinPath()
 	for _, dirInPATH := range filepath.SplitList(os.Getenv("PATH")) {
-		if dirInPATH == binPath {
+		normalizedDirInPATH, err := filepath.Abs(dirInPATH)
+		if err != nil {
+			klog.Warningf("Cannot get absolute path: %v, %v", normalizedDirInPATH, err)
+			continue
+		}
+		if normalizedDirInPATH == binPath {
 			return true
 		}
 	}
--- a/cmd/krew/cmd/internal/setup_check_test.go
+++ b/cmd/krew/cmd/internal/setup_check_test.go
@@ -23,6 +23,8 @@ import (
 	"sigs.k8s.io/krew/internal/testutil"
 )
 
+const environmentPath = "/home/../home/user//./.krew"
+
 func TestIsBinDirInPATH_firstRun(t *testing.T) {
 	tempDir := testutil.NewTempDir(t)
 
@@ -40,6 +42,35 @@ func TestIsBinDirInPATH_secondRun(t *testing.T) {
 	res := IsBinDirInPATH(paths)
 	if res == true {
 		t.Errorf("expected negative result on second run")
+	}
+}
+
+// TestIsBinDirInPATH_NonNormalized case when PATH content is not well normalized
+func TestIsBinDirInPATH_NonNormalized(t *testing.T) {
+	tempDir := testutil.NewTempDir(t)
+
+	// backup and restore initial PATH value
+	defer func(backupPath string) {
+		err := os.Setenv("PATH", backupPath)
+		if err != nil {
+			t.Fatalf(`os.Setenv("PATH", %s) failed with %v`, backupPath, err)
+		}
+	}(os.Getenv("PATH"))
+
+	// set PATH with non-normalized folder path
+	err := os.Setenv("PATH", tempDir.Path(environmentPath+"/bin"))
+	if err != nil {
+		t.Fatalf(`os.Setenv("PATH", %s/bin) failed with %v`, environmentPath, err)
+	}
+	paths := environment.NewPaths(tempDir.Path(environmentPath))
+	err = os.MkdirAll(paths.BasePath(), os.ModePerm)
+	if err != nil {
+		t.Fatalf("os.MkdirAll(%s) failed with %v", paths.BasePath(), err)
+	}
+
+	got := IsBinDirInPATH(paths)
+	if got == false {
+		t.Errorf("IsBinDirPATH(%v) = %t, want true", paths, got)
 	}
 }
 


cmd/krew/cmd/internal/setup_check.go
```diff
@@ -58,7 +59,11 @@ func IsBinDirInPATH(paths environment.Paths) bool {
 
 	binPath := paths.BinPath()
 	for _, dirInPATH := range filepath.SplitList(os.Getenv("PATH")) {
-		if dirInPATH == binPath {
+		normalizedDirInPATH, err := filepath.Abs(dirInPATH)
+		if err != nil {
+			panic(errors.Wrapf(err, "cannot get absolute path for PATH part %s", dirInPATH))
```
corneliusweig, line 64: Let's not panic here. This is a convenience check and maybe `krew` would just work fine, even though we are not able to resolve the path. So maybe log the result at warning level and `continue`?

cmd/krew/cmd/internal/setup_check.go
```diff
@@ -58,7 +59,11 @@ func IsBinDirInPATH(paths environment.Paths) bool {
 
 	binPath := paths.BinPath()
 	for _, dirInPATH := range filepath.SplitList(os.Getenv("PATH")) {
-		if dirInPATH == binPath {
+		normalizedDirInPATH, err := filepath.Abs(dirInPATH)
```
corneliusweig, line 62:
```suggestion
		dirInPATH, err := filepath.Abs(dirInPATH)
```
It's fine to reuse the old name here.

cmd/krew/cmd/internal/setup_check_test.go
```diff
@@ -43,6 +43,31 @@ func TestIsBinDirInPATH_secondRun(t *testing.T) {
...
 }
 
+// Test case when PATH content is not well normalized
+func TestIsBinDirInPATH_Custom(t *testing.T) {
+	tempDir := testutil.NewTempDir(t)
+	backupPath := os.Getenv("PATH")
+
+	err := os.Setenv("PATH", tempDir.Root()+"/home/../home/user//./.krew/bin")
+	if err != nil {
+		t.Errorf("fail to set PATH with not clean path")
```
corneliusweig, line 53:
```suggestion
		t.Fatalf("fail to set PATH with not clean path")
```
There's no point to continue the test if the setup failed.

cmd/krew/cmd/internal/setup_check_test.go
```diff
@@ -43,6 +43,31 @@ func TestIsBinDirInPATH_secondRun(t *testing.T) {
...
+	}
+	res := IsBinDirInPATH(paths)
+	if res == false {
+		t.Errorf("expected positive result even if PATH directory are not normalized form")
+	}
+
+	err = os.Setenv("PATH", backupPath)
+	if err != nil {
+		t.Errorf("fail to restore PATH env variable after test")
+	}
```
corneliusweig, line 68: Please move the cleanup into a defer'ed function right after (or before) `os.Setenv` in line 51.

cmd/krew/cmd/internal/setup_check_test.go
```diff
@@ -43,6 +43,31 @@ func TestIsBinDirInPATH_secondRun(t *testing.T) {
...
+		t.Errorf("fail to set PATH with not clean path")
+	}
+	paths := environment.NewPaths(tempDir.Path("/home/../home/user//./.krew"))
+	err = os.MkdirAll(paths.BasePath(), os.ModePerm)
+	if err != nil {
+		t.Errorf("fail to create KREW_ROOT path before IsBinDirInPATH invocation to don4t be on first call")
+	}
+	res := IsBinDirInPATH(paths)
+	if res == false {
+		t.Errorf("expected positive result even if PATH directory are not normalized form")
```
corneliusweig, line 62:
```suggestion
		t.Errorf("IsBinDirPATH(%v) = %t, want true", paths, res)
```

cmd/krew/cmd/internal/setup_check_test.go
```diff
@@ -43,6 +43,31 @@ func TestIsBinDirInPATH_secondRun(t *testing.T) {
...
+	err := os.Setenv("PATH", tempDir.Root()+"/home/../home/user//./.krew/bin")
+	if err != nil {
+		t.Errorf("fail to set PATH with not clean path")
+	}
+	paths := environment.NewPaths(tempDir.Path("/home/../home/user//./.krew"))
+	err = os.MkdirAll(paths.BasePath(), os.ModePerm)
+	if err != nil {
+		t.Errorf("fail to create KREW_ROOT path before IsBinDirInPATH invocation to don4t be on first call")
+	}
+	res := IsBinDirInPATH(paths)
```
corneliusweig, line 60:
```suggestion
	got := IsBinDirInPATH(paths)
```

corneliusweig (commented):
Thanks for fixing this! I left some nit comments, otherwise this looks good.

When you push your next revision, please @ me, then I'll take a look more quickly.

cmd/krew/cmd/internal/setup_check.go
RouxAntoine: Now I reuse old name dirInPath but i am not sure that it is a good idea. Indeed if `filepath.Abs(dirInPATH)` function call fail we will have dirInPath set with wrong value for logging message line 64 ?

cf code here :

https://github.com/kubernetes-sigs/krew/pull/665/files#diff-8e3a2729df3d5b4798b7b59ea867eef83cff5f821442bd739451e5f059155cd2R61-R66

RouxAntoine:
@corneliusweig I fix your review, with just one reserve. Thanks for your review

cmd/krew/cmd/internal/setup_check_test.go
```diff
@@ -43,6 +43,40 @@ func TestIsBinDirInPATH_secondRun(t *testing.T) {
...
+// Test case when PATH content is not well normalized
+func TestIsBinDirInPATH_NonNormalized(t *testing.T) {
+	tempDir := testutil.NewTempDir(t)
+
+	// backup and restore initial PATH value
+	backupPath := os.Getenv("PATH")
+	defer func(backupedPath string) {
+		err := os.Setenv("PATH", backupedPath)
+		if err != nil {
+			t.Errorf("fail to restore PATH env variable after test")
```
corneliusweig, line 55:
```suggestion
			t.Fatalf("os.Setenv(todo) failed with %v", todo, err)
```
It is better to use Fatalf here, because we want to make sure that no other test runs in this process. Otherwise, this test could fail erroneously due to the polluted PATH variable.

cmd/krew/cmd/internal/setup_check_test.go
```diff
@@ -43,6 +43,40 @@ func TestIsBinDirInPATH_secondRun(t *testing.T) {
 	}
 }
 
+// Test case when PATH content is not well normalized
+func TestIsBinDirInPATH_NonNormalized(t *testing.T) {
+	tempDir := testutil.NewTempDir(t)
+
+	// backup and restore initial PATH value
+	backupPath := os.Getenv("PATH")
```
corneliusweig, line 51: Since you are not capturing the variable, but passing as argument, consider to inline in line 57.

cmd/krew/cmd/internal/setup_check_test.go
```diff
@@ -43,6 +43,40 @@ func TestIsBinDirInPATH_secondRun(t *testing.T) {
...
+		}
+	}(backupPath)
+
+	// given
+	// set PATH with non-normalized folder path
+	err := os.Setenv("PATH", tempDir.Root()+"/home/../home/user//./.krew/bin")
+	if err != nil {
+		t.Fatalf("fail to set PATH with not clean path")
+	}
+	paths := environment.NewPaths(tempDir.Path("/home/../home/user//./.krew"))
```
corneliusweig, line 65: Please extract the path into a const variable in the test function, so that you don't have to repeat it.

cmd/krew/cmd/internal/setup_check_test.go
```diff
@@ -43,6 +43,40 @@ func TestIsBinDirInPATH_secondRun(t *testing.T) {
 	}
 }
 
+// Test case when PATH content is not well normalized
```
corneliusweig, line 46: This comment should conform to godoc convention and start with the test function name. Or move this comment into the function.

corneliusweig (commented):
Just some last nits and then let's merge this!

RouxAntoine:
> Just some last nits and then let's merge this!

Second fix batch in commit e102013. Do you see my response here https://github.com/kubernetes-sigs/krew/pull/665#discussion_r523793423 ?

cmd/krew/cmd/internal/setup_check_test.go
```diff
@@ -43,6 +45,35 @@ func TestIsBinDirInPATH_secondRun(t *testing.T) {
...
 
+// TestIsBinDirInPATH_NonNormalized case when PATH content is not well normalized
+func TestIsBinDirInPATH_NonNormalized(t *testing.T) {
+	tempDir := testutil.NewTempDir(t)
+
+	// backup and restore initial PATH value
+	defer func(backupPath string) {
+		err := os.Setenv("PATH", backupPath)
+		if err != nil {
+			t.Fatalf("os.Setenv(\"PATH\", %s) failed with %v", backupPath, err)
```
corneliusweig, line 56: No need to change this. But if you want to not mess around with escaping, you can use backticks in Go. This is often more readable:
```suggestion
			t.Fatalf(`os.Setenv("PATH", %s) failed with %v`, backupPath, err)
```

cmd/krew/cmd/internal/setup_check.go
corneliusweig: Oops. I thought I had already responded that you are right here, but apparently that reply got lost 😅 

I don't think it is critical, but it would be cleaner to change that.

corneliusweig:
/approve
I'll wait for you if you want to change the shadowing glitch that I lured you into. Please @ me.

RouxAntoine:
> /approve
> I'll wait for you if you want to change the shadowing glitch that I lured you into. Please @ me.

thanks i fixed both now all seem good for me :)

RouxAntoine:
As i am not sure if edited message with @ ping really ping the personne So i allow myself to ping again @corneliusweig (sorry if duplicate)

corneliusweig:
> As i am not sure if edited message with @ ping really ping the personne So i allow myself to ping again @corneliusweig (sorry if duplicate)

I think it doesn't, I had the same problem before :)

Thanks again for looking into this!
/lgtm
/approve

Issue #637, opened 2020-08-22
Allow overriding `DefaultIndexURI` from an env variable


corneliusweig:
We want to give operators who do not want to install from the default krew index the possibility to provide their own release branch of `krew`. For that, we should allow overriding the `DefaultIndexURI` from an env variable, e.g. `KREW_DEFAULT_INDEX_URI`.
Then operators can start using their own index (containing some krew plugin) from scratch. This will then also set up the default index and point to this custom index location.

- [ ] Implement override via env variable.
- [ ] Update the documentation for custom indices.

/cc @chriskim06

ahmetb:
I'm supportive. This likely only benefits people who never want to touch krew’s own default.

chriskim06:
Is this something we need before 0.4.0?

ahmetb:
No, I think it can be done later.

chriskim06:
Been busy this week, but I'll try to take a crack at this this weekend. I have a feeling integration testing might be interesting but I haven't taken a close look yet.

chriskim06:
This completely fell off my radar, I'll try to get to this soon

/assign

PR #666, opened 2020-11-14
Override default index

chriskim06:
Let me know what you guys think about the integration tests. I decided to reuse the krew-index tarball when `KREW_DEFAULT_INDEX_URI` is passed in a test by unarchiving it in the directory specified. Even though it's the same contents as krew-index I felt it still tests this feature since the default index is changing to some temp dir instead of https://github.com/kubernetes-sigs/krew-index. 

I figured I could do the docs change in a follow up PR, but let me know if you want me to add it here.

Related issue: #637

### DIFF
--- a/integration_test/install_test.go
+++ b/integration_test/install_test.go
@@ -141,6 +141,18 @@ func TestKrewInstall_CustomIndex(t *testing.T) {
 	test.AssertExecutableNotInPATH("kubectl-" + validPlugin2)
 }
 
+func TestKrewInstall_CustomDefaultIndexURI(t *testing.T) {
+	skipShort(t)
+	test := NewTest(t)
+
+	index := filepath.Join(test.Root(), "custom")
+	test.WithEnv("KREW_DEFAULT_INDEX_URI", index).WithDefaultIndex()
+
+	test.Krew("install", validPlugin).RunOrFail()
+	test.AssertExecutableInPATH("kubectl-" + validPlugin)
+	test.AssertPluginFromIndex(validPlugin, "default")
+}
+
 func TestKrewInstallNoSecurityWarningForCustomIndex(t *testing.T) {
 	skipShort(t)
 
--- a/integration_test/install_test.go
+++ b/integration_test/install_test.go
@@ -141,18 +141,6 @@ func TestKrewInstall_CustomIndex(t *testing.T) {
 	test.AssertExecutableNotInPATH("kubectl-" + validPlugin2)
 }
 
-func TestKrewInstall_CustomDefaultIndexURI(t *testing.T) {
-	skipShort(t)
-	test := NewTest(t)
-
-	index := filepath.Join(test.Root(), "custom")
-	test.WithEnv("KREW_DEFAULT_INDEX_URI", index).WithDefaultIndex()
-
-	test.Krew("install", validPlugin).RunOrFail()
-	test.AssertExecutableInPATH("kubectl-" + validPlugin)
-	test.AssertPluginFromIndex(validPlugin, "default")
-}
-
 func TestKrewInstallNoSecurityWarningForCustomIndex(t *testing.T) {
 	skipShort(t)
 
--- a/integration_test/testutil_test.go
+++ b/integration_test/testutil_test.go
@@ -269,9 +269,7 @@ func (it *ITest) TempDir() *testutil.TempDir {
 }
 
 // InitializeIndex initializes the krew index in `$root/index` with the actual krew-index.
-// It caches the index tree as in-memory tar after the first run. If KREW_DEFAULT_INDEX_URI
-// is set then the krew-index tar will be extracted at the target directory to simulate a
-// custom default index.
+// It caches the index tree as in-memory tar after the first run.
 func (it *ITest) initializeIndex() {
 	initIndexOnce.Do(func() {
 		persistentCacheFile := filepath.Join(os.TempDir(), persistentIndexCache)
--- a/integration_test/testutil_test.go
+++ b/integration_test/testutil_test.go
@@ -293,29 +293,16 @@ func (it *ITest) initializeIndex() {
 		}
 	})
 
-	// check for KREW_DEFAULT_INDEX_URI
-	for _, e := range it.env {
-		if strings.Contains(e, "KREW_DEFAULT_INDEX_URI") {
-			val := strings.Split(e, "=")[1]
-			it.extractKrewIndexTar(val)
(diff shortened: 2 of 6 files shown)


pkg/constants/constants.go
```diff
@@ -14,14 +14,24 @@
...
 )
+
+// DefaultIndexURI points to the upstream index.
+var DefaultIndexURI = getDefaultIndexURI()
+
+func getDefaultIndexURI() string {
+	if uri := os.Getenv("KREW_DEFAULT_INDEX_URI"); uri != "" {
+		return uri
+	}
+	return "https://github.com/kubernetes-sigs/krew-index.git"
```
corneliusweig, line 36: Maybe keep this as an unexported constant?

pkg/constants/constants.go
```diff
@@ -14,14 +14,24 @@
...
+}
```
corneliusweig, line 37: I'd suggest to use this pattern:

```suggestion
const defaultIndexURI = "https://github.com/kubernetes-sigs/krew-index.git"
var DefaultIndexURI string = defaultIndexURI

func init() {
	if uri := os.Getenv("KREW_DEFAULT_INDEX_URI"); uri != "" {
		DefaultIndexURI = uri
	}
}
```

Then you can unit-test this logic by explicitly calling `init()` in your test. However, I think this function is so simple that it does necessarily not need one.

integration_test/version_test.go
```diff
@@ -36,7 +48,7 @@ func TestKrewVersion(t *testing.T) {
 		}
 		optionValue := lineSplit.Split(line, 2)
 		if len(optionValue) < 2 {
-			t.Errorf("Expect `krew version` to output `OPTION VALUE` pair separated by spaces, got: %v", optionValue)
+			test.t.Errorf("Expect `krew version` to output `OPTION VALUE` pair separated by spaces, got: %v", optionValue)
```
corneliusweig, line 51: Better return an error from this function and log the result in the caller, i.e.

```suggestion
return fmt.Errorf("Expect `krew version` to output `OPTION VALUE` pair separated by spaces, got: %v", optionValue)
```

integration_test/version_test.go
```diff
@@ -27,7 +27,19 @@ func TestKrewVersion(t *testing.T) {
 	test := NewTest(t)
 
 	stdOut := string(test.Krew("version").RunOrFailOutput())
+	checkRequiredSubstrings(test, "https://github.com/kubernetes-sigs/krew-index.git", stdOut)
+}
+
+func TestKrewVersion_CustomDefaultIndexURI(t *testing.T) {
```
corneliusweig, line 33: I think this test is the best to keep, because it is simple and does not do any extra installation. :+1:

integration_test/install_test.go
```diff
@@ -141,6 +141,18 @@ func TestKrewInstall_CustomIndex(t *testing.T) {
 	test.AssertExecutableNotInPATH("kubectl-" + validPlugin2)
 }
 
+func TestKrewInstall_CustomDefaultIndexURI(t *testing.T) {
```
corneliusweig, line 144: I don't think this test is necessary. We don't need to check all our machinery with & without setting all env variables. You can't do that exhaustively anyway, because you get a factorial explosion of test cases.

I'd recommend to have a single simple test that checks if the env var gets picked up.

integration_test/testutil_test.go
```diff
@@ -291,16 +293,29 @@ func (it *ITest) initializeIndex() {
...
 
+	// check for KREW_DEFAULT_INDEX_URI
+	for _, e := range it.env {
+		if strings.Contains(e, "KREW_DEFAULT_INDEX_URI") {
+			val := strings.Split(e, "=")[1]
+			it.extractKrewIndexTar(val)
+			it.Krew("index", "add", "default", val).RunOrFail()
+			return
+		}
+	}
```
corneliusweig, line 304: Let's not complicate the test setup further. This adds considerable complexity for a one-line config change.

I'd prefer not to have a special case for `KREW_DEFAULT_INDEX_URI` here.

corneliusweig (commented):
Thanks for looking into this, @chriskim06!

Given that this is a really simple logic change, I don't think we should go crazy with testing all config variations. So I think it would be better to keep just one of the integration tests, and maybe add unit tests? However, the logic is simple enough so that it does not require a unit test IMO.

corneliusweig:
Btw, look at the issue number :smiling_imp:

chriskim06:
Thanks for the feedback! You're right that the integration tests were definitely the most complex piece of this. I'll make the changes you suggested.

integration_test/version_test.go
```diff
@@ -36,7 +56,7 @@ func TestKrewVersion(t *testing.T) {
 		}
 		optionValue := lineSplit.Split(line, 2)
 		if len(optionValue) < 2 {
-			t.Errorf("Expect `krew version` to output `OPTION VALUE` pair separated by spaces, got: %v", optionValue)
+			return errors.Errorf("Expected `krew version` to output `OPTION VALUE` pair separated by spaces, got: %v", optionValue)
```
corneliusweig, line 59:
```suggestion
			return errors.Errorf("`%v` is not an `OPTION VALUE` pair separated by spaces", optionValue)
```
It is much better to start errors with lower case

corneliusweig (approved):
Just one last nit, then let's merge this!

chriskim06:
@corneliusweig @ahmetb could I get another round of reviews 🙏

corneliusweig:
@chriskim06 Sure, looks good! Thanks for taking care of this!
/lgtm
/approve

pkg/constants/constants.go
```diff
@@ -14,14 +14,26 @@
...
-	DefaultIndexURI = "https://github.com/kubernetes-sigs/krew-index.git"
 	// DefaultIndexName is a magic string that's used for a plugin name specified without an index.
 	DefaultIndexName = "default"
+
+	// defaultIndexURI is the default krew-index URI that gets used unless it's overridden
+	defaultIndexURI = "https://github.com/kubernetes-sigs/krew-index.git"
 )
+
+// DefaultIndexURI points to the upstream index.
+var DefaultIndexURI = defaultIndexURI
```
ahmetb, line 33: sorry for the after-the-merge review.

I think we should not have code in pkg/constants that are actually not constants and change in runtime.

I recommend keeping this package as-is and implementing a DefaultIndex() method in another package, like pkg/index.

chriskim06: no worries, that makes sense. I think I can get a PR for that in later today

PR #668, opened 2020-11-24
Default index function

chriskim06:
Adding this based on the comment here https://github.com/kubernetes-sigs/krew/pull/666#discussion_r528472892. This reverts back `constants.go` so that it only has the constant values and adds a `pkg/index/util.go` that has the `DefaultIndex` function in it.

/assign @ahmetb 
/assign @corneliusweig 

Related issue: #637

### DIFF
--- a/pkg/index/util.go
+++ b/pkg/index/util.go
@@ -1,4 +1,4 @@
-// Copyright 2019 The Kubernetes Authors.
+// Copyright 2020 The Kubernetes Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/pkg/index/util.go
+++ b/pkg/index/util.go
@@ -12,16 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package constants
+package index
 
-const (
-	CurrentAPIVersion = "krew.googlecontainertools.github.com/v1alpha2"
-	PluginKind        = "Plugin"
-	ManifestExtension = ".yaml"
-	KrewPluginName    = "krew" // plugin name of krew itself
+import (
+	"os"
 
-	// DefaultIndexURI points to the upstream index.
-	DefaultIndexURI = "https://github.com/kubernetes-sigs/krew-index.git"
-	// DefaultIndexName is a magic string that's used for a plugin name specified without an index.
-	DefaultIndexName = "default"
+	"sigs.k8s.io/krew/pkg/constants"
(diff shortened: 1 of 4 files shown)


pkg/index/util.go
```diff
@@ -0,0 +1,28 @@
+// Copyright 2019 The Kubernetes Authors.
```
corneliusweig, line 1:
```suggestion
// Copyright 2020 The Kubernetes Authors.
```

corneliusweig (commented):
This is the right call to make `constants` really constant.

Looks all good, but you might want to adjust the date :)
/lgtm
/approve
/hold

pkg/index/util.go
```diff
@@ -0,0 +1,28 @@
...
+
+package index
+
+import (
+	"os"
+
+	"sigs.k8s.io/krew/pkg/constants"
+)
+
+func DefaultIndex() string {
```
ahmetb, line 23: a mini unit-test would be great for this.

ahmetb:
/approve

I'm now thinking if this option is worth documenting somewhere. We have other options like $KREW_ROOT that also are not documented, maybe in the future.

pkg/index/util.go
chriskim06: for sure, I'll add that in

chriskim06:
I was planning on adding the docs in a separate PR but if you guys think we don't need it for now I can get to it later on

pkg/index/util_test.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"os"
+	"testing"
+
+	"sigs.k8s.io/krew/pkg/constants"
+)
+
+func TestDefaultIndex(t *testing.T) {
+	if got := DefaultIndex(); got != constants.DefaultIndexURI {
+		t.Errorf("expected DefaultIndex() to be %q; got=%q", constants.DefaultIndexURI, got)
```
corneliusweig, line 26:
```suggestion
		t.Errorf("DefaultIndex() = %q, want %q", got, constants.DefaultIndexURI)
```
Same below.

corneliusweig (commented):
If you don't forget about it, you can add the docs later. But we are in no hurry with this and it is far less useful while not documented.

corneliusweig:
/lgtm
Release the hold of you want to do the documentation in a separate PR.

chriskim06:
Ya I'd like to do them in a separate PR, I'll be able to submit that PR this weekend.

/hold cancel

PR #669, opened 2020-11-28
Add documentation for default index URI variable

### LINKED ISSUE #637: Allow overriding `DefaultIndexURI` from an env variable

chriskim06:
So I wasn't super sure where this should go. After looking around I was thinking about either adding it as a separate page under the `setup/` tree or consolidating this with the update check doc. I landed on consolidating the two but let me know what you guys think.

Fixes #637 

/assign @ahmetb 
/assign @corneliusweig


ahmetb:
I think this should go under User Guide directly, right after "Using Custom Plugin Indexes" which is currently the last page in there.

We can call this "Advanced Configuration" to indicate most users should not be bothered by this page while reading the user guide.

site/content/docs/user-guide/setup/configuration.md
```diff
@@ -0,0 +1,38 @@
...
+environment variable. To permanently disable this, add the following to your
+`~/.bashrc`, `~/.bash_profile`, or `~/.zshrc`:
+
+```shell
+export KREW_NO_UPGRADE_CHECK=1
+```
+
+### Use a different default index
+
+By default Krew uses [krew-index][ki] as the default plugin index. This is the
```
ahmetb, line 25: I'd rewrite this paragraph as follows:


> When Krew is installed, it automatically initializes an index named `default` pointing to [krew-index][ki] repository. If you choose to remove this index, see [custom indexes] documentation. You can also force Krew to point it to another repository by setting `KREW_DEFAULT_INDEX_URI`.

+mash that up with existing links + next paragraph you have.

chriskim06: What about:
 
> When Krew is installed, it automatically initializes an index named `default`
pointing to the [krew-index][ki] repository. You can force Krew to use a
different repository by setting `KREW_DEFAULT_INDEX_URI` before running the
[installation instructions]({{<ref "install.md">}}) or after [removing the
default index]({{<ref "using-custom-indexes.md">}}).

> To use a different default index, set the `KREW_DEFAULT_INDEX_URI` environment
variable in your `~/.bashrc`, `~/.bash_profile`, or `~/.zshrc`:

I think it would make sense to keep the parts on when the env var needs to be set together instead of mentioning that you can set it after deleting the default index in the first paragraph and before krew installation in the second paragraph.

Also should I make this `draft: true`? IIRC the docs get published automatically.

ahmetb: Don't worry about setting it as draft without a release being available. I highly suspect anyone’ll come looking for it.

I think your edit is good, make sure to mention this URI needs to point to a git repository reference (can be a real url, or another git-remote protocol like foo://bar/quux.git).

chriskim06: Made the change, @ahmetb mind taking another look?

corneliusweig (commented):
/lgtm
Thanks for giving our docs some :heart:

chriskim06:
@corneliusweig I think your review comments were on an older version of this PR

corneliusweig:
Yes indeed. Epic fail with GitHub app for mobile :-S
Looks like I also need to
/approve

PR #671, opened 2020-12-02
Fix rights of the installDir

JrCs:
Before this fix the installDir have the default mode of the ioutil.Tempdir function which is 0700.
If krew is install globally for multi users, they can't access the krew plugins.
This fix force the 0755 mode on the installDir.

### DIFF
--- a/internal/installation/move.go
+++ b/internal/installation/move.go
@@ -178,6 +178,10 @@ func moveToInstallDir(srcDir, installDir string, fos []index.FileOperation) erro
 		}()
 		return errors.Wrapf(err, "could not rename/copy directory %q to %q", tmp, installDir)
 	}
+
+	if err = os.Chmod(installDir, 0755); err != nil {
+		return errors.Wrapf(err, "could not change rights on directory %q", installDir)
+	}
 	return nil
 }
 


ahmetb:
That might not be the best place to fix it.
It's happening because we create a temp extraction directory and pass it to `downloadAndExtract`.
To create that dir, we use ioutil.TempDir as you said. 

So I think what we should do is refactor `downloadAndExtract` to chmod the specified `extractDir` (with a `// note` indicating why we do it). If we don't fix it at the right place, it might get lost during some refactoring in the future.

corneliusweig:
Yes, this should be the place: https://github.com/kubernetes-sigs/krew/blob/01b33505c4105bdc7f58e4c67acc9c61af2224a7/internal/download/downloader.go#L194

papuneftali (approved):

papuneftali (approved):

corneliusweig:
@JrCs have you seen the comments above?

JrCs:
@corneliusweig yes i seen the comment but i don't have time to make a new PR.
I was thinking that it's better to change the rights in moveToInstallDir when the directory is move to the final place instead of change the rights in the tmp dir (where it can create some security issues).

corneliusweig:
I see. Well the place may not be ideal, but it's not bad either. For example, in nearby code the `installDir` is already created with mode `0755`. So I would be okay to merge as-is.
@ahmetb do you have a strong opinion here?

ahmetb:
I definitely prefer where `os.Chmod` resides right next to where we create `installDir`. We do it because TempDir sets 0700, so if it sits next to it (+with a nice `// Note:`) it would be most preferable.

Otherwise we'd end up bunch of chmods throughout the codebase that are hard to trace back what their context was.

PR #677, opened 2020-12-17
Changes install doc to remove question of path authority

### LINKED ISSUE #676: Heavy handed path setting?
The install docs say to set the path to:
"${KREW_ROOT:-$HOME/.krew}/bin:$PATH"

Why put Krew at the front? This seems a bit heavy handed? Could it simply be put at the end? Is it overriding something in my path?

flickerfly:
As the linux path prefers running what shows up first in the path, there is a potential for someone to create a script in .krew/bin to override another non-krew binary or script from anywhere in the system, even /bin. This puts krew at the end of the path to remove that potential and improve overall security.

Fixes #676

### DIFF
--- a/site/content/docs/user-guide/setup/install.md
+++ b/site/content/docs/user-guide/setup/install.md
@@ -33,7 +33,7 @@ Krew self-hosts).
    this, update your `.bashrc` or `.zshrc` file and append the following line:
 
      ```sh
-     export PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH"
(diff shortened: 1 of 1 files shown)


corneliusweig:
@flickerfly you referenced a closed issue. Are you still proposing to merge this? If so, you need to look into the CLA issue.

flickerfly:
No, I made my case for this, but I think @ahmetb was not interested because it would remove overriding other installed, but potentially older, versions of krew.

corneliusweig:
Ok, thanks for clarifying. I'll go ahead and close this then.
/close

PR #692, opened 2021-02-16
Support LICENSES filename in manifest validation

### LINKED ISSUE #691: Validation for cert-manager plugin fails because of license check
The validation of the [cert-manager](https://github.com/jetstack/cert-manager) plugin currently fails because of the license check that was introduced in https://github.com/kubernetes-sigs/krew/pull/642.

The cert-manager plugin artefacts contain a file named `LICENSES` which is not covered by the valid license file names specified in https://github.com/kubernetes-sigs/krew/blob/master/cmd/validate-krew-manifest/main.go#L168-L174

Example artefact: https://github.com/jetstack/cert-manager/releases/download/v1.2.0/kubectl-cert_manager-darwin-amd64.tar.gz
Failing validation run: https://github.com/kubernetes-sigs/krew-index/pull/1037/checks?check_run_id=1910464291

bashofmann:
Some plugins like https://github.com/jetstack/cert-manager store the license information in a file called LICENSES. This adds support for this filename to the manifest validation.

Fixes #691

### DIFF
--- a/cmd/validate-krew-manifest/main.go
+++ b/cmd/validate-krew-manifest/main.go
@@ -166,11 +166,14 @@ func installPlatformSpec(manifestFile string, p index.Platform) error {
 }
 
 var licenseFiles = map[string]struct{}{
-	"license":     {},
-	"license.txt": {},
-	"license.md":  {},
-	"copying":     {},
-	"copying.txt": {},
+	"license":      {},
+	"license.txt":  {},
+	"license.md":   {},
+	"licenses":     {},
+	"licenses.txt": {},
+	"licenses.md":  {},
+	"copying":      {},
+	"copying.txt":  {},
 }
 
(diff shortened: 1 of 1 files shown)


corneliusweig:
Makes sense. Thank you!
/lgtm
/approve

ahmetb:
do we need a new release ASAP or not?

bashofmann:
The CI pipeline in krew-index currently uses the validate check from a specific commit hash (https://github.com/kubernetes-sigs/krew-index/blob/master/.github/workflows/ci.yml#L49), which points to this commit https://github.com/kubernetes-sigs/krew/commit/9fc1c540a0c2670ce794995084d10875a13e528c.
I guess we could just change it there to the new commit without having to da a krew release. What do you think?

ahmetb:
Yeah I guess only the validation is broken, there's no user impact. That would work.

PR #693, opened 2021-02-17
add support for darwin/arm64 (Apple Silicon)

### LINKED ISSUE #686: Failed to install on Apple Silicon
I tried to install `krew` using the script from the installation documentation on an Apple Silicon.

```
(
  set -x; cd "$(mktemp -d)" &&
  curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/latest/download/krew.tar.gz" &&
  tar zxvf krew.tar.gz &&
  KREW=./krew-"$(uname | tr '[:upper:]' '[:lower:]')_$(uname -m | sed -e 's/x86_64/amd64/' -e 's/arm.*$/arm/')" &&
  "$KREW" install krew
)
```

Output

```
+-zsh:21> mktemp -d
+-zsh:21> cd /var/folders/k6/swv_sxmx2810__qbn27864sh0000gn/T/tmp.TqCOwwsk
+-zsh:22> curl -fsSLO https://github.com/kubernetes-sigs/krew/releases/latest/download/krew.tar.gz
+-zsh:23> tar zxvf krew.tar.gz
x ./LICENSE
x ./krew-darwin_amd64
x ./krew-linux_amd64
x ./krew-linux_arm
x ./krew-windows_amd64.exe
+-zsh:24> KREW=+-zsh:24> uname
+-zsh:24> KREW=+-zsh:24> tr '[:upper:]' '[:lower:]'
+-zsh:24> KREW=+-zsh:24> uname -m
+-zsh:24> KREW=+-zsh:24> sed -e s/x86_64/amd64/ -e 's/arm.*$/arm/'
+-zsh:24> KREW=./krew-darwin_arm
+-zsh:25> ./krew-darwin_arm install krew
zsh: no such file or directory: ./krew-darwin_arm
```

Can we support Apple Silicon, or is there any workaround to install it on Apple Silicon?

joelanford:
This PR adds support for Apple Silicon by adding `darwin/arm64` to the build matrix.

The install scripts are a bit more complicated now because `uname -m` on Apple Silicon returns `arm64`, so the `sed -e 's/arm.*$/arm/'` substitution resulted in `darwin_arm`, rather than `darwin_arm64`.

I fixed that as best I could with a case statement, but perhaps there's a simpler approach. I looked into a pure sed solution, but it doesn't seem like sed supports the necessary lookahead features that would enable search/replace for `arm` + anything other than `64`.

Fixes #686


site/content/docs/user-guide/setup/install.md
```diff
@@ -22,9 +22,11 @@ Krew self-hosts).
     ```sh
     (
       set -x; cd "$(mktemp -d)" &&
+      OS="$(uname | tr '[:upper:]' '[:lower:]')" &&
+      ARCH=`case "${OS}" in "darwin") uname -m | sed -e 's/x86_64/amd64/' ;; *) uname -m | sed -e 's/x86_64/amd64/' -e 's/arm.*$/arm/' -e 's/aarch64$/arm64/' ;; esac` &&
```
ahmetb, line 26: this part is a little suspicious: why did you need `case "${OS}" in "darwin")` ?

it seems that `uname -m` already returns arm64 in Apple Silicon chip, so it would work as is?

if you did this because of `-e 's/arm.*$/arm/'`, that's understandable as it could return values like `armv7l`, `armv8` in other platforms such as Linux. in that case, we can rewrite the regexp pattern in a way that does not replace `amd64` (coming from M1).

joelanford: Yeah, this is the part that I mentioned in the PR description.

> The install scripts are a bit more complicated now because uname -m on Apple Silicon returns arm64, so the sed -e 's/arm.*$/arm/' substitution resulted in darwin_arm, rather than darwin_arm64.
>
> I fixed that as best I could with a case statement, but perhaps there's a simpler approach. I looked into a pure sed solution, but it doesn't seem like sed supports the necessary lookahead features that would enable search/replace for arm + anything other than 64.

Any better ideas?

joelanford: Pushed a new update with a better sed-only replacement that seems to work. WDYT?

ahmetb:
Looks good. Thanks.
/lgtm
/approve

PR #695, opened 2021-02-27
Add creation timestamp to receipts on install

### LINKED ISSUE #632: creationTimestamp is null in all receipts
All my receipts have such a header:

```yaml
metadata:
   creationTimestamp: null
   name: krew
```

including `creationTimestamp`. This field carries no information and is annoying. There are two options:

* Get rid of this obsolete field.
* Fill the field when installing the plugin.

Following https://kubernetes.io/docs/reference/using-api/api-concepts/#generated-values, it seems most inline with k8s concepts, if the stored timestamp is the time when the plugin was first installed.

/help

chriskim06:
This PR adds the current time as the timestamp when a plugin is installed and leaves it alone when something is upgraded. I wasn't sure if we should do something about existing receipts with null `creationTimestamp`'s but figured it would be ok without something.

/assign @ahmetb 
/assign @corneliusweig 

Fixes #632

### DIFF
--- a/integration_test/install_test.go
+++ b/integration_test/install_test.go
@@ -47,8 +47,7 @@ func TestKrewInstall(t *testing.T) {
 	test.AssertPluginFromIndex(validPlugin, "default")
 
 	receiptPath := environment.NewPaths(test.Root()).PluginInstallReceiptPath(validPlugin)
-	pluginReceipt := test.loadReceipt(receiptPath)
-	if pluginReceipt.CreationTimestamp.Time.IsZero() {
+	if r := test.loadReceipt(receiptPath); r.CreationTimestamp.Time.IsZero() {
 		t.Fatal("expected receipt to have a valid creationTimestamp")
 	}
 }
--- a/integration_test/install_test.go
+++ b/integration_test/install_test.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"testing"
 
+	"sigs.k8s.io/krew/internal/environment"
 	"sigs.k8s.io/krew/pkg/constants"
 )
 
@@ -44,6 +45,12 @@ func TestKrewInstall(t *testing.T) {
 	test.Krew("install", validPlugin).RunOrFailOutput()
 	test.AssertExecutableInPATH("kubectl-" + validPlugin)
 	test.AssertPluginFromIndex(validPlugin, "default")
+
+	receiptPath := environment.NewPaths(test.Root()).PluginInstallReceiptPath(validPlugin)
+	pluginReceipt := test.loadReceipt(receiptPath)
+	if pluginReceipt.CreationTimestamp.Time.IsZero() {
+		t.Fatal("expected receipt to have a valid creationTimestamp")
+	}
 }
 
 func TestKrewInstallReRun(t *testing.T) {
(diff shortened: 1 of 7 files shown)


integration_test/testutil_test.go
```diff
@@ -161,12 +162,9 @@ func (it *ITest) AssertPluginFromIndex(plugin, indexName string) {
 	it.t.Helper()
 
 	receiptPath := environment.NewPaths(it.Root()).PluginInstallReceiptPath(plugin)
-	r, err := receipt.Load(receiptPath)
-	if err != nil {
-		it.t.Fatalf("error loading receipt: %v", err)
-	}
-	if r.Status.Source.Name != indexName {
-		it.t.Errorf("wanted index '%s', got: '%s'", indexName, r.Status.Source.Name)
+	pluginReceipt := it.loadReceipt(receiptPath)
```
corneliusweig, line 165:
```suggestion
	r := it.loadReceipt(receiptPath)
```
I'd prefer to stick with `r` here. The scope where this variable is used is very small.

integration_test/install_test.go
```diff
@@ -44,6 +45,12 @@ func TestKrewInstall(t *testing.T) {
 	test.Krew("install", validPlugin).RunOrFailOutput()
 	test.AssertExecutableInPATH("kubectl-" + validPlugin)
 	test.AssertPluginFromIndex(validPlugin, "default")
+
+	receiptPath := environment.NewPaths(test.Root()).PluginInstallReceiptPath(validPlugin)
+	pluginReceipt := test.loadReceipt(receiptPath)
+	if pluginReceipt.CreationTimestamp.Time.IsZero() {
+		t.Fatal("expected receipt to have a valid creationTimestamp")
+	}
```
corneliusweig, line 53:
```suggestion
	if r := test.loadReceipt(receiptPath); r.CreationTimestamp.Time.IsZero() {
		t.Fatal("expected receipt to have a valid creationTimestamp")
	}
```

integration_test/upgrade_test.go
```diff
@@ -45,15 +45,23 @@ func TestKrewUpgrade(t *testing.T) {
 
 	// plugins installed via manifest get the special "detached" index so this needs to
 	// be changed to default in order for it to be upgraded here
-	receipt := environment.NewPaths(test.Root()).PluginInstallReceiptPath(validPlugin)
-	modifyReceiptIndex(t, receipt, "default")
+	receiptPath := environment.NewPaths(test.Root()).PluginInstallReceiptPath(validPlugin)
```
corneliusweig, line 48: Let's stick with this: 
```suggestion
	receipt := environment.NewPaths(test.Root()).PluginInstallReceiptPath(validPlugin)
```

integration_test/upgrade_test.go
```diff
@@ -45,15 +45,23 @@ func TestKrewUpgrade(t *testing.T) {
...
+	modifyReceiptIndex(t, receiptPath, "default")
+	pluginReceipt := test.loadReceipt(receiptPath)
+	initialCreationTimestamp := pluginReceipt.CreationTimestamp
```
corneliusweig, line 51: We don't need `pluginReceipt`. So let's simplify to
```suggestion
	initialTimestamp := test.loadReceipt(receiptPath).CreationTimestamp
```
(same below)

corneliusweig (commented):
Thanks for taking a look. I have some small suggestions, but looks good overall!

ahmetb:
> adds the current time as the timestamp when a plugin is installed and leaves it alone when something is upgraded

Is there a particular reason why we leave it untouched during the upgrade? I’m ok with it, just curious why? To let user figure out when they first installed a plugin?

PR #699, opened 2021-03-13
Release separated archives for each platform (#697)

### LINKED ISSUE #697: [releases] provide platform-specific binaries instead of tarball
Would be nice to have separate binaries that could be downloaded independently for installation.

There could be compressed version of each also to save bandwidth when needed.

Rebirth of #622

zigarn:
Fixes #697 

In install instructions, use `--archive` (and therefore `--manifest-url`) to not download again the tarball when launching `krew install` command.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,7 +80,7 @@ jobs:
       uses: svenstaro/upload-release-action@v2
       with:
         repo_token: ${{ secrets.GITHUB_TOKEN }}
-        file: out/krew.*
+        file: out/krew*
         tag: ${{ github.ref }}
         overwrite: true
         file_glob: true
--- a/docs/RELEASING_KREW.md
+++ b/docs/RELEASING_KREW.md
@@ -13,12 +13,10 @@
     ```sh
     krew=out/bin/krew-darwin_amd64 # assuming macOS amd64
 
-    KREW_ROOT="$(mktemp -d --tmpdir krew-XXXXXXXXXX)" KREW_OS=darwin \
-        $krew install --manifest=out/krew.yaml --archive=out/krew.tar.gz && \
-    KREW_ROOT="$(mktemp -d --tmpdir krew-XXXXXXXXXX)" KREW_OS=linux \
-        $krew install --manifest=out/krew.yaml --archive=out/krew.tar.gz && \
-    KREW_ROOT="$(mktemp -d --tmpdir krew-XXXXXXXXXX)" KREW_OS=windows \
-        $krew install --manifest=out/krew.yaml --archive=out/krew.tar.gz
+    for osarch in darwin_amd64 darwin_arm64 linux_amd64 linux_arm linux_arm64 windows_amd64; do
+      KREW_ROOT="$(mktemp -d --tmpdir krew-XXXXXXXXXX)" KREW_OS="${osarch%_*}" KREW_ARCH="${osarch#*_}" \
+          $krew install --manifest=out/krew.yaml --archive="out/krew-${osarch}.tar.gz"
+    done
     ```
 
 ### Release a new version
--- a/hack/make-release-artifacts.sh
+++ b/hack/make-release-artifacts.sh
@@ -25,46 +25,48 @@ if [[ ! -d "${bin_dir}" ]]; then
   exit 1
 fi
 
-krew_tar_archive="krew.tar.gz"
-krew_exe="krew.exe"
-
-# copy license
-cp -- "${SCRIPTDIR}/../LICENSE" "./${bin_dir}"
-
-# create a out/krew.exe convenience copy
-if [[ -x "./${bin_dir}/krew-windows_amd64.exe" ]]; then
-  cp -- "./${bin_dir}/krew-windows_amd64.exe" "./out/krew.exe"
-fi
-
-# consistent timestamps for files in bindir to ensure consistent checksums
-while IFS= read -r -d $'\0' f; do
-  echo "modifying atime/mtime for $f"
-  TZ=UTC touch -at "0001010000" "$f"
-  TZ=UTC touch -mt "0001010000" "$f"
-done < <(find "${bin_dir}" -print0)
-
-echo >&2 "Creating ${krew_tar_archive} archive."
-(
-  cd "${bin_dir}"
-  tar --use-compress-program "gzip --no-name" -cvf "${SCRIPTDIR}/../out/${krew_tar_archive}" ./*
-)
-
 checksum_cmd="shasum -a 256"
 if hash sha256sum 2>/dev/null; then
   checksum_cmd="sha256sum"
 fi
+checksum_sed=""
 
-tar_sumfile="out/${krew_tar_archive}.sha256"
-tar_checksum="$(eval "${checksum_cmd[@]}" "out/${krew_tar_archive}" | awk '{print $1;}')"
-echo >&2 "${krew_tar_archive} checksum: ${tar_checksum}"
-echo "${tar_checksum}" >"${tar_sumfile}"
-echo >&2 "Written ${tar_sumfile}."
+while IFS= read -r -d $'\0' f; do
+  archive_dir="$(mktemp -d)"
+  cp "$f" "${archive_dir}"
+  cp -- "${SCRIPTDIR}/../LICENSE" "${archive_dir}"
+  name="$(basename "$f" .exe)"
+  archive="${name}.tar.gz"
+  echo >&2 "Creating ${archive} archive."
+  (
+    cd "${archive_dir}"
+    # consistent timestamps for files in archive dir to ensure consistent checksums
+    TZ=UTC touch -t "0001010000" ./*
+    tar --use-compress-program "gzip --no-name" -cvf "${SCRIPTDIR}/../out/${archive}" ./*
+  )
 
-exe_sumfile="out/krew.exe.sha256"
-exe_checksum="$(eval "${checksum_cmd[@]}" "out/${krew_exe}" | awk '{print $1;}')"
-echo >&2 "${krew_exe} checksum: ${exe_checksum}"
-echo "${exe_checksum}" >"${exe_sumfile}"
-echo >&2 "Written ${exe_sumfile}."
+  # create sumfile
+  sumfile="out/${archive}.sha256"
+  checksum="$(eval "${checksum_cmd[@]}" "out/${archive}" | awk '{print $1;}')"
+  echo >&2 "${archive} checksum: ${checksum}"
+  echo "${checksum}" >"${sumfile}"
+  echo >&2 "Written ${sumfile}."
+
+  # prepare krew manifest sed
+  checksum_sed="${checksum_sed};s/$(tr "[[:lower:]-]" "[[:upper:]_]" <<<${name})_CHECKSUM/${checksum}/"
+
+done < <(find "${bin_dir}" -type f -print0)
+
+# create a out/krew.exe convenience copy
+if [[ -x "./${bin_dir}/krew-windows_amd64.exe" ]]; then
+  krew_exe="krew.exe"
+  cp -- "./${bin_dir}/krew-windows_amd64.exe" "./out/${krew_exe}"
+  exe_sumfile="out/krew.exe.sha256"
+  exe_checksum="$(eval "${checksum_cmd[@]}" "out/${krew_exe}" | awk '{print $1;}')"
+  echo >&2 "${krew_exe} checksum: ${exe_checksum}"
+  echo "${exe_checksum}" >"${exe_sumfile}"
+  echo >&2 "Written ${exe_sumfile}."
+fi
 
 # Copy and process krew manifest
 git_describe="$(git describe --tags --dirty --always)"
@@ -74,7 +76,5 @@ if [[ ! "${git_describe}" =~ v.* ]]; then
   git_describe="v0.0.0-detached+${git_describe}"
 fi
 krew_version="${TAG_NAME:-$git_describe}"
-cp ./hack/krew.yaml ./out/krew.yaml
-sed -i "s/KREW_TAR_CHECKSUM/${tar_checksum}/g" ./out/krew.yaml
-sed -i "s/KREW_TAG/${krew_version}/g" ./out/krew.yaml
+sed "${checksum_sed};s/KREW_TAG/${krew_version}/g" ./hack/krew.yaml >./out/krew.yaml
 echo >&2 "Written out/krew.yaml."
--- a/hack/krew.yaml
+++ b/hack/krew.yaml
@@ -26,8 +26,8 @@ spec:
       https://krew.sigs.k8s.io/docs/user-guide/quickstart/.
 
   platforms:
-  - uri: https://github.com/kubernetes-sigs/krew/releases/download/KREW_TAG/krew.tar.gz
-    sha256: KREW_TAR_CHECKSUM
+  - uri: https://github.com/kubernetes-sigs/krew/releases/download/KREW_TAG/krew-darwin_amd64.tar.gz
+    sha256: KREW_DARWIN_AMD64_CHECKSUM
     bin: krew
     files:
     - from: ./krew-darwin_amd64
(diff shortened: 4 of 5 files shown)


docs/RELEASING_KREW.md
```diff
@@ -14,11 +14,11 @@
...
 
     KREW_ROOT="$(mktemp -d --tmpdir krew-XXXXXXXXXX)" KREW_OS=darwin \
-        $krew install --manifest=out/krew.yaml --archive=out/krew.tar.gz && \
+        $krew install --manifest=out/krew.yaml --archive=out/krew-darwin_amd64.tar.gz && \
     KREW_ROOT="$(mktemp -d --tmpdir krew-XXXXXXXXXX)" KREW_OS=linux \
-        $krew install --manifest=out/krew.yaml --archive=out/krew.tar.gz && \
+        $krew install --manifest=out/krew.yaml --archive=out/krew-linux_amd64.tar.gz && \
     KREW_ROOT="$(mktemp -d --tmpdir krew-XXXXXXXXXX)" KREW_OS=windows \
-        $krew install --manifest=out/krew.yaml --archive=out/krew.tar.gz
+        $krew install --manifest=out/krew.yaml --archive=out/krew-windows_amd64.tar.gz
```
corneliusweig, line 21: Can you please do something like this instead:

```sh
for osarch in darwin_amd64 darwin_arm64 linux_amd64 linux_arm linux_arm64 windows_amd64; do
   KREW_ROOT="$(mktemp -d --tmpdir krew-XXXXXXXXXX)" KREW_OS="${osarch%_*}" \
     $krew install --manifest=out/krew.yaml --archive="out/krew-${osarch}.tar.gz"
done
```

site/content/docs/user-guide/setup/install.md
```diff
@@ -24,10 +24,10 @@ Krew self-hosts).
       set -x; cd "$(mktemp -d)" &&
       OS="$(uname | tr '[:upper:]' '[:lower:]')" &&
       ARCH="$(uname -m | sed -e 's/x86_64/amd64/' -e 's/\(arm\)\(64\)\?.*/\1\2/' -e 's/aarch64$/arm64/')" &&
-      curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/latest/download/krew.tar.gz" &&
-      tar zxvf krew.tar.gz &&
-      KREW=./krew-"${OS}_${ARCH}" &&
-      "$KREW" install krew
+      KREWNAME="krew-${OS}_${ARCH}" &&
```
corneliusweig, line 27: Why didn't you like `KREW`?
```suggestion
      KREW="krew-${OS}_${ARCH}" &&
```

site/content/docs/user-guide/setup/install.md
```diff
@@ -24,10 +24,10 @@ Krew self-hosts).
...
+      curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/latest/download/${KREWNAME}.tar.gz" &&
+      tar zxvf "${KREWNAME}.tar.gz" &&
+      ./"${KREWNAME}" install --manifest-url https://github.com/kubernetes-sigs/krew/releases/latest/download/krew.yaml --archive ./${KREWNAME}.tar.gz
```
corneliusweig, line 30:
```suggestion
      ./"${KREWNAME}" install krew
```

The `--manifest` and `--archive` options are only for testing. When a plugin is installed like this, it will be in a detached state where it is not updated when new versions are published on `krew-index`. So we _do_ need to download twice here.

site/content/docs/user-guide/setup/install.md
```diff
@@ -52,10 +52,10 @@ Krew self-hosts).
       set -x; set temp_dir (mktemp -d); cd "$temp_dir" &&
       set OS (uname | tr '[:upper:]' '[:lower:]') &&
       set ARCH (uname -m | sed -e 's/x86_64/amd64/' -e 's/\(arm\)\(64\)\?.*/\1\2/' -e 's/aarch64$/arm64/') &&
-      curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/latest/download/krew.tar.gz" &&
-      tar zxvf krew.tar.gz &&
       set KREWNAME krew-$OS"_"$ARCH &&
```
corneliusweig, line 55:
```suggestion
      set KREW krew-$OS"_"$ARCH &&
```

Aha, so that's where `KREWNAME` came from (see other comment above). It's good to use the same name. I have a slight preference for `KREW` though.

site/content/docs/user-guide/setup/install.md
```diff
@@ -52,10 +52,10 @@ Krew self-hosts).
...
-      ./$KREWNAME install krew &&
+      curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/latest/download/$KREWNAME.tar.gz" &&
+      tar zxvf $KREWNAME.tar.gz &&
+      ./$KREWNAME install --manifest-url https://github.com/kubernetes-sigs/krew/releases/latest/download/krew.yaml --archive ./$KREWNAME.tar.gz &&
```
corneliusweig, line 58:
```suggestion
      ./$KREWNAME install krew &&
```

Same as above.

site/content/docs/user-guide/setup/install.md
corneliusweig, line 30: Changes in this file will affect the installation instructions seen by users for the current release. However, these instructions only become effective, once we craft a new release with the other changes within this PR. So the doc updates must go into a different PR that will be merged after the next release.
So please extract this into a separate PR.

hack/make-release-artifacts.sh
```diff
@@ -25,46 +25,47 @@ if [[ ! -d "${bin_dir}" ]]; then
...
-done < <(find "${bin_dir}" -print0)
+  archive_dir="$(mktemp -d)"
+  cp "$f" "${archive_dir}"
+  cp -- "${SCRIPTDIR}/../LICENSE" "${archive_dir}"
+  archive="$(basename "$f" .exe).tar.gz"
+  echo >&2 "Creating ${archive} archive."
+  (
+    cd "${archive_dir}"
+    # consistent timestamps for files in archive dir to ensure consistent checksums
+    TZ=UTC touch -amt "0001010000" ./*
```
corneliusweig, line 43: I imagine this could make trouble on Mac/BSD, because the man says

> -a     change only the access time
> -m     change only the modification time

So I'm worried they take the _only_ too seriously. If you can check it works then great.

However, we can still make the change, because nowadays the release is only done by GH  actions and they run on Linux where I have verified that the result is the same.

hack/make-release-artifacts.sh
```diff
@@ -25,46 +25,47 @@ if [[ ! -d "${bin_dir}" ]]; then
...
-# consistent timestamps for files in bindir to ensure consistent checksums
 while IFS= read -r -d $'\0' f; do
-  echo "modifying atime/mtime for $f"
-  TZ=UTC touch -at "0001010000" "$f"
-  TZ=UTC touch -mt "0001010000" "$f"
-done < <(find "${bin_dir}" -print0)
+  archive_dir="$(mktemp -d)"
+  cp "$f" "${archive_dir}"
+  cp -- "${SCRIPTDIR}/../LICENSE" "${archive_dir}"
+  archive="$(basename "$f" .exe).tar.gz"
```
corneliusweig, line 38: Let's extract this into a variable, because it is reused a few times.
```suggestion
  name=$(basename "$f" .exe)
  archive="${name}.tar.gz"
```

hack/make-release-artifacts.sh
```diff
@@ -25,46 +25,47 @@ if [[ ! -d "${bin_dir}" ]]; then
...
+  echo >&2 "${archive} checksum: ${checksum}"
+  echo "${checksum}" >"${sumfile}"
+  echo >&2 "Written ${sumfile}."
 
-checksum_cmd="shasum -a 256"
-if hash sha256sum 2>/dev/null; then
-  checksum_cmd="sha256sum"
-fi
+  # prepare krew manifest sed
+  checksum_sed="${checksum_sed};s/KREW_$(basename "$f" .exe | sed 's/krew-\(.*\)/\U\1/')_CHECKSUM/${checksum}/"
```
corneliusweig, line 55: Let's make the intent clearer by using `tr` here:
```suggestion
  checksum_sed="${checksum_sed};s/$(tr "[[:lower:]-]" "[[:upper:]_]" <<< ${name}
)_CHECKSUM/${checksum}/"
```

hack/make-release-artifacts.sh
```diff
@@ -75,6 +76,6 @@ if [[ ! "${git_describe}" =~ v.* ]]; then
 fi
 krew_version="${TAG_NAME:-$git_describe}"
 cp ./hack/krew.yaml ./out/krew.yaml
-sed -i "s/KREW_TAR_CHECKSUM/${tar_checksum}/g" ./out/krew.yaml
+sed -i "${checksum_sed}" ./out/krew.yaml
```
corneliusweig, line 79:
```suggestion
sed "${checksum_sed}" ./hack/krew.yaml > ./out/krew.yaml
```
And remove the `cp ./hack/krew.yaml ./out/krew.yaml` in the previous line

corneliusweig (commented):
Thanks for taking this on! This is a welcome improvement.

Please take a look at my comments.

.github/workflows/ci.yml
```diff
@@ -80,7 +80,7 @@ jobs:
       uses: svenstaro/upload-release-action@v2
       with:
         repo_token: ${{ secrets.GITHUB_TOKEN }}
-        file: out/krew.*
+        file: out/krew*
```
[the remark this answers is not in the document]
zigarn: This would exclude the `krew.exe[.sha256]` and `krew.yaml`

docs/RELEASING_KREW.md
zigarn: I thought about it but didn't know if not everything was tested on purpose.

zigarn: Oh! It was on purpose to not mix architectures. Will add `KREW_ARCH`.

site/content/docs/user-guide/setup/install.md
zigarn: As commented further, bash and fish installs were using different variable names. I thought `KREWNAME` was more appropriate as the var is used for the name of the tarball and things, but `KREW` is OK for me.

hack/make-release-artifacts.sh
zigarn: The doc start with 
> Update the access and modification times of each FILE to the current time.
So I guess the *only* means that this option will change this one only compared to both.
And at the end, using both options is same as using none...

hack/make-release-artifacts.sh
zigarn: Ah ah, I did hesitate between both.

hack/make-release-artifacts.sh
zigarn: Did also integrated the next sed in it.

site/content/docs/user-guide/setup/install.md
zigarn: Extracted in #700

Previous comments on it have been applied.

zigarn: Arf, too bad.
Maybe would be possible to allow `--archive` without `--manifest[-url]` to use local archive?

corneliusweig: That does not work either, I'm afraid. It should not be a big deal though. Installation is only done once, and you are already cutting down the download size.

hack/make-release-artifacts.sh
corneliusweig: :D good catch

docs/RELEASING_KREW.md
```diff
@@ -13,12 +13,10 @@
...
     krew=out/bin/krew-darwin_amd64 # assuming macOS amd64
 
-    KREW_ROOT="$(mktemp -d --tmpdir krew-XXXXXXXXXX)" KREW_OS=darwin \
-        $krew install --manifest=out/krew.yaml --archive=out/krew.tar.gz && \
-    KREW_ROOT="$(mktemp -d --tmpdir krew-XXXXXXXXXX)" KREW_OS=linux \
-        $krew install --manifest=out/krew.yaml --archive=out/krew.tar.gz && \
-    KREW_ROOT="$(mktemp -d --tmpdir krew-XXXXXXXXXX)" KREW_OS=windows \
-        $krew install --manifest=out/krew.yaml --archive=out/krew.tar.gz
+    for osarch in darwin_amd64 darwin_arm64 linux_amd64 linux_arm linux_arm64 windows_amd64; do
+      KREW_ROOT="$(mktemp -d --tmpdir krew-XXXXXXXXXX)" KREW_OS="${osarch%_*}" KREW_ARCH="${osarch#*_}" \
```
corneliusweig, line 17: :) I didn't even think of setting `KREW_ARCH`. Great that we do now!

corneliusweig (commented):
Works flawlessly, thanks! The only thing that I could not verify is that our release automation uploads the right files. It looks right though.

/lgtm
/approve
/hold
In case @ahmetb wants to take a look too.

site/content/docs/user-guide/setup/install.md
zigarn: Looks like it works, by removing the lines doing the check in cmd/krew/cmd/install.go and compiling.

```diff
diff --git i/cmd/krew/cmd/install.go w/cmd/krew/cmd/install.go
index a574346..ac4b265 100644
--- i/cmd/krew/cmd/install.go
+++ w/cmd/krew/cmd/install.go
@@ -96,10 +96,6 @@ Remarks:
                                return errors.New("must specify either specify either plugin names (via positional arguments or STDIN), or --manifest/--manifest-url; not both")
                        }
 
-                       if *archiveFileOverride != "" && *manifest == "" && *manifestURL == "" {
-                               return errors.New("--archive can be specified only with --manifest or --manifest-url")
-                       }
-
                        var install []pluginEntry
                        for _, name := range pluginNames {
                                indexName, pluginName := pathutil.CanonicalPluginName(name)
```

```console
$ curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/latest/download/krew.tar.gz"
$ ./out/bin/krew-linux_amd64 install krew --archive ./krew.tar.gz
Adding "default" plugin index from https://github.com/kubernetes-sigs/krew-index.git.
Installing plugin: krew
Installed plugin: krew
\
 | Use this plugin:
 | 	kubectl krew
 | Documentation:
 | 	https://krew.sigs.k8s.io/
 | Caveats:
 | \
 |  | krew is now installed! To start using kubectl plugins, you need to add
 |  | krew's installation directory to your PATH:
 |  | 
 |  |   * macOS/Linux:
 |  |     - Add the following to your ~/.bashrc or ~/.zshrc:
 |  |         export PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH"
 |  |     - Restart your shell.
 |  | 
 |  |   * Windows: Add %USERPROFILE%\.krew\bin to your PATH environment variable
 |  | 
 |  | To list krew commands and to get help, run:
 |  |   $ kubectl krew
 |  | For a full list of available plugins, run:
 |  |   $ kubectl krew search
 |  | 
 |  | You can find documentation at
 |  |   https://krew.sigs.k8s.io/docs/user-guide/quickstart/.
 | /
/

$ tail ~/.krew/receipts/krew.yaml
      matchLabels:
        arch: amd64
        os: windows
    sha256: a26deea175f70264260d59a4e061778a892f8a8e301ac261660dd7d24c551c99
    uri: https://github.com/kubernetes-sigs/krew/releases/download/v0.4.1/krew.tar.gz
  shortDescription: Package manager for kubectl plugins.
  version: v0.4.1
status:
  source:
    name: default
```

Source is not 'detached'.
And I can confirm tat there is no download of archive through the install.

ahmetb: I think we reserved those arguments specifically for local testing so they are used for installing "detached" plugins. I’m not sure if it's worth changing that just to prevent downloading a small tarball once more.

corneliusweig:
/hold cancel
Sorry, this slipped through. Feel free to be more impatient and ping me next time :)

zigarn:
@corneliusweig, it has to wait for a release anyway, so no reason to be impatient.
BTW, any ETA for next release?

PR #702, opened 2021-03-15
ISSUE-696: Explore use of t.TempDir()

### LINKED ISSUE #696: Explore use of t.TempDir()
Go 1.15 started offering t.TempDir() function https://golang.org/doc/go1.15#io/ioutil
We can now deprecate our implementation of `testutil.NewTempDir(t)`.

/help-wanted
/good-first-issue

Evalle:
**PR Summary:**
- Fixes #696

**For the reviewers:** 
I'm not sure this is the right implementation, as @ahmetb asked to _deprecate_ the implementation of `testutil.NewTempDir(t)` should I add some `deprecation notice`? Thanks in advance for the advices/reviews.

### DIFF
--- a/internal/testutil/tempdir.go
+++ b/internal/testutil/tempdir.go
@@ -36,15 +36,7 @@ type TempDir struct {
 // when the test exits.
 func NewTempDir(t *testing.T) *TempDir {
 	t.Helper()
-	root, err := ioutil.TempDir("", "krew-test")
-	if err != nil {
-		t.Fatal(err)
-	}
-	t.Cleanup(func() {
-		if err := os.RemoveAll(root); err != nil {
-			t.Logf("warning: failed to remove tempdir %s: %+v", root, err)
-		}
-	})
+	root := t.TempDir()
 
 	return &TempDir{t: t, root: root}
 }


ahmetb:
Ideally we'd get rid of this type.

Evalle:
@ahmetb thanks for the feedback! Just to be clear, you would like to get rid of [NewTempDir](https://github.com/kubernetes-sigs/krew/blob/master/internal/testutil/tempdir.go#L37) completely or of [type Tempdir](https://github.com/kubernetes-sigs/krew/blob/master/internal/testutil/tempdir.go#L30) I'm pretty new in golang so my question could sound strange. Thanks!

corneliusweig:
@Evalle Can you have a look how the methods on the `TempDir` type are used? If there are just a few usages, it would be nice to get rid of the `tempdir.go` file and move the methods as free functions elsewhere.

If there are a lot of usages, your current PR looks good to me.

Evalle:
@corneliusweig thanks for the comment. My first intention was to get rid of `NewTempDir` completely in favor of the native `t.TempDir()` but I can see that there are a lot of methods that use `TempDir`:
- https://github.com/kubernetes-sigs/krew/blob/master/internal/testutil/tempdir.go#L53
- https://github.com/kubernetes-sigs/krew/blob/master/internal/testutil/tempdir.go#L59
- https://github.com/kubernetes-sigs/krew/blob/master/internal/testutil/tempdir.go#L67
- https://github.com/kubernetes-sigs/krew/blob/master/internal/testutil/tempdir.go#L79
- https://github.com/kubernetes-sigs/krew/blob/master/internal/testutil/tempdir.go#L88

And the above functionality used by many tests, for example:
- https://github.com/kubernetes-sigs/krew/blob/master/cmd/krew/cmd/internal/setup_check_test.go#L29 
- https://github.com/kubernetes-sigs/krew/blob/master/cmd/krew/cmd/internal/setup_check_test.go#L31

ahmetb:
Yeah I was expecting maybe we have simpler unit test usages that just do NewTestUtil() in a single test function. For those, t.TempDir() can be better. But right now, there's no reason to split the usage into two. We can just accept this.
/lgtm
/approve

Any developer/build system will need to update to go1.16.

Evalle:
Thanks @ahmetb and @corneliusweig for the reviews

Issue #683, opened 2021-01-14
Krew immediately fails after installation under Macos Big Sur 


goafabric:
Following this Process:
https://krew.sigs.k8s.io/docs/user-guide/setup/install/

Krew just fails directliy with:
0114 13:48:16.421460    7059 root.go:79] failed to retrieve plugin indexes: failed to list the remote URL for index .DS_Store: command execution failure, output="": fork/exec /usr/local/bin/git: not a directory

Envirnomnet:
- Macos Big Sur
- zsh

git is accessible , as well as in the path, as also in /usr/local/bin

goafabric:
Ok the problem seems to be that Macos as always likes to create .DS_Store Folders . .which is a common thing 
After krew update there will also be on inside .krew/index .. and than everything fails.
Manually removing the .DS_Store solves the issue temp. until the next update ..

chriskim06:
Interesting, I'll try to investigate this weekend after upgrading to big sur. Thanks for doing some digging and filing this issue!

ahmetb:
I had Krew installed prior to updating to Big Sur, and I'm not observing this.

I can repro this by doing `touch ~/.krew/index/foo` followed by `kubectl krew update`.

@chriskim06 it sounds like we need to Walk only directories under $KREW_ROOT/index but we currently aren't. There's probably no situation we should consider non-dir entries as index candidates.

ahmetb:
Although I must note that a clean install didn't result in this error.
@goafabric I'm suspecting you might've opened ~/.krew directory or its subdirectories in Finder.app once? (Though, even that didn't create the issue for me.)

chriskim06:
I think we are only walking directories now that I'm taking a closer look https://github.com/kubernetes-sigs/krew/blob/master/internal/index/indexoperations/index.go#L38-L58. We use `ioutil.ReadDir` which according to the [docs](https://golang.org/pkg/io/ioutil/#ReadDir):
> ReadDir reads the directory named by dirname and returns a list of directory entries sorted by filename.

ahmetb:
@chriskim06 yeah that returns all entries (it's a readdir syscall).

chriskim06:
Oh ya makes sense, I misread the doc thinking it meant it only reads "entries that are directories" instead of "the entries in the directory"

goafabric:
> Interesting, I'll try to investigate this weekend after upgrading to big sur. Thanks for doing some digging and filing this issue!

thanks for the quick reply

[...]

PR #703, opened 2021-03-15
Fixed ListIndexes treating file as index candidates

lauchokyip:
Fixes #...https://github.com/kubernetes-sigs/krew/issues/683
Related issue: #... ListIndexes treating file as index candidates

### DIFF
--- a/internal/index/indexoperations/index.go
+++ b/internal/index/indexoperations/index.go
@@ -36,14 +36,17 @@ type Index struct {
 // ListIndexes returns a slice of Index objects. The path argument is used as
 // the base path of the index.
 func ListIndexes(paths environment.Paths) ([]Index, error) {
-	dirs, err := ioutil.ReadDir(paths.IndexBase())
+	entries, err := ioutil.ReadDir(paths.IndexBase())
 	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/index/indexoperations/index.go
```diff
@@ -42,8 +42,11 @@ func ListIndexes(paths environment.Paths) ([]Index, error) {
 	}
 
 	indexes := []Index{}
-	for _, dir := range dirs {
-		indexName := dir.Name()
+	for _, file := range dirs {
```
corneliusweig, line 45: It is confusing that `dirs` does not contain directories but also files. Let's make this clear by renaming to `entries`.

```suggestion
	for _, e := range entries {
```

corneliusweig (commented):
Just a minor naming improvements, otherwise LGTM

internal/index/indexoperations/index.go
lauchokyip: Thanks, will work on it now :)

lauchokyip:
@corneliusweig Fixed 👍🏼

corneliusweig:
Thanks!
/lgtm
/approve

PR #707, opened 2021-04-05
Clarify the role of the plugin name

corneliusweig:
This is a follow-up to https://github.com/kubernetes-sigs/krew-index/pull/1157. Apparently, it was not sufficiently clear that the plugin name determines how a plugin is invoked.

Related issue: https://github.com/kubernetes-sigs/krew-index/pull/1157
/assign @chriskim06


ssup2:
@corneliusweig I think that's enough to understand the relationship between plugin name and invoking plugin. Thank :)

PR #718, opened 2021-09-18
Show plugin version on plugin overview page

corneliusweig:
I have been looking for that information several times and think it's fine to have it there.


chriskim06:
does the preview page work for you? I'm seeing a 502 when trying to load the plugins on that page (but this change seems ok)

PR #719, opened 2021-09-26
feat: make fmt.FPrintln instead of klog while printing error message

### LINKED ISSUE #717: [Bug]: additional information within the error message which might not be helpful to the users
Hello, once we run the **_info_** subcommand without given any arguments, we realized that there is additional information within the error message which might not be helpful to the users like the following:

```shell
$ k krew info
F0917 11:53:32.628902   56188 root.go:79] accepts 1 arg(s), received 0
```

This error message comes from [here](https://github.com/kubernetes-sigs/krew/blob/master/cmd/krew/cmd/root.go#L79), so we propose to fix this to avoid showing this information.

cc: @dentrax @erkanzileli

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

Fixes #717 
Related issue: #717

Screen Shot 2021-09-26 at 11 14 37

cc: @dentrax @erkanzileli


cmd/krew/cmd/root.go
```diff
@@ -74,9 +74,9 @@ You can invoke krew through kubectl: "kubectl krew [command]..."`,
 func Execute() {
 	if err := rootCmd.Execute(); err != nil {
 		if klog.V(1).Enabled() {
-			klog.Fatalf("%+v", err) // with stack trace
+			fmt.Fprintf(os.Stderr, "%+v\n", err) // with stack trace
```
ahmetb, line 77: this probably should be still klog. users don't run in this mode anyway.

developer-guy: okay, makes sense, thank you 🙏

corneliusweig:
As far as I understand, @developer-guy you agree that keeping `klog` here is the better choice, right? If not, please reopen.
/close

developer-guy:
hi @corneliusweig, yes but just for printing the stack trace of the error like ahmetb said, fmt package would be ok for the other one.

developer-guy:
@corneliusweig how can I re-open it?

ahmetb:
@corneliusweig I think for the user-visible message at -v=0 we can print without klog.

/lgtm
/approve

developer-guy:
sorry for the ping @corneliusweig but can you please tell me what can I do for completing this issue? Because I think it is ready to merge, isn't it?

PR #736, opened 2021-11-07
feat: skip install on validate-krew-manifest

caarlos0:
I'm working in [integrating the generation of krew manifests into GoReleaser](https://github.com/goreleaser/goreleaser/pull/2639), and one thing I would like to do there is a test to validate that test-generated manifests are valid... 

But, while testing, the URLs are all faked, so, installing them won't work.

I would still like to be able to validate everything else though, so I added a new flag which we can pass `-install=false` so it won't actually install anything...

Happy to hear your thoughts on this!

Cheers!

### DIFF
--- a/cmd/validate-krew-manifest/main.go
+++ b/cmd/validate-krew-manifest/main.go
@@ -39,9 +39,11 @@ import (
 )
 
 var flManifest string
+var flInstall bool
 
 func init() {
 	flag.StringVar(&flManifest, "manifest", "", "path to plugin manifest file")
+	flag.BoolVar(&flInstall, "install", true, "install the plugin as part of the validation")
 }
 
 func main() {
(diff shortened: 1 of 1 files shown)


cmd/validate-krew-manifest/main.go
```diff
@@ -39,9 +39,11 @@ import (
 )
 
 var flManifest string
+var flInstall bool
 
 func init() {
 	flag.StringVar(&flManifest, "manifest", "", "path to plugin manifest file")
+	flag.BoolVar(&flInstall, "install", true, "install the plugin as part of the validation")
```
ahmetb, line 46: I think it could be better to have a `-skip-install` what do you think?

ahmetb:
I am supportive of this. But I also want to point out that _very frequently_ we encounter issues with the archive files themselves (e.g. wrong paths in the manifest causing the plugin to fail installing, or validateLicenseFileExists which this PR deactivates). This PR can still help you do some basic structural validation.

How do you plan on calling this program via goreleaser? Will you need a Go package, or does the existing binary suffice?

cmd/validate-krew-manifest/main.go
caarlos0: sounds good!

caarlos0:
> How do you plan on calling this program via goreleaser? Will you need a Go package, or does the existing binary suffice?

right now the plan was to run it only on our own generated files on test phase (aka our golden files).

its already doing so using my fork: https://github.com/goreleaser/goreleaser/blob/a5719cf0095bd1394cb9e8706b73a826139eaf74/internal/pipe/krew/krew_test.go#L794-L806

cmd/validate-krew-manifest/main.go
caarlos0: pushed @ahmetb

ahmetb:
> right now the plan was to run it only on our own generated files on test phase (aka our golden files).

IMO this is probably not quite necessary. The plugin spec has been fairly stable now for 2+ years, and if we were to introduce a change like apiVersion bump, we'd most likely do it in a way that's backwards compatible (because there are 150+ manifests out there that we cannot possibly update, and we don't wanna break all that).

If you ask me, there's no need to test your golden files. If somebody using Goreleaser notices their auto-generated manifest is not accepted by Krew's validator, they would let you know by opening an issue. what do you think?

caarlos0:
> IMO this is probably not quite necessary. The plugin spec has been fairly stable now for 2+ years, and if we were to introduce a change like apiVersion bump, we'd most likely do it in a way that's backwards compatible (because there are 150+ manifests out there that we cannot possibly update, and we don't wanna break all that).

the idea is to make sure the files we generate are valid...

ahmetb:
> the idea is to make sure the files we generate are valid...

okay :D I was mostly proposing that you can probably validate them manually (using eyes) and then check them in. If you want to test them as well, it's acceptable.

I think we can add this, I am not too hesitant since this is mostly a tool we don't ship to users.

/lgtm
/approve

caarlos0:
Nice, thanks @ahmetb !

PR #739, opened 2021-11-16
Prevent windows msys shell from interpreting @{upstream} arg

### LINKED ISSUE #737: Windows can not handle @{upstream} correcetly
```
W1116 14:16:38.877539    8220 update.go:168] failed to update index "default": reset index at "C:\\Users\\mattn\\.krew\\index\\default" failed: command execution failure, output="fatal: ambiguous argument '@upstream': unknown revision or path not in the working tree.\nUse '--' to separate paths from revisions, like this:\n'git <command> [<revision>...] -- [<file>...]'\n": exit status 128
failed to update the following indexes: default
: reset index at "C:\\Users\\mattn\\.krew\\index\\default" failed: command execution failure, output="fatal: ambiguous argument '@upstream': unknown revision or path not in the working tree.\nUse '--' to separate paths from revisions, like this:\n'git <command> [<revision>...] -- [<file>...]'\n": exit status 128
```

mattn:
Fixes #737

### DIFF
--- a/internal/gitutil/git.go
+++ b/internal/gitutil/git.go
@@ -80,7 +80,8 @@ func Exec(pwd string, args ...string) (string, error) {
 	klog.V(4).Infof("Going to run git %s", strings.Join(args, " "))
 	cmd := osexec.Command("git", args...)
 
-	// NOTE: https://github.com/kubernetes-sigs/krew/pull/739
+	// NOTE: on windows, msys2 and cygwin interpret args e.g. @{upstream} and therefore
+	// acts differently than git-for-windows. see #739
 	if runtime.GOOS == "windows" {
 		environ := os.Environ()
 
--- a/internal/gitutil/git.go
+++ b/internal/gitutil/git.go
@@ -81,10 +81,10 @@ func Exec(pwd string, args ...string) (string, error) {
 	cmd := osexec.Command("git", args...)
 
 	// NOTE: on windows, msys2 and cygwin interpret args e.g. @{upstream} and therefore
-	// acts differently than git-for-windows. see #739
+	// acts differently than git-for-windows.
+	// see https://github.com/kubernetes-sigs/krew/pull/739
 	environ := os.Environ()
 	if runtime.GOOS == "windows" {
-
 		env := "MSYS=noglob"
 		if v := os.Getenv("MSYS"); v != "" {
 			env += " " + v
--- a/internal/gitutil/git.go
+++ b/internal/gitutil/git.go
@@ -79,9 +79,26 @@ func GetRemoteURL(dir string) (string, error) {
 func Exec(pwd string, args ...string) (string, error) {
 	klog.V(4).Infof("Going to run git %s", strings.Join(args, " "))
 	cmd := osexec.Command("git", args...)
+
+	// NOTE: https://github.com/kubernetes-sigs/krew/pull/739
 	if runtime.GOOS == "windows" {
-		cmd.Env = append(os.Environ(), "MSYS=noglob")
+		environ := os.Environ()
+
+		env := "MSYS=noglob"
+		if v := os.Getenv("MSYS"); v != "" {
+			env += " " + v
+		}
+		environ = append(environ, env)
+
+		env = "CYGWIN=noglob"
+		if v := os.Getenv("CYGWIN"); v != "" {
+			env += " " + v
+		}
+		environ = append(environ, env)
+
+		cmd.Env = environ
 	}
+
 	cmd.Dir = pwd
 	buf := bytes.Buffer{}
 	var w io.Writer = &buf
--- a/internal/gitutil/git.go
+++ b/internal/gitutil/git.go
@@ -83,8 +83,9 @@ func Exec(pwd string, args ...string) (string, error) {
 	// NOTE: on windows, msys2 and cygwin interpret args e.g. @{upstream} and therefore
 	// acts differently than git-for-windows.
 	// see https://github.com/kubernetes-sigs/krew/pull/739
-	environ := os.Environ()
 	if runtime.GOOS == "windows" {
+		environ := os.Environ()
+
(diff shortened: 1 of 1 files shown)


ahmetb:
Thanks for the patch. I'd love to understand this first.

To me this is very awkward because you'd think something like this would be necessary only if we were executing the `git` process in a `/bin/bash -c "..."` or `cmd.exe /c "..."`.

but we're doing this execution through the Go os/exec package which uses [CreateProcess](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa?redirectedfrom=MSDN) API in Win32, that's why I am not fully getting why such escaping is necessary.

Does this syscall somehow have an arguments interpretation/evaluation step that we're not aware of? It would be good to learn why's this happening first.

mattn:
msys2 shell handle `@{}` in peculiar argument parser.

image

mattn:
image

mattn:
FYI, On Windows, arguments are parsed in the process it-self. (not a shell part like UNIX)

ahmetb:
@mattn I still don't understand the problem and I do need to understand the problem fully so that we do not break other hundreds or thousands of Windows users who have been using Krew on Windows just fine.

So it seems in your gcc example that the args are sent to the process correctly. (I don't know why you're showing msys echo.exe here).

What matters here is whether `git` is handling the `@{upstream}` arg correctly and based on other Windows users I have asked, the answer for them seems to be yes, that argument works just fine. I want to understand why is your git behaving differently.

mattn:
If the program is linked with msys2 or cygwin runtime, that escaping is required. AFAIK, git on Windows shound be linked those runtimes.

mattn:
One another candidate for fixing is quote.
image

mattn:
This is a part converting that `@{}`.

https://github.com/cygwin/cygwin/blob/44a79a6eca3d322020dda0919023d78dda129d4d/winsup/cygwin/dcrt0.cc#L338-L344

mattn:
How about to use `remotes/origin/HEAD` instead of `@{upstream}` ?

mattn:
If found a spell to fix this issue `MSYS=noglob`.

internal/gitutil/git.go
```diff
@@ -78,6 +79,9 @@ func GetRemoteURL(dir string) (string, error) {
 func Exec(pwd string, args ...string) (string, error) {
 	klog.V(4).Infof("Going to run git %s", strings.Join(args, " "))
 	cmd := osexec.Command("git", args...)
+	if runtime.GOOS == "windows" {
+		cmd.Env = append(os.Environ(), "MSYS=noglob")
```
mattn, line 83: If you prefer to append CYGWI=noglob also, please let me known.

Or we should append the variable like below?
```
msys := os.Getenv("MSYS") + ":noglob"
```

ahmetb:
> How about to use `remotes/origin/HEAD` instead of `@{upstream}` ?

IIRC this was previously discussed and we wanted to do it this way for a reason, but I might be wrong. See the blame history, maybe that helps.



> I found a spell to fix this issue `MSYS=noglob`.

That's great, although again, I am not sure if _we_ should set this or whether it should be set by the user in their .bashrc or whatever they use. I am not sure what % of Krew users are using MSYS/Cygwin, but this little env-override (especially since it's only for windows, and only does _something_ on cygwin/msys) might be okay...

mattn:
I think this is less wrong for users for msys2 git.

* This change only affect to users who use msys2 git.
* This change only affect to this git command.

internal/gitutil/git.go
```diff
@@ -78,6 +79,26 @@ func GetRemoteURL(dir string) (string, error) {
 func Exec(pwd string, args ...string) (string, error) {
 	klog.V(4).Infof("Going to run git %s", strings.Join(args, " "))
 	cmd := osexec.Command("git", args...)
+
+	// NOTE: https://github.com/kubernetes-sigs/krew/pull/739
+	if runtime.GOOS == "windows" {
+		environ := os.Environ()
```
ahmetb, line 85: do you mind if we move `environ:=...` and `cmd.Env=environ` to outside the `if`?

internal/gitutil/git.go
```diff
@@ -78,6 +79,26 @@ func GetRemoteURL(dir string) (string, error) {
 func Exec(pwd string, args ...string) (string, error) {
 	klog.V(4).Infof("Going to run git %s", strings.Join(args, " "))
 	cmd := osexec.Command("git", args...)
+
+	// NOTE: https://github.com/kubernetes-sigs/krew/pull/739
```
ahmetb, line 83: let's add a bit of more explanation here, e.g.

// NOTE: on windows, msys2 and cygwin interpret args e.g. `@{upstream}` and therefore
// acts differently than git-for-windows. see https://github.com/kubernetes-sigs/krew/pull/739

mattn: Thank you.

internal/gitutil/git.go
mattn: Sure.

ahmetb:
/retitle Prevent windows msys shell from interpreting @{upstream} arg

ahmetb:
for future reference, I'll drop the issue you opened 5 years ago here ;) https://github.com/msys2/MSYS2-packages/issues/522#issue-144797722

/cc @corneliusweig 
any concerns?

corneliusweig (approved):
Looks good to me. It's a bit sad that this setting needs to  be added in application code and not in a more central place. But I guess that  would break other msys/cygwin usage patterns.

Fun story: by chance I just saw that we transitively depend on your go-colorable project :)

mattn:
> But I guess that would break other msys/cygwin usage patterns.

FYI, go-colorable does not break msys/cygwin.

image

mattn:
Sorry, I was confused and commented wrong mistaken with another PR.

PR #751, opened 2021-12-30
check all existing index

### LINKED ISSUE #750: List fails without default index
```
[N] ❯ kubectl krew list
F1230 10:47:56.057894   30356 root.go:79] krew local plugin index is not initialized (run "kubectl krew update")
```

I still have the same problem after update.

list command always check default index, but update command will skip add default when found other index. https://github.com/kubernetes-sigs/krew/blob/master/cmd/krew/cmd/update.go#L139-L142

Junnplus:
Fixes #750.

We shouldn't just check the default index.

### DIFF
--- a/cmd/krew/cmd/root.go
+++ b/cmd/krew/cmd/root.go
@@ -17,6 +17,7 @@ package cmd
 import (
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"math/rand"
 	"os"
 	"strings"
@@ -209,10 +210,23 @@ func cleanupStaleKrewInstallations() error {
 }
 
 func checkIndex(_ *cobra.Command, _ []string) error {
-	if ok, err := gitutil.IsGitCloned(paths.IndexPath(constants.DefaultIndexName)); err != nil {
-		return errors.Wrap(err, "failed to check local index git repository")
-	} else if !ok {
+	entries, err := ioutil.ReadDir(paths.IndexBase())
+	if err != nil {
+		return errors.Wrap(err, "failed to list directory")
+	}
+	if len(entries) == 0 {
 		return errors.New(`krew local plugin index is not initialized (run "kubectl krew update")`)
+	}
+	for _, e := range entries {
+		if !e.IsDir() {
+			continue
+		}
+		indexPath := paths.IndexPath(e.Name())
+		if ok, err := gitutil.IsGitCloned(indexPath); err != nil {
+			return errors.Wrap(err, "failed to check local index git repository")
+		} else if !ok {
+			return errors.New(`krew local plugin index is not initialized (run "kubectl krew update")`)
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


cmd/krew/cmd/root.go
```diff
@@ -209,11 +210,21 @@ func cleanupStaleKrewInstallations() error {
...
-		return errors.Wrap(err, "failed to check local index git repository")
-	} else if !ok {
+	entries, err := ioutil.ReadDir(paths.IndexBase())
+	if err != nil {
+		return errors.Wrap(err, "failed to list directory")
+	}
+	if len(entries) == 0 {
 		return errors.New(`krew local plugin index is not initialized (run "kubectl krew update")`)
 	}
+	for _, e := range entries {
```
chriskim06, line 220: I think we should add something like [this](https://github.com/kubernetes-sigs/krew/blob/master/internal/index/indexoperations/index.go#L46-L48) as a check here. We've seen before for mac users who get a `.DS_STORE` file in their krew directories (not sure how).

cmd/krew/cmd/root.go
```diff
@@ -209,11 +210,21 @@ func cleanupStaleKrewInstallations() error {
...
+		indexPath := paths.IndexPath(e.Name())
+		if ok, err := gitutil.IsGitCloned(indexPath); err != nil {
+			return errors.Wrap(err, "failed to check local index git repository")
+		} else if !ok {
+			return errors.New(`krew local plugin index is not initialized (run "kubectl krew update")`)
```
chriskim06, line 225: I think this error message needs to be updated but I'm not exactly sure to what. if we've hit this case that means that there was a non git directory in `paths.IndexBase()`. my first instinct is to probably fail if we encounter this since we don't want users manually messing around with their krew folders, krew should take care of that. maybe an error like `non git directory found in index folder`? although I think that's somewhat cryptic.

cc @ahmetb @corneliusweig for your inputs on this.

Junnplus: Can we just issue a warning to remind users? like "Invalid index directory"?

Junnplus:
@chriskim06 I'm sorry to bother you, do you have any other thoughts on this PR. cc @ahmetb

cmd/krew/cmd/root.go
chriskim06: sorry this fell off my radar.

I think maybe "invalid index found" is ok. My main problem with this is that there's no additional info for the user if they run into this case. Although I think this is most likely ok since if a user runs into this that means they either manually tampered with their krew directory or there's a bug in krew. If its the latter then hopefully an issue would be filed alerting us.

Junnplus: @chriskim06 Thanks your reply, I updated error message.  Please take a look.

ahmetb:
I've manually verified this. It seems to be working fine. I added an extra integration test to ensure this scenario works.
@Junnplus sorry to kept you waiting so long. 
/approve
/lgtm

PR #755, opened 2022-02-04
Add ppc64le support

mgiessing:
Fixes #...
Related issue: #...

Added support for ppc64le architecture and therefore not really a feature content-wise

### DIFF
--- a/cmd/validate-krew-manifest/main.go
+++ b/cmd/validate-krew-manifest/main.go
@@ -242,6 +242,7 @@ func allPlatforms() []installation.OSArchPair {
 		{OS: "linux", Arch: "amd64"},
 		{OS: "linux", Arch: "arm"},
 		{OS: "linux", Arch: "arm64"},
+		{OS: "linux", Arch: "ppc64le"},
 		{OS: "darwin", Arch: "386"},
 		{OS: "darwin", Arch: "amd64"},
 		{OS: "darwin", Arch: "arm64"},
--- a/docs/RELEASING_KREW.md
+++ b/docs/RELEASING_KREW.md
@@ -13,7 +13,7 @@
     ```sh
     krew=out/bin/krew-darwin_amd64 # assuming macOS amd64
 
-    for osarch in darwin_amd64 darwin_arm64 linux_amd64 linux_arm linux_arm64 windows_amd64; do
+    for osarch in darwin_amd64 darwin_arm64 linux_amd64 linux_arm linux_arm64 linux_ppc64le windows_amd64; do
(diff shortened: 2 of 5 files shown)


prb112:
HI All, I'm a ppc64le user and would like to follow along on this work. Is there a thread to follow/discussion on updates?

mgiessing:
@soltysh can you approve or tell me what needs to be changed in order to merge?

ahmetb:
Sorry, I forgot to review this one. I don't know much about ppc architecture. Can you please explain how it is used on developer/operator machines where kubectl plugins would be installed? I'm trying to distinguish if it is more of a server architecture than it is for dev machines.

prb112:
@ahmetb the ppc64le architecture is used as a foundation for OpenShift nodes.  The kubectl is used mainly by the operators to manage the environment from either a jump server/bastion server.  

I hope this helps. @mgiessing may have more to add/or a different view point.

mgiessing:
You've brought it to the point! Besides OpenShift, ppc64le can of course also be used as the foundation for "vanilla" Kubernetes. As indicated by @prb112  kubectl is in this context mainly used by the operators of such a cluster. Since Power has a strong focus on Linux & open source this commit was part of the upstreaming work in kubespray to enable ppc64le along many of its components (e.g. krew).

ahmetb:
Thanks for the explanation.

> The kubectl is used mainly by the operators to manage the environment from either a jump server/bastion server.

Do you think people actually install kubectl plugins (i.e. convenience tools) to the bastion/jump servers? Nearly none of the kubectl plugins are not thoroughly vetted from a security perspective and I am hoping people don't run plugins from their bastion servers.

This has no extra cost to us, but I'd love to really understand which plugins you'll install with this. Today, none of the plugins distribute for `ppc64le` and I believe [the only submission we ever got](https://github.com/kubernetes-sigs/krew-index/pull/1618) was from Red Hat, so I'm trying to ensure there's actually >1 users who need this.

prb112:
Hi @ahmetb here is an example I hit today. I went to install the [minio operator](https://github.com/minio/operator/blob/master/README.md#minio-operator-and-kubectl-plugin), and it requires the krew plugin before I can install it. 

Many thanks, and I hope this helps further this PR

Paul

ahmetb:
Thanks, yeah we should probably add. 
Also keep in mind that minio operator probably doesn't offer a binary for this platform because up until now, we didn't distribute krew for it. So those plugins have to update with the addition.

mkumatag:
> Thanks, yeah we should probably add.
> Also keep in mind that minio operator probably doesn't offer a binary for this platform because up until now, we didn't distribute krew for it. So those plugins have to update their manifests here in krew-index.

@ahmetb wondering if we can merge this PR if no further comments?

mkumatag:
@ahmetb any idea when this will get into the release?

PR #760, opened 2022-03-28
Adding clarity for plugin repo naming

rothgar:
I'm adding this based on convention I've seen with other plugins but if there's something else we recommend I'm fine with that too. This clarity can help plugin developers and companies looking for naming guidance and who want to avoid any trademark confusion.

Fixes #...
Related issue: #...


ahmetb:
/lgtm
/approve
Thanks!

PR #765, opened 2022-04-11
Add s390x support

madeelibm:
Fixes #...
Related issue: #...

Add s390x support which is similar to the ppc64le architecture.


ahmetb:
Hi, do you mind explaining what sort of machines run s390x? Krew is intended to be a human tool (and not a server tool), so what would be the point of installing it on the server machines? I'm interested in hearing your use case first.

PR #772, opened 2022-05-11
Fix error format

lou-lan:


ahmetb:
I think the existing code is correct. If there's a warning from a linter, it's likely the linter has issues.

lou-lan:
> I think the existing code is correct. If there's a warning from a linter, it's likely the linter has issues.

Thanks, I tried to change my editor configuration to make him ignore it.

Issue #767, opened 2022-04-22
krew outputs incorrect warning during "krew completion [..]" command if `KREW_ROOT` is set


nifr:
## Scenario

I set `KREW_ROOT` to `${XDG_DATA_HOME}/krew`.
I add `${KREW_ROOT}/bin` to `PATH` after running `krew completion [..]`. 

Note: I'm using `zsh`.

## Result

I get a misleading warning.

## Code

```
if (( $+commands[krew] )); then
  source <(krew completion zsh)
  path=(
    $path
    "${KREW_ROOT}/bin"
  )
fi
```

## Misleading Warning

```
WARNING: To be able to run kubectl plugins, you need to add
the following to your ~/.zshrc:

    export PATH="${PATH}:${HOME}/.krew/bin"

and restart your shell.
```

## Issues

1. the warning does not reflect `KREW_ROOT` 
    * should say i.e. `export PATH="${PATH}:${KREW_ROOT:-"${HOME}/.krew"}/bin"` 
3. the warning should be muted for the `krew completion` command
    * completions itself work without krew's `bin` dir in `PATH` 
    * adding krew's `bin` dir to `PATH` after loading completions still results in a working setup

## Notes

I'm aware that it is possible to get rid of the warning by loading the completions **after** adding krew's `bin` dir to`PATH`.

PR #774, opened 2022-05-11
Mention `KREW_ROOT` on instruction

JohnTitor:
Related issue: #767
The instruction on https://github.com/kubernetes-sigs/krew/blob/93827db2b9954e356456f24cc030419a899e95b1/hack/krew.yaml#L13-L16 mentions `KREW_ROOT` but one on cmd/krew/cmd/internal/setup_check.go doesn't.
This tweaks that instruction to mention `KREW_ROOT` there too.
The fish command is a bit long as there's no simple alternative to the unset variables on fish AFAIK.


ahmetb:
/lgtm
/approve
Thanks!

PR #779, opened 2022-05-12
Resolve inconsistency on long explanations

JohnTitor:
Other long explanations are like "Do ~~~." style but the index delete command is "Does ~~~". This resolves that inconsistency.


ahmetb:
/lgtm
/approve
Thanks!

PR #783, opened 2022-05-21
Fix tests for users that set KREW_ROOT

avorima:
I've noticed that updating the [krew AUR package](https://aur.archlinux.org/packages/krew) failed and after looking into
it I saw that the tests failed because I had set `KREW_ROOT`.
This patch unsets `KREW_ROOT` for the test case that asserts the default path.

### DIFF
--- a/internal/environment/environment_test.go
+++ b/internal/environment/environment_test.go
@@ -26,9 +26,15 @@ import (
 	"sigs.k8s.io/krew/pkg/constants"
 )
 
+func setKrewRootEnv(t *testing.T, value string) {
+	t.Helper()
+	oldKrewRoot := os.Getenv("KREW_ROOT")
+	os.Setenv("KREW_ROOT", value)
+	t.Cleanup(func() { os.Setenv("KREW_ROOT", oldKrewRoot) })
+}
+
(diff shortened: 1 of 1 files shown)


JohnTitor:
@avorima CI issue has been fixed, could you rebase?

avorima:
@JohnTitor done

avorima:
@ahmetb I added a helper that sets and unsets KREW_ROOT, because the envOverride test case also modifies that value.

ahmetb:
I'm still not entirely sure why you need this. 
1. why are you setting KREW_ROOT during a test suite?
2. why is https://pkg.go.dev/testing#t.Setenv not doing what you want? you basically reimplemented it.

avorima:
When KREW_ROOT is set one (1) test case fails, because it is not expected to be set which is why it must be unset. This is the only fix that I want to get through.
Another one (1) test case expects KREW_ROOT to **have** a value, so it must be set for that one (1) test case. So to ensure that the `os.Unsetenv` is only active for one test case (like you asked), I added the helper which sets and restores the value of KREW_ROOT.

So the tests expect KREW_ROOT to have certain values at certain times and I'm just trying to make things work when it is set externally, i.e. by using `export` in the shell instead of `os.Setenv`.

PR #799, opened 2022-09-04
Support `krew search` plugins by name and description

### LINKED ISSUE #798: Support `krew search` plugins by name and description
Currently, `krew search` plugins just by plugin name, like this:
```
$ kubectl krew search kubeconfig
                           
NAME                            DESCRIPTION                                         INSTALLED
view-serviceaccount-kubeconfig  Show a kubeconfig setting to access the apiserv...  no
```

But it is better to fuzzy search plugins by name and description, like this:
```
$ kubectl krew search kubeconfig

NAME                            DESCRIPTION                                         INSTALLED
allctx                          Run commands on contexts in your kubeconfig         no
config-cleanup                  Automatically clean up your kubeconfig              no
config-registry                 Switch between registered kubeconfigs               no
ctx                             Switch between contexts in your kubeconfig          yes
discover                        Find/export kubeconfigs for cloud clusters          no
kc                              Interactive CRUD operations to manage kubeconfig    yes
konfig                          Merge, split or import kubeconfig files             no
passman                         Store kubeconfig credentials in keychains or pa...  no
switch-config                   Switches between kubeconfig files                   no
view-serviceaccount-kubeconfig  Show a kubeconfig setting to access the apiserv...  no
```

astraw99:
Fixes #798 

Support fuzzy search plugins by name and description, like this:
```
$ kubectl krew search kubeconfig

NAME                            DESCRIPTION                                         INSTALLED
allctx                          Run commands on contexts in your kubeconfig         no
config-cleanup                  Automatically clean up your kubeconfig              no
config-registry                 Switch between registered kubeconfigs               no
ctx                             Switch between contexts in your kubeconfig          yes
discover                        Find/export kubeconfigs for cloud clusters          no
kc                              Interactive CRUD operations to manage kubeconfig    yes
konfig                          Merge, split or import kubeconfig files             no
passman                         Store kubeconfig credentials in keychains or pa...  no
switch-config                   Switches between kubeconfig files                   no
view-serviceaccount-kubeconfig  Show a kubeconfig setting to access the apiserv...  no
```

Also added test case to make it correct, and checked all the other search test cases.


integration_test/search_test.go
```diff
@@ -47,6 +47,18 @@ func TestKrewSearchOne(t *testing.T) {
...
+func TestKrewSearchByNameAndDesc(t *testing.T) {
+	skipShort(t)
+
+	test := NewTest(t)
+
+	plugins := lines(test.WithDefaultIndex().Krew("search", "kubeconfig").RunOrFailOutput())
+	if len(plugins) < 11 {
+		// the first line is the header
+		t.Errorf("Expected at least %d plugins", len(plugins)-1)
+	}
```
ahmetb, line 59: This is not a very good test, I think at this point we should extract the core search functionality into a separate func and unit-test that. The integration tests for this still could be useful but they should probably test something like "a particular plugin that doesn't have x in the name but x in the description".

cmd/krew/cmd/search.go
```diff
@@ -63,10 +63,12 @@ Examples:
 		}
 
 		pluginCanonicalNames := make([]string, len(plugins))
+		pluginCanonicalDescs := make([]string, len(plugins))
```
ahmetb, line 66: you can say `pluginDescriptions`

cmd/krew/cmd/search.go
```diff
@@ -82,9 +84,21 @@ Examples:
 
 		var searchResults []string
 		if len(args) > 0 {
+			// find by name
 			matches := fuzzy.Find(strings.Join(args, ""), pluginCanonicalNames)
+			searchResultsMap := make(map[int]bool)
```
ahmetb, line 89: I wonder if we should stop sorting by name (and sort by relevance/how close it is to an exact match) instead when there's a query specified.

cmd/krew/cmd/search.go
astraw99: Updated.

cmd/krew/cmd/search.go
astraw99: The name is just one word without space, so the fuzzy search did not check the `match.Score`.
There are word spaces when search by description, so checked the `match.Score > 0`, to ignore the low score match.

astraw99:
/ping @ahmetb 
PTAL thanks.

ahmetb:
Thanks, I'll try to find some time, play with this hands-on and review.

astraw99:
OK, thanks.

cmd/krew/cmd/search.go
```diff
@@ -118,6 +113,32 @@ Examples:
 	PreRunE: checkIndex,
 }
 
+func searchByNameAndDesc(keyword string, nameData, descData []string) []string {
+	var searchResults []string
+	if len(keyword) > 0 {
```
ahmetb, line 118: Looks like you can make this easier to read this method by checking the ==0 first and returning, then moving this this block outside the if.

cmd/krew/cmd/search.go
```diff
@@ -118,6 +113,32 @@ Examples:
...
+		// find by name
+		matches := fuzzy.Find(keyword, nameData)
+		searchResultsMap := make(map[int]bool)
+		for _, m := range matches {
+			searchResults = append(searchResults, m.Str)
+			searchResultsMap[m.Index] = true
+		}
+
+		// find by short description
+		matches = fuzzy.Find(keyword, descData)
```
ahmetb, line 128: call this fuzzyMatches to make it easier to read?

cmd/krew/cmd/search_test.go
```diff
@@ -0,0 +1,67 @@
...
+		},
+		{
+			name:  "ns",
+			desc:  "Switch between Kubernetes namespaces",
+			isHit: false,
+		},
+	}
+
+	nameData := make([]string, len(tests))
+	descData := make([]string, len(tests))
```
ahmetb, line 46: No need to make() these, "var" is easier to read.

cmd/krew/cmd/search_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+package cmd
+
+import (
+	"testing"
+)
+
+func Test_searchByNameAndDesc(t *testing.T) {
+	keyword := "config"
+	tests := []struct {
```
ahmetb, line 23: testPlugins is a better name?

cmd/krew/cmd/search_test.go
```diff
@@ -0,0 +1,67 @@
...
+		{
+			name:  "ns",
+			desc:  "Switch between Kubernetes namespaces",
+			isHit: false,
+		},
+	}
+
+	nameData := make([]string, len(tests))
+	descData := make([]string, len(tests))
+	expectHitCnt := 0
```
ahmetb, line 47: You can just len(testData)?

cmd/krew/cmd/search_test.go
```diff
@@ -0,0 +1,67 @@
...
+	for _, tt := range tests {
+		nameData = append(nameData, tt.name)
+		descData = append(descData, tt.desc)
+		if tt.isHit {
+			expectHitCnt++
+		}
+	}
+
+	searchResults := searchByNameAndDesc(keyword, nameData, descData)
```
ahmetb, line 56: Perhaps a better way to structure this test func would be to:

- declare testPlugins slice
- declare scenarios and expected plugin match list
- run scenarios (using t.Run subtests) and compare expected vs got


This way you can test things explicitly like:

- positive cases
- negative cases (doesn't match)
- fuzzy cases (name matches outrank description matches)

ahmetb (changes requested):
Mostly cosmetic comments. 
/approve

ahmetb:
Sorry, the linter seems to be extra-aggressive about not preallocating slices, it seems. This wasn't the case in the past, hence my comment about making them simpler.

astraw99:
Got it, updated.
PTAL thanks.

astraw99:
@ahmetb Could you please give a lgtm label to make it merged, thanks.

cmd/krew/cmd/search.go
```diff
@@ -118,6 +113,33 @@ Examples:
 	PreRunE: checkIndex,
 }
 
+func searchByNameAndDesc(keyword string, names, descs []string) []string {
```
ahmetb, line 116: it feels weird that names and descs are separate arrays.
can you create an internal struct to carry this info perhaps?

cmd/krew/cmd/search_test.go
```diff
@@ -0,0 +1,58 @@
...
+package cmd
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func Test_searchByNameAndDesc(t *testing.T) {
+	keyword := "config"
```
ahmetb, line 24: if you're doing table driven testing, consider making this part of the struct below.

cmd/krew/cmd/search_test.go
```diff
@@ -0,0 +1,58 @@
...
+		expected []string
+	}{
+		{
+			names: []string{"switch-config", "ctx", "kc"},
+			descs: []string{
+				" Switches between kubeconfig files",
+				"Switch between contexts in your kubeconfig",
+				"Interactive CRUD operations to manage kubeconfig",
+			},
+			expected: []string{"switch-config", "ctx", "kc"},
```
ahmetb, line 37: my advice would be to use some abstract descriptions like "foo bar quux foobar" so that we can clearly see which part is it matching (name? desc? is it full match? partial/fuzzy match?)

cmd/krew/cmd/search_test.go
```diff
@@ -0,0 +1,58 @@
...
+				"Interactive CRUD operations to manage kubeconfig",
+			},
+			expected: []string{"ctx", "kc"},
+		},
+	}
+
+	for _, tp := range testPlugins {
+		t.Run(tp.names[0], func(t *testing.T) {
+			result := searchByNameAndDesc(keyword, tp.names, tp.descs)
+			if diff := cmp.Diff(tp.expected, result); diff != "" {
```
ahmetb, line 53: Do you think fzf or whatever fuzzy search library would change their algorithm and break this test (as the test expects a particular order)?

ahmetb:
Sorry for the latency here. I did a review again. I think if you refactor the tests to clearly show something like "name matches are shown first, then the description matches" it would help us maintainers reason about the code/tests later on. Thanks.

cmd/krew/cmd/search.go
astraw99: Updated.

cmd/krew/cmd/search_test.go
astraw99: Updated.

cmd/krew/cmd/search_test.go
astraw99: Updated.

cmd/krew/cmd/search_test.go
astraw99: 1. If the search library makes a change, they will ensure the compatibility, as fuzzy search is based on matching scores.
2. We specified the lib version in go.mod, so it is safe to manage the version upgrading.

astraw99:
> Sorry for the latency here. I did a review again. I think if you refactor the tests to clearly show something like "name matches are shown first, then the description matches" it would help us maintainers reason about the code/tests later on. Thanks.

Updated, PTAL thanks.

cmd/krew/cmd/search.go
```diff
@@ -30,6 +30,11 @@ import (
 	"sigs.k8s.io/krew/internal/installation"
 )
 
+type searchTarget struct {
+	names        []string
+	descriptions []string
+}
```
ahmetb, line 36: I meant 

```
type searchTarget struct {
	name        string
	description string
}
```

does that not make more sense?

ahmetb: Although I see we pass `fuzzy.Find(keyword, descriptions)`, in that case, maybe the best course of action is to specify a type like:

```
type searchItem struct { name, description string }
type searchCorpus []searchItem
func (searchCorpus) descriptions() []string
func (searchCorpus) names() []string
```

https://pkg.go.dev/golang.org/x/exp/slices and https://pkg.go.dev/golang.org/x/exp/maps (you can bring them) may have some methods to deal with

cmd/krew/cmd/search.go
```diff
@@ -118,6 +121,53 @@ Examples:
...
+		// use map to deduplicate and score > 0 to match more accurately
+		if _, exist := searchResultsMap[m.Index]; !exist && isMatchContinuous(m.MatchedIndexes) {
+			searchResults = append(searchResults, target.names[m.Index])
+		}
+	}
+
+	return searchResults
+}
+
+func isMatchContinuous(matchIndexes []int) bool {
```
ahmetb, line 151: can you explain what this does and what is the "continuous" check is for? 
ideally methods like these warrant tests. :)

cmd/krew/cmd/search.go
```diff
@@ -118,6 +121,53 @@ Examples:
...
+	// find by names
+	matches := fuzzy.Find(keyword, target.names)
+	searchResultsMap := make(map[int]bool)
+	for _, m := range matches {
+		searchResults = append(searchResults, m.Str)
+		searchResultsMap[m.Index] = true
+	}
+
+	// find by short descriptions
+	fuzzyMatches := fuzzy.Find(keyword, target.descriptions)
```
ahmetb, line 140: nit: `matches` above are technically `fuzzyMatches`, too. maybe reuse that variable?

cmd/krew/cmd/search.go
ahmetb: I now see you're looking at whether the fuzzy match occurs continuously.
May I ask why did you do that? If you do that, that is no longer becomes "fuzzy" search? 
The intent is also not clear from the code (and one has to really pay attention to see that) --or the tests?

cmd/krew/cmd/search.go
astraw99: Updated.

cmd/krew/cmd/search.go
astraw99: Deleted.
The reason is after we use fuzzy search, the matched result may be not continuous words.
So use `m.Score > 0` to make it more accurately.

cmd/krew/cmd/search.go
astraw99: Updated.
PTAL thanks.

astraw99:
ping @ahmetb
PTAL thanks.

astraw99:
ping @ahmetb 
PTAL thanks.

astraw99:
ping @ahmetb 
PTAL thanks.

ahmetb:
/lgtm
/approve
Thanks!

PR #800, opened 2022-09-04
Update plugin manifest doc

astraw99:
To make the plugin manifest doc better to new developers.

### DIFF
--- a/site/content/docs/developer-guide/plugin-manifest.md
+++ b/site/content/docs/developer-guide/plugin-manifest.md
@@ -39,6 +39,7 @@ spec:
     Restarts a pod with the given name. The existing pod
     will be deleted and created again, not a true restart.
   # 'platforms' specify installation methods for various platforms (os/arch)
+  # See all supported platforms below.
   platforms:
   - selector:
       matchExpressions:
@@ -49,7 +50,7 @@ spec:
         - linux
     # 'uri' specifies .zip or .tar.gz archive URL of a plugin
     uri: https://github.com/achanda/kubectl-restart/archive/v0.0.3.zip
-    # 'sha256' is the sha256sum of the url above
+    # 'sha256' is the sha256sum of the url (archive file) above
     sha256: d7079b79bf4e10e55ded435a2e862efe310e019b6c306a8ff04191238ef4b2b4
     # 'files' lists which files should be extracted out from downloaded archive
     files:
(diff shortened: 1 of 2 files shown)


site/content/docs/developer-guide/plugin-manifest.md
```diff
@@ -99,6 +100,8 @@ Krew makes it possible to install the same plugin on different operating systems
 (e.g., `windows`, `darwin` (macOS), and `linux`) and different architectures
 (e.g., `amd64`, `386`, `arm`, `arm64` and `ppc64le`).
 
+See [all supported platforms](https://github.com/kubernetes-sigs/krew/blob/master/cmd/validate-krew-manifest/main.go#L233-L249).
```
ahmetb, line 103: This link will almost certainly go out of date. :)

I think the right approach is to point to krew repo's releases page and say "Only plugins in platforms that krew itself is distributed are supported.", right?

site/content/docs/developer-guide/plugin-manifest.md
```diff
@@ -99,6 +100,9 @@ Krew makes it possible to install the same plugin on different operating systems
 (e.g., `windows`, `darwin` (macOS), and `linux`) and different architectures
 (e.g., `amd64`, `386`, `arm`, `arm64` and `ppc64le`).
 
+See [all supported platforms](https://github.com/kubernetes-sigs/krew/releases).
+Only plugins in platforms that krew itself is distributed (Release Assets) are supported.
```
chriskim06, line 104: also wanted to float the idea of having a static list here. i know its not great to have to update in multiple places when something new is supported but this isnt a frequent occurrence. when ppc64le was added the contributor also [remembered to update the docs](https://github.com/kubernetes-sigs/krew/pull/755/files#diff-bbfd929e7884a7e220b4bcddf4c4f992cde11c8b31fab08f874f7caf9ec87fa1R100). an option to help any drift between what's supported and the docs would be to add a comment (maybe to [here](https://github.com/kubernetes-sigs/krew/blob/master/hack/krew.yaml#L28)) as a reminder to update the docs area with the new platform.

if not then i think it reads better as part of the previous paragraph like
```suggestion
The supported platforms for plugins are the ones that Krew are distributed in.
See all supported platforms on the [releases page](https://github.com/kubernetes-sigs/krew/releases).
```

astraw99: Got it, updated.
PTAL thanks.

astraw99:
/ping @ahmetb @chriskim06 
PTAL thanks.

ahmetb:
/lgtm
/approve
Thanks!

Issue #810, opened 2022-10-23
Krew completion (for fish, likely others as well) only works if ln -s kubectl-krew krew, and even then not in kubectl krew


vorburger:
#769 similar issue

When installing krew as per https://krew.sigs.k8s.io/docs/user-guide/setup/install/, for Fish (but it's likely similar for other shells),

and then activating completion like this:

    $HOME/.krew/bin/kubectl-krew completion fish > $HOME/.config/fish/completions/kubectl-krew.fish

then `kubectl krew ...` still doesn't auto-complete. However if doing:

    ln -s $HOME/.krew/bin/kubectl-krew $HOME/.krew/bin/krew

then at least `krew ...` auto-completes.

vorburger:
> and then activating completion like this:

as suggested to be added to the doc in #811, as at least a first step

ahmetb:
Krew doesn't offer completion, because kubectl plugins don't support completion yet.

If it appears like krew does have completion scripts, it's a side effect of the library we use, not by design.

/close

chriskim06:
also the next time we publish a new version the `completion` command won't be there https://github.com/kubernetes-sigs/krew/pull/771

maelvls:
With the latest kubectl version (1.26), the kubectl plugins completion now works!

To enable completion for `kubectl krew`, run the following commands:

```bash
echo '#!/bin/sh\nkubectl krew __complete "$@"' >~/.krew/bin/kubectl_complete-krew
chmod u+x ~/.krew/bin/kubectl_complete-krew
```

ahmetb:
@maelvls we should probably implement a native feature for this in the plugin manifest. Do you mind opening a separate issue so I don't forget?

maelvls:
I think https://github.com/kubernetes-sigs/krew/issues/812#issue-1439452083 covers this 👍

PR #811, opened 2022-10-23
Add Fish shell completion step to krew install doc

vorburger:
Related issue: #810


ahmetb:
https://github.com/kubernetes-sigs/krew/issues/810#issuecomment-1289253747
Also `krew` is meant to be used as `kubectl krew` so this is not the right fix. Need to wait until this is fixed in kubectl.
/close

PR #820, opened 2023-02-20
detect download failures when status code > 200

### LINKED ISSUE #819: download failures are interpreted as checksum mismatch
The krew client ignores the status code. This results in misleading errors that show a checksum mismatch when in fact the artifact was never downloaded. According to the go documentation:

```A non-2xx response doesn't cause an error```

daniel-garcia:
detect download failures when status code is > 200
#819

### DIFF
--- a/internal/download/fetch.go
+++ b/internal/download/fetch.go
@@ -41,6 +41,9 @@ func (HTTPFetcher) Get(uri string) (io.ReadCloser, error) {
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to download %q", uri)
 	}
+	if resp.StatusCode > 200 {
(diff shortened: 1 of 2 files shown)


ahmetb:
Thanks for this patch, sorry it didn't catch my attention earlier. Can we also add a test case (with a httptest server responding with 404 ) to validate this doesn't regress?

daniel-garcia:
> Thanks for this patch, sorry it didn't catch my attention earlier. Can we also add a test case (with a httptest server responding with 404 ) to validate this doesn't regress?

done.

ankishagarwal (approved):

ankishagarwal:
Can we complete this if all okay ? We need this change, thank you

ahmetb:
/lgtm

thanks.

PR #830, opened 2023-07-18
fix contributors list in release notes

chriskim06:
Fixes #...
Related issue: #...

### DIFF
--- a/hack/make-release-notes.sh
+++ b/hack/make-release-notes.sh
@@ -74,7 +74,7 @@ for f in "${download_assets[@]}"; do
 done
 echo
 echo "Thanks to our contributors for helping out with ${TAG}:"
-git log "$(git describe --tags --abbrev=0)..HEAD" --format=%an |
+git log "$(git describe --tags --match 'v*' --abbrev=0 "${TAG}^")..${TAG}" --format=%an |
   sort | uniq -c | sort -rn |
   sed -E 's,^(\s+[0-9]+\s),- ,g'
 echo
--- a/hack/make-release-notes.sh
+++ b/hack/make-release-notes.sh
@@ -74,17 +74,18 @@ for f in "${download_assets[@]}"; do
 done
 echo
 echo "Thanks to our contributors for helping out with ${TAG}:"
-git log "$(git describe --tags --match 'v*' --abbrev=0 "${TAG}^")..${TAG}" --format=%an |
+previous_version="$(git describe --tags --match 'v*' --abbrev=0 "${TAG}^")"
+git log "${previous_version}..${TAG}" --format=%an |
   sort | uniq -c | sort -rn |
   sed -E 's,^(\s+[0-9]+\s),- ,g'
 echo
-echo "(krew ${TAG} is tagged on $(date -u).)"
+echo "(krew ${TAG} was tagged on $(date -u).)"
 echo
 echo '<details>'
 echo '<summary>Merged pull requests</summary>'
 echo # this empty line is important for correct markdown rendering
 # you can pass your github token with --token here if you run out of requests
 
-"${gopath}/bin/release-notes" kubernetes-sigs krew
+"${gopath}/bin/release-notes" kubernetes-sigs krew --since "${previous_version}"
 echo '</details>'
 echo
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,8 @@ jobs:
 
     - name: Check out code into the Go module directory
       uses: actions/checkout@v3
+      with:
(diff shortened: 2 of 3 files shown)


hack/make-release-notes.sh
```diff
@@ -74,7 +74,7 @@ for f in "${download_assets[@]}"; do
 done
 echo
 echo "Thanks to our contributors for helping out with ${TAG}:"
-git log "$(git describe --tags --abbrev=0)..HEAD" --format=%an |
+git log "$(git describe --tags --match 'v*' --abbrev=0 "${TAG}^")..${TAG}" --format=%an |
```
chriskim06, line 77: we should be checking `prev version..new version` but with what it was it was checking `TAG..HEAD` which is the same commit. not sure what changed in the ci but this should fix it.

some tests i ran locally
```
$ TAG=v0.4.3 bash -c 'git log "$(git describe --tags --match "v*" --abbrev=0 "${TAG}^")..${TAG}" --format=%an | sort | uniq -c | sort -rn | sed -E "s,^(\s+[0-9]+\s),- ,g"'
- Chris Kim
- Ahmet Alp Balkan
- zhengtianbao
- Cornelius Weig
- Alexandre Garnier
- ferhat elmas
- Vincent
- Samuel Walker
- Kazuki Suda
- Kai
- Jun
- Gábor Lipták
- Evgeny Shmarnev
- Chok Yip Lau
- Carlos Alexandro Becker
- Batuhan Apaydın
- Allianzcortex

$ TAG=v0.4.2 bash -c 'git log "$(git describe --tags --match "v*" --abbrev=0 "${TAG}^")..${TAG}" --format=%an | sort | uniq -c | sort -rn | sed -E "s,^(\s+[0-9]+\s),- ,g"'
- Chris Kim
- Cornelius Weig
- Ahmet Alp Balkan
- ferhat elmas
- Vincent
- Evgeny Shmarnev
- Chok Yip Lau
- Alexandre Garnier
```

these match the outputs in the release notes (https://github.com/kubernetes-sigs/krew/releases/tag/v0.4.3 and https://github.com/kubernetes-sigs/krew/releases/tag/v0.4.2)

i added that glob match in case we ever tag with something else (highly unlikely but figured it didnt hurt to add)

ahmetb:
Funny I don't remember running into this. I ran `git show v0.4.4` after invoking this script and the contributor list was (and still is) there. And I also see https://github.com/kubernetes-sigs/krew/releases/tag/v0.4.4 is populated (did you edit it?) - so I don't know if there's a problem.

hack/make-release-notes.sh
ahmetb, line 77: this is a bit interesting though:
```
git describe --tags --match 'v*' --abbrev=0 "v0.4.3^"
v0.4.1
```
it skips v0.4.2?

chriskim06: ya im not sure why. that was the release i did so maybe i messed something up? although i do see it when i `git tag -l` and i see it here

chriskim06:
> Funny I don't remember running into this. I ran `git show v0.4.4` after invoking this script and the contributor list was (and still is) there. And I also see https://github.com/kubernetes-sigs/krew/releases/tag/v0.4.4 is populated (did you edit it?) - so I don't know if there's a problem.

ya sorry i mentioned it in slack but i manually edited it

chriskim06:
actually think i need to look at this more

/hold

ahmetb:
> when i run this on master the contributor list is empty

that's normal since you probably have v0.4.4 already tagged, but release notes must be computed before tagging in https://github.com/kubernetes-sigs/krew/blob/22f508a02fba9e1b4039ac69d583a70f1b5e510f/docs/RELEASING_KREW.md.

So if you `git tag -d v0.4.4`, I expect existing instructions to work.

> i manually edited it

I think this is how we've always done it on every release. I don't think GitHub picks up a tag's message/annotations as release description, unless you integrate with GitHub API.

chriskim06:
ah ok i see. i forgot that we manually uploaded the release notes and didnt have it in the ci till https://github.com/kubernetes-sigs/krew/pull/781/files

`actions/checkout` does a shallow clone so thats why checking the git log doesnt work there. think we'll need to update this https://github.com/actions/checkout/blob/v2.7.0/action.yml#L56-L58 to 0

chriskim06:
setting the fetch depth fixes it. here's a run i did https://github.com/chriskim06/krew/actions/runs/5602643039/jobs/10248173640 and here's the release it created https://github.com/chriskim06/krew/releases/tag/v0.4.4

the contributor list is messed up since i dont have the 0.4.3 tag but this simplifies the release process for sure.

Issue #832, opened 2023-08-10
Alternative branching for custom indexes instead default branch


ardaguclu:
Currently, when user adds a custom index, Krew clones this repository from the default branch statically. Since distributions of most of plugins are managed in these default branches(master, main, etc.), it works without problem.

On the other hand, this mechanism does not work for the plugins that using branching strategy per release. For a specific cluster version, plugin maintainers want to force users to use only specific major version(and auto updates of minor versions via Krew). For example, for my cluster v1.2.X, user should use 1.2.X which means that user should add Krew custom index only being encompassed by this branch and I'll continue working on master branch for the next release cycle, v2.0.X without impacting v1.2.X releases.

This feature proposal may look very specific, but this provides an option to plugin maintainers about just cutting a branch for a specific release and continue working on new features on their master. And old clients can still use it safely because auto upgrade will not upgrade to a very latest version which is apparently not compatible.

As maintainers of Krew, you know the use cases and features requests better than me and if my proposal still looks very exception, feel free to close this issue.

/kind proposal

ardaguclu:
I added a solution with respect to my proposal https://github.com/kubernetes-sigs/krew/pull/833. Appending branch name with `#` will set the cloned repositories default branch to the given one. It worked without any problem in my tests.

chriskim06:
it seems a little strange to me to handle this with different versions of the index. homebrew handles this by allowing [versioned formulae](https://docs.brew.sh/Versions) in their main tap. i dont know off the top of my head if that is already possible with krew (i forget exactly what parts of the plugin spec are used to setup the binary installed in `.krew/store` and the symlink in `.krew/bin`). there are other things in addition to having separate plugins that would be needed though, like how to link a specific version.

i think your implementation makes sense for solving this by versioning the index (as long as that clone command properly sets `refs/remotes/origin/{branch}` so that the `git reset --hard @{upstream}` is referencing the correct remote), but im not totally convinced it should be handled there instead of supporting different plugin versions.

ardaguclu:
I incorrectly closed issue :). Sorry for noise.

PR #833, opened 2023-08-10
Add using alternative branch support for custom indexes

ardaguclu:
Currently, when user adds a custom index, Krew clones this repository from the default branch statically. Since distributions of most of plugins are
managed in these default branches(master, main, etc.), it works without problem.

On the other hand, this mechanism does not work for the plugins that using branching strategy per release. For a specific cluster version, plugin maintainers want to force users to use only specific major version(and auto updates of minor versions via Krew). For example, for my cluster v1.2.X, user should use 1.2.X which means that user should add Krew custom index only being encompassed by this branch and I'll continue working on master branch for the next release cycle, v2.0.X. without impacting v1.2.X releases.

This PR proposes alternative branching support by appending branch name with `#` symbol. Since we are setting this branch name as default during clone phase, the rest of the functionality will remain same. If user wants to use different branch name for the same git uri, user should remove index first and re-add it with the name branch name.

Related issue: https://github.com/kubernetes-sigs/krew/issues/832

### DIFF
--- a/cmd/krew/cmd/index.go
+++ b/cmd/krew/cmd/index.go
@@ -64,9 +64,13 @@ each configured index in table format.`,
 }
 
 var indexAddCmd = &cobra.Command{
-	Use:     "add",
-	Short:   "Add a new index",
-	Long:    "Configure a new index to install plugins from.",
+	Use:   "add",
+	Short: "Add a new index",
+	Long: `Configure a new index to install plugins from.
+
+This command also supports alternative branch other than the default one
+by appending the branch name to the end of git url with # symbol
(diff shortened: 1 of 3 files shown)


ardaguclu:
I created a temporary repository https://github.com/ardaguclu/krew-index-test for testing and only `v1.2.3` branch in that repository contains a plugin(such as `example`) and uses patch in this PR by adding;
```sh
$ kubectl krew index add test https://github.com/ardaguclu/krew-index-test#v1.2.3
```

and it worked properly.

ardaguclu:
As discussed on referenced issue;
/close

PR #836, opened 2023-08-18
clone single branch with depth of 1

### LINKED ISSUE #403: Fetch only master of index, and in shallow mode
Currently, we do a plain `git fetch`:

https://github.com/kubernetes-sigs/krew/blob/b303e3ae0179431750deab95be05e7b163286c74/pkg/gitutil/git.go#L52-L54

This has some drawbacks, because it

* fetches all remote branches and tags
* .. and it fetches every commit in each branch.

It would be better, if instead we could

* just fetch the relevant branch (`master`)
* .. and only fetch the tip commit of that branch.

That way, we could keep the `index` folder as lean as possible.

---

This should do the trick:

```go
if err := git(destinationPath, "fetch", "origin", "master", "--verbose", "--depth=1"); err != nil {
}
```

```go
func EnsureCloned(uri, destinationPath string) error {
	if ok, err := IsGitCloned(destinationPath); err != nil {
		return err
	} else if !ok {
		return exec("", "clone", "--branch=master", "--depth=1", "-v", uri, destinationPath)
	}
	return nil
}
```

slimm609:
cloning only the default branch and with a depth of 1 speeds up the clone time greatly

```sh
I0818 19:01:24.786890   27900 git.go:79] Going to run git clone --depth=1 --single-branch -v git@gitlab.com:test/test.git /Users/dbrian/.krew/index/test
Cloning into '/Users/dbrian/.krew/index/test'...
I0818 19:01:25.342370   27900 fetch_tag.go:47] Parsing response from GitHub
```

```sh
I0818 19:00:39.064036   27722 git.go:79] Going to run git clone -v git@gitlab.com:test/test.git /Users/dbrian/.krew/index/test
Cloning into '/Users/dbrian/.krew/index/test'...
I0818 19:00:48.527775   27722 root.go:180] Upgrade check was skipped or has not finished
```

Fixes #403

### DIFF
--- a/internal/gitutil/git.go
+++ b/internal/gitutil/git.go
@@ -31,7 +31,7 @@ func EnsureCloned(uri, destinationPath string) error {
 	if ok, err := IsGitCloned(destinationPath); err != nil {
 		return err
 	} else if !ok {
-		_, err = Exec("", "clone", "-v", uri, destinationPath)
+		_, err = Exec("", "clone", "--depth=1", "--single-branch", "-v", uri, destinationPath)
 		return err
 	}
 	return nil


ahmetb:
I will try to take a look at this sometime but any mistake here can brick Krew installations forever, and our commit history is not particularly bloated so it's not top priority right now. Let me know if there's a particular reason you're pushing for it.

slimm609:
thanks @ahmetb.  Will larger private repos for plugins, it can take a very long time to install or update because it downloads the entire history of the repo just to install and all branch references just to install from the main branch.  This changes to only download the most recent commit (no history) and only the repo's main branch instead of the main branch plus all additional branches.

ahmetb:
Sorry for the delay here. I have a follow up question: doesn't the subsequent "git fetch origin && git reset -hard @{upstream}" command re-download the repository anyway?

If that's the case, I don't think this optimization is worth doing (and may end up being worse) for long-term installations that install/update plugins on the same machine over time.

slimm609:
> Sorry for the delay here. I have a follow up question: doesn't the subsequent "git fetch origin && git reset -hard @{upstream}" command re-download the repository anyway?
> 
> If that's the case, I don't think this optimization is worth doing (and may end up being worse) for long-term installations that install/update plugins on the same machine over time.

`git reset --hard @{upstream}` will just reset but `git fetch -v` can be updated to `git fetch -v origin HEAD` to only apply to the that branch

PR #841, opened 2023-09-13
fix: chmod 0755 store/PLUGIN/VER dir

### LINKED ISSUE #840: $KREW_ROOT/store/<plugin>/* has wrong permissions, 0700 instead of 0755
Hi there,

I tried to install `krew` using a different `KREW_ROOT`, exactly like documented here: https://krew.sigs.k8s.io/docs/user-guide/advanced-configuration/

I basically ran this command from a `sudo` user:
```
sudo KREW_ROOT=/usr/local/krew ./krew-linux_amd64 install krew
```
After that the binary was not accessible from a normal user. This is because the symlink pointed to a file that was only accessible from `root`:
```
ls -l /usr/local/krew/bin/
lrwxrwxrwx 1 root root   38 Sep 11 18:42 kubectl-krew -> /usr/local/krew/store/krew/v0.4.4/krew
```
```
ls -l /usr/local/krew/store/krew
total 12
drwx------ 2 root root 4096 Sep 11 18:42 v0.4.4/
```
After a `sudo chmod 750 /usr/local/krew/store/krew/v0.4.4` it worked fine.

Unfortunately this happens with every plugin that is installed this way. The subdirectories in `/usr/local/krew/store/*/v*` always have the wrong permissions.

For example I installed `resource-capacity` from the `sudo` user like this:
```
sudo PATH="/usr/local/krew/bin:$PATH" KREW_ROOT=/usr/local/krew kubectl krew install resource-capacity
```
Then the new directory had the wrong permissions again:
```
sudo ls -l /usr/local/krew/store/resource-capacity
total 4
drwx------ 2 root root 4096 Sep 11 18:54 v0.7.4
```

ahmetb:
Because we're doing `mktemp` (which gives 0700) and then moving it to `$KREW_ROOT/store/PLUGIN/VERSION`, we need to make sure that the directory is 0755 like the other dirs in `store`.

See #840.

### DIFF
--- a/internal/installation/move.go
+++ b/internal/installation/move.go
@@ -161,7 +161,12 @@ func moveToInstallDir(srcDir, installDir string, fos []index.FileOperation) erro
 	tmp, err := os.MkdirTemp("", "krew-temp-move")
 	klog.V(4).Infof("Creating temp plugin move operations dir %q", tmp)
 	if err != nil {
-		return errors.Wrap(err, "failed to find a temporary director")
+		return errors.Wrap(err, "failed to find a temporary directory")
+	}
+	klog.V(4).Infof("Chmoding tmpdir to 0755", tmp)
+	if err := os.Chmod(tmp, 0o755); err != nil {
+		// mktemp gives a 0700 directory but since we move this to KREW_ROOT, we need to make it 0755
+		return errors.Wrap(err, "failed to chmod temp directory")
 	}
 	defer os.RemoveAll(tmp)
 


ngoeddel-openi:
Can please someone add this `lgtm` label and remove `do-not-merge/invalid-commit-message` or what's the issue here?
Thank you!

ngoeddel-openi:
@ahmetb May I ask when you will bring out a new release?

Issue #835, opened 2023-08-18
Proposal: Download Binary File and Download License via URI


joshbrgs:
I believe it would be a great feature to accept standalone binaries as well as Archives. This could open the possibility for more projects to index their plugins with krew without needing to change their GH workflows if they are not releasing Archives. One could modify the yaml to have a license field where the URI to the License would be located. As an example with the argo rollouts project: 

```
  - uri: https://github.com/argoproj/argo-rollouts/releases/download/v1.5.1/kubectl-argo-rollouts-windows-amd64
    sha256: 4c71b696d5821b4740c0d16e55e8793146c9dc0675d1be869f73934e9f847431
    bin: kubectl-argo-rollouts
    license: https://raw.githubusercontent.com/argoproj/argo-rollouts/v1.5.1/LICENSE
    selector:
      matchLabels:
        os: windows
        arch: amd64
```

I would be willing to contribute this. 👍🏼 

/kind proposal

djmcgreal-cc:
Just ran into this over at https://github.com/ryane/kfilt/issues/14

PR #844, opened 2023-10-25
Proposal: Download Binary File and Download License via URI

joshbrgs:
Related issue: #835 

Enables the use of a binary to be declared in the yaml manifest uri, and creates a license field to pair with the bin, so that when the plugin is installed, it is packaged with bin and license similar to when archive types are installed.


ahmetb:
/reopen
/lifecycle frozen
Sorry I just couldn't find the time

PR #849, opened 2024-01-10
Update install.md: Install via `brew`

guettli:
I think installing via brew is the most straight forward way and works on most OSs (Mac, Linux, WSL).

### DIFF
--- a/site/content/docs/user-guide/setup/install.md
+++ b/site/content/docs/user-guide/setup/install.md
@@ -7,10 +7,15 @@ weight: 200
 Krew itself is a `kubectl` plugin that is installed and updated via Krew (yes,
 Krew self-hosts).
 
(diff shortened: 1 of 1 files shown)


ahmetb:
It's not a fair assumption that Homebrew is the most widely used package manager across distros. 

Separately, Krew installs and updates itself as a plugin. And we don't actively maintain the Brew formulae as part of our releases (Brew community does the upkeep of the formula.)

So I'm -1 on placing it on the top as the recommended way.

kundan2707 (commented):
/lgtm

guettli:
@ahmetb thank you for your feedback. I discovered that if I install via brew, then the shell autocomplete gets added automatically. AFAIK this is not possible with krew. At the moment I don't see a reason to use krew. I close the issue. But again, thank you for your feedback.

PR #858, opened 2024-06-25
Document Homebrew option for MacOS

kasadaamos:
Fixes #...
Related issue: #...

### DIFF
--- a/site/content/docs/user-guide/setup/install.md
+++ b/site/content/docs/user-guide/setup/install.md
@@ -14,6 +14,12 @@ Krew self-hosts).
 
 ## macOS/Linux {#posix}
 
+#### Homebrew
+
+```
+brew install krew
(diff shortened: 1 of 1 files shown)


ahmetb:
This is currently intentionally not documented, as Krew doesn't manage and update itself when installed via Homebrew. We also have no control or release automation for Homebrew at the moment. So this may require some thinking. But I'm not strongly against it.

kasadaamos:
> This is currently intentionally not documented, as Krew doesn't manage and update itself when installed via Homebrew. We also have no control or release automation for Homebrew at the moment. So this may require some thinking. But I'm not strongly against it.

Would it address the points you raise if I added them to the PR?

kasadaamos:
This isn't going anywhere, closing.

Issue #862, opened 2024-09-30
Bad output on MacOS


wallentx:
When I would run `kubeclt krew search`, it would return exactly the following:
```bash
NAME                            DESCRIPTION                                         INSTALLED
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
                                                                                    unavailable on darwin/arm64
[...]

PR #863, opened 2024-09-30
Fixing search results

wallentx:
Fixes https://github.com/kubernetes-sigs/krew/issues/862

I'm not really sure what might could be added to the tests to account for this.

### DIFF
--- a/cmd/krew/cmd/search.go
+++ b/cmd/krew/cmd/search.go
@@ -38,7 +38,7 @@ type searchItem struct {
 type searchCorpus []searchItem
 
 func (s searchCorpus) descriptions() []string {
-	var res = make([]string, len(s))
+	res := make([]string, 0, len(s))
 	for _, corpus := range s {
 		res = append(res, corpus.description)
 	}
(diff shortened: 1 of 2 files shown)


ahmetb (approved):

wallentx:
> we should probably add a test case here https://github.com/kubernetes-sigs/krew/blob/master/cmd/krew/cmd/search_test.go#L31-L62 with an empty keyword expecting the full set of names. this bug has probably been in krew for some time

I've made the requested changes to the search_test 👍

chriskim06:
actually looking into it a bit more, we have this integration test https://github.com/kubernetes-sigs/krew/blob/master/integration_test/search_test.go#L24-L34 but it didnt catch this because it just checks that there are at least 10 lines in the output. wonder if we should update this to an exact number (1 + number of plugins in the cloned index; the 1 being the header line in the search output).

wallentx:
@chriskim06 I've got `Allow edits and access to secrets by maintainers` checked, so feel free to do whatever you deem is most appropriate.

Issue #865, opened 2024-10-09
Shipping "kubectl create something" plugin is impossible


sttts:
Wile looking into shipping a `kubectl create something` plugin via krew I noticed that the `- => _` rewriting in https://github.com/kubernetes-sigs/krew/blob/48b3ea15a8f640bbfca550782a9ccf9051f13731/internal/installation/install.go#L242 makes this impossible right now.

Background: kubectl allows these create plugins (and only those) since https://github.com/kubernetes/kubernetes/pull/124123.

Basically we have this situation:
* `$ kubectl create-something` maps to the binary `kubectl-create_something`
* `$ kubectl create something` maps to the binary `kubectl-create-something`
* krew does the mentioned rewriting, turning the binary of the `create-something` plugin into the `kubectl-create_something` binary on disk (even though the binary in the plugin tar is properly named.

How could we allow to ship a `kubectl-create-something` binary?

xref https://kubernetes.slack.com/archives/CTJNMDM24/p1728488438493869

ahmetb:
TIL `kubectl create [...]` now allows plugins as of v1.30. I think we should have a `spec` field in the Plugin manifest to facilitate this. Open questions:

1. Do we do this just for `kubectl create`? what if kubectl changes its mind to allow more builtin commands to support plugins, too?
2. Do we only skip the first dash, or do we preserve all dashes

   For example should a Plugin with `name: create-x-y-z` be linked as `kubectl-create-x_y_z` or `kubectl-create-x-y-z`? The latter technically allows invocations like:

   - `kubectl create x-y-z`
   - `kubectl create x y z`
   - `kubectl create x-y z`
   - ...

  (Krew intentionally only supports the first form, i.e. plugin is registered only with its `metadata.name`, as-is).

/kind feature
/priority important-soon

mjudeikis:
@ahmetb, is there any interest in reopening this? I had proof of concept for this a long time ago, and we discussed it at KubeCon.

If there is, I can try working on this again

PR #889, opened 2025-10-07
feat: allow spaces in plugin names in order to enable sub-commands

### LINKED ISSUE #888: Feature request: Allow spaces in plugin names in order to enable sub-commands
`kubectl` [allows plugins with spaces in the name](https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/#naming-a-plugin) to provide the user experience of sub-commands:
> a plugin that wishes to be invoked whenever the command `kubectl foo bar baz` is invoked by the user, would have the filename of `kubectl-foo-bar-baz`

`krew` does not allow for this option, because it uses `.metadata.name` for the plugin name, and this cannot have spaces. Any hyphens in the name get translated to underscores in the filename, which `kubectl` translates back to hyphens in the command name.

This could be resolved by having an optional field in the Plugin YAML, `.spec.command` which could be set to an array of strings much like the `spec.containers.command` field in a Kubernetes Pod. For example:

```
apiVersion: krew.googlecontainertools.github.com/v1alpha3
kind: Plugin
metadata:
  name: foo-bar-baz
spec:
  command: ["foo", "bar", "baz"]
```

dhague:
Fixes #888

### DIFF
--- a/internal/installation/install_test.go
+++ b/internal/installation/install_test.go
@@ -81,6 +81,7 @@ func Test_createOrUpdateLink(t *testing.T) {
 	tests := []struct {
 		name       string
 		pluginName string
+		command    []string
 		binary     string
 		wantErr    bool
 	}{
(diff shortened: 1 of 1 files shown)


kavishdahekar-sap:
Also fixes https://github.com/kubernetes-sigs/krew/issues/865

Issue #886, opened 2025-08-07
Unable to install plugins without elevated privileges (Windows)


avoidik:
Hello,

An elevated privileges are need in Windows environments to be able to create symlinks. That's why I'd like to suggest an option to skip this step while installing plugins.

https://github.com/kubernetes-sigs/krew/pull/885

PR #904, opened 2026-04-26
installation: Windows hard link for plugin shims, then symlink (EvalSymlinks fallback)

carterpewpew:
[WIP]
**What this PR does**

On Windows, creating plugin shims in .krew/bin no longer depends only on os.Symlink. The code tries os.Link first (same volume, no extra symlink privilege in the common case), then falls back to os.Symlink. The source path is passed through filepath.EvalSymlinks before the symlink path so the fallback does not use an unresolved path if the install path goes through a junction or directory symlink.

Uninstall and upgrades use removeLink with an isLink helper on Windows that also allows removing hard-linked bin entries, not only symlinks.

On non-Windows, behavior is unchanged: createSymlink is a plain os.Symlink and isLink is symlink-only, matching the old logic.

**Fixes** #886

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/sahilm/fuzzy v0.1.1
 	github.com/spf13/cobra v1.10.1
 	github.com/spf13/pflag v1.0.10
+	golang.org/x/sys v0.31.0
 	k8s.io/apimachinery v0.34.1
 	k8s.io/client-go v0.34.1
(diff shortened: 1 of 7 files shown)


carterpewpew:
Sorry, accidentally opened, please ignore

ahmetb:
Looks promising, especially if you could figure out how not to break existing installations that already installed some plugins via symlinks (it should work, right?). Thanks for taking this on.

PR #908, opened 2026-08-31
fix: exclude macOS metadata from release archives

### LINKED ISSUE #892: Release archive contain additional files
I'm in the process of automating the Krew install on our workstations and found that at least the latest Krew release archive (v0.4.5) contains some MacOS specific files that IMHO should not be there. I haven't checked the archives for older versions yet. 

If it's appreciated, I could make some changes to https://github.com/kubernetes-sigs/krew/blob/master/hack/make-release-artifacts.sh and create a pull request, so that cleaner archives are created in the future.

Bisman-Singh:
Fixes #892
Related issue: #892

Release tarballs built with bsdtar on macOS contain AppleDouble (`._LICENSE`, `._krew-*`) entries and `LIBARCHIVE.xattr.com.apple.provenance` pax headers, because bsdtar stores resource forks and extended attributes by default. GNU tar warns `Ignoring unknown extended header keyword` when extracting them (see the listing in #892).

This sets `COPYFILE_DISABLE=1` (bsdtar-specific environment variable, ignored by GNU tar) to suppress the AppleDouble entries, and passes `--no-xattrs` (supported by both bsdtar and GNU tar) to keep extended attributes out of the archive.

Verification:
- On macOS, the unfixed script produced archives with `./._krew-linux_amd64`, `./._LICENSE` and `com.apple.provenance` xattr headers; with this change the same build produces exactly two clean entries per archive, no pax headers.
- Extracted archives are intact: the binary runs and `LICENSE` matches the repo copy (sha256).
- On Linux (GNU tar 1.35, as on ubuntu-latest), the new invocation is accepted and produces a byte-identical archive to the current one, so CI-built release checksums are unchanged.
- `shfmt -l -i=2 hack/` (v3.0.0, the repo's pinned version) is clean; shellcheck reports no new findings.

Credit to @ruudbleeker for the diagnosis and initial patch in #893, which could not proceed due to CLA constraints. This re-derives the archive fix while keeping the existing fixed-timestamp behavior (the `touch --date` variant suggested there is GNU-only and would break running the script on macOS).


Bisman-Singh:
The `build` failure here is pre-existing and unrelated to this change: every CI run since 2026-05-15 (pushes to `master` included) fails at workflow setup because the org policy requires actions pinned to full commit SHAs while ci.yml uses floating major tags. Fix in #909 — once that lands, a rebase/rerun here should go green.

PR #909, opened 2026-08-31
ci: pin GitHub Actions to full commit SHAs

Bisman-Singh:
Related issue: #892 (blocks CI on its fix, #908)

CI is currently failing on every run — pushes to `master` included — before any job starts (last green run: 2026-03-27; failing continuously since 2026-05-15, e.g. [this master push run](https://github.com/kubernetes-sigs/krew/actions/runs?query=branch%3Amaster)):

```
The actions actions/setup-go@v6, actions/checkout@v5, and softprops/action-gh-release@v2
are not allowed in kubernetes-sigs/krew because all actions must be pinned to a full-length commit SHA.
```

The kubernetes-sigs org policy requires full-SHA pinning, and the workflow currently uses floating major tags. This pins each action to the exact commit its current major tag resolves to (verified against the upstream repos' tags), so behavior is identical:

- `actions/setup-go@v6` → `924ae3a1cded613372ab5595356fb5720e22ba16` (v6.5.0)
- `actions/checkout@v5` → `fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09` (v5.1.0)
- `softprops/action-gh-release@v2` → `3bb12739c298aeb8a4eeaf626c5b8d85266b0e65` (v2.6.2)

No workflow logic changes.

PR #910, opened 2026-09-16
shallow clone and fetch the plugin index

### LINKED ISSUE #403: Fetch only master of index, and in shallow mode

magic-peach:
Fixes #403
Related issue: n/a

Cloning and updating the plugin index currently does a full clone and a full fetch of every branch and every commit, even though only the tip of the default branch is ever used.

This makes `clone` and each subsequent `fetch` shallow with `--depth=1`. Since a `--depth` clone already implies `--single-branch`, only the tip of the default branch is fetched from the start, and passing `--depth=1` on every later `fetch` keeps it from growing back into a full history the way a plain `fetch` after a shallow clone would.

There was a previous attempt at this in #836 that only patched the `clone` call. ahmetb pointed out there that the subsequent `fetch` during updates could still end up re-downloading everything, which would make the optimization pointless for long running installs. This change also passes `--depth=1` to the `fetch` in `updateAndCleanUntracked`, so index updates stay shallow too.

`internal/gitutil` had no tests before. Added coverage for `EnsureCloned`, `EnsureUpdated`, and `GetRemoteURL`, including a test that checks the clone stays at a single commit across repeated updates.

Testing:
- go build ./...
- go vet ./...
- go test ./...


soltysh:
Once we bump the versions in https://github.com/kubernetes-sigs/krew/pull/909 this will be able to land, but overall this lgtm

soltysh (approved):
/lgtm
/approve

soltysh:
/lgtm cancel

@magic-peach can you rebase your PR on top of master, which should have the gh actions fix from https://github.com/kubernetes-sigs/krew/pull/909

magic-peach:
@soltysh could you please check?

soltysh (approved):
/lgtm

PR #911, opened 2026-09-20
pin the index clone's remote name to origin

### LINKED ISSUE #853: krew fails to work if `defaultRemoteName` is assigned in ~/.gitconfig
I have this config option set in my global ~/.gitconfig file:

```
[clone]
	defaultRemoteName = home
```

Which changes the default remote name to `home`. However, `krew` assumes that the default remote name is always going to be `origin`, due to https://github.com/kubernetes-sigs/krew/blob/efaaa22897a069937e3017dbeb01dfeb6a6aec80/internal/gitutil/git.go#L75

This causes most commands to fail:

```
> kubectl krew install [PACKAGE_NAME_HERE]
failed to retrieve plugin indexes: failed to list the remote URL for index default: command execution failure, output="": exit status 1
```

It would be better if one of the following were true:
1. `krew` would detect this setting and use the actual remote name
2. `krew` did not use this setting from the global config, perhaps by running with a blank config
3. The error message was better when `GetRemoteURL` returned empty data

magic-peach:
Fixes #853
Related issue: n/a

If a user has `[clone] defaultRemoteName` set in their global `~/.gitconfig`, `EnsureCloned`'s `git clone` picks up that setting instead of using `origin`, since the clone call never pins the remote name explicitly. Every later command in this package hardcodes `origin` though, `fetch`, `reset --hard @{upstream}`, `clean`, and `GetRemoteURL`'s `remote.origin.url` lookup, so the index ends up cloned under the wrong remote name and every subsequent operation on it fails. This is the exact failure reported in the issue: `kubectl krew install` errors with "failed to retrieve plugin indexes."

This adds `-o origin` to the clone command so the remote name is always `origin` regardless of the user's global git config.

Added a test that sets `GIT_CONFIG_GLOBAL` to a temporary gitconfig with `defaultRemoteName` set to something other than `origin`, then clones and confirms `GetRemoteURL` still succeeds. Confirmed the test actually fails against the code before this fix, with the exact symptom from the issue, before trusting it as a real regression test.

Testing:
- go build ./...
- go vet ./...
- go test ./internal/gitutil/... (full package, all tests pass)

## AI Disclosure
PR description was written using AI


skitt:
See https://kubernetes.slack.com/archives/C01672LSZL0/p1789992857325439

/close

